PR #441, opened 2020-02-08
Need4speed - Migrate to openZeppelin test-environment

NicholasDotSol:
- Moved testing contracts to the main contracts folder under `testContracts` This is so they get compiled into the main build folder. `contract.fromArtifact` looks there for the `.json`

- `deploySystem` utility was altered to use the double-parameter `link` after `detect network()`. function needs some cleanup.

Must compile before the test run. Test does not automatically compile. this is a great feature. 
Will look into getting gas reporting back, or other alternatives. For now, massive speed-ups are much more useful


Shadowfiend:
@NicholasDotSol I'm ready to move on this if you want to make the change to the `testContracts` dir.

NicholasDotSol:
@Shadowfiend,
`test` jsut tests, no compile 
`refresh` compiles and tests. 

How about having `test` compile and tests, and `refresh` just run tests without compiling. Naming might be a little confusing, definitely want to avoid confusion when changing contract code and tests not reflecting

implementation/.soliumignore
```diff
@@ -1,2 +1,3 @@
 # Vendored
-contracts/bitcoin-spv
\ No newline at end of file
+contracts/bitcoin-spv
+contracts/testContracts
```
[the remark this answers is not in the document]
NicholasDotSol: Totally agree. Removed the condition and fixed all errors and other accumulated warnings

implementation/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "clean": "rm -rf build/",
     "compile": "truffle compile",
-    "test": "NODE_ENV=test truffle test",
+    "test": "npx mocha --exit --recursive test",
```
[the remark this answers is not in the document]
NicholasDotSol: true, no `npx` needed

Shadowfiend:
> How about having test compile and tests, and refresh just run tests without compiling. Naming might be a little confusing, definitely want to avoid confusion when changing contract code and tests not reflecting

Could even just have `quick-test` or `rerun-tests` or `rerun`. I think all of those are a little clearer, and I like having the default path of `test` do the compile. Make the dangerous-but-quick one a hidden gem, and the safe-but-slow one the default… Though also we should just fix this and introduce some proper change detection somewhere along the line. Someday :p

implementation/test/BTCETHPriceFeedTest.js
```diff
@@ -1,14 +1,13 @@
-const BTCETHPriceFeed = artifacts.require('BTCETHPriceFeed')
-const MockMedianizer = artifacts.require('MockMedianizer')
 
-import { createSnapshot, restoreSnapshot } from './helpers/snapshot'
-const BN = require('bn.js')
-const chai = require('chai')
-const expect = chai.expect
-const bnChai = require('bn-chai')
-chai.use(bnChai(BN))
+const { createSnapshot, restoreSnapshot } = require('../testHelpers/helpers/snapshot.js')
```
Shadowfiend: > I don't remember why I did this. Proabably mocha was getting confused.

Haha if you don't remember why, let's undo it and see if it's still broken.

implementation/test/BTCETHPriceFeedTest.js
```diff
@@ -37,7 +36,7 @@ contract('BTCETHPriceFeed', (accounts) => {
       // 7152.55 / 142.28 = 50.2709446162
       // 50.2709446162 * 10^10
       // 502,709,446,162 wei
-      expect(price).to.eq.BN('502709446162')
+      expect(price).to.be.bignumber.equal(new BN('502709446162'))
```
Shadowfiend: Do we gain anything from using this (much more verbose) version vs continuing to use `bn-chai`?

implementation/testHelpers/testDeployer.js
```diff
@@ -0,0 +1,171 @@
+const { contract } = require('@openzeppelin/test-environment')
```
Shadowfiend: By committing this as an entirely new file after deleting the old one, we unfortunately lose all the history on the old file, so we lose all sense that “It's mostly similar to deployTestDeposit.js”, and any tracking of what OpenZeppelin specifically required :(

Another way to do it would have been to move `deployTestDeposit.js` alongside the other files, then edit it to do what you wanted here. Git would have tracked that more or less correctly.

implementation/test/BTCETHPriceFeedTest.js
NicholasDotSol: we gain code size.
going back to `to.eq.BN`

implementation/test/BTCETHPriceFeedTest.js
NicholasDotSol: `json` files are okay in the test directory, but I prefer the file outside the test directory, which I think should only contain test files

Shadowfiend: While I'm not really planning on blocking this PR on it, that's not really how this works in any language, for better or worse. `test` directories are more or less universally for test-*related* files, whether helpers or not. At a high level, consider top-level directories to be “compile units”. You wouldn't really need to compile test helpers without tests or vice versa.

This is also why I hate having to move our test contracts into `contracts/`, but we can cross that bridge later.

PR #459, opened 2020-02-13
Documentation updates and improvements

NicholasDotSol:
Quality of life doc changes. Update docs, Add them where needed and use cleaner language.


Shadowfiend:
PR description please!

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -40,12 +40,12 @@ library DepositFunding {
...
-    /// @dev            This should be called by an approved contract, not a developer
-    /// @param _d       deposit storage pointer
-    /// @param _m       m for m-of-n
-    /// @param _m       n for m-of-n
-    /// @return         True if successful, otherwise revert
+    /// @notice         The system can spin up a new deposit.
+    /// @dev            This should be called by an approved contract, not a developer.
+    /// @param _d       deposit storage pointer.
+    /// @param _m       m for m-of-n.
+    /// @param _m       n for m-of-n.
```
Shadowfiend: For m-of-n what? We can use more info on `can spin up a new deposit`. `@notice` should answer the question “What happens when I call this?” The answer can't be “The system can spin up a new deposit.“ The answer *could* be, “Spins up a new deposit with ... ”

This is also meant to be called internally. `@notice` is meant for [*end users*](https://solidity.readthedocs.io/en/develop/natspec-format.html#tags); the most we should tell an end user here is “don't call this, call `DepositFactory.createNewDeposit` instead!”.

For a developer, we probably want to tell them what happens if they do call it, so they know not to.

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -74,25 +74,25 @@ library DepositFunding {
         return true;
     }
 
-    /// @notice     slashes the signers partially for committing fraud before funding occurs
-    /// @dev        called only by notifyFraudFundingTimeout
+    /// @notice     slashes the signers partially for committing fraud before funding occurs.
```
Shadowfiend: Caps here.

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -74,25 +74,25 @@ library DepositFunding {
...
     function distributeSignerBondsToFunder(DepositUtils.Deposit storage _d) internal {
         uint256 _seized = _d.seizeSignerBonds();
         _d.depositOwner().transfer(_seized);  // Transfer whole amount
     }
 
-    /// @notice     Anyone may notify the contract that signing group setup has timed out
-    /// @dev        We rely on the keep system punishes the signers in this case
-    /// @param  _d  deposit storage pointer
+    /// @notice     Anyone may notify the contract that signing group setup has timed out.
+    /// @dev        We rely on the keep system punishes the signers in this case.
```
Shadowfiend: rely on the keep system to punish*

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -141,15 +141,15 @@ library DepositFunding {
...
-    /// @dev                    ECDSA is NOT SECURE unless you verify the digest
-    /// @param  _d              deposit storage pointer
-    /// @param  _v              Signature recovery value
-    /// @param  _r              Signature R value
-    /// @param  _s              Signature S value
-    /// @param _signedDigest    The digest signed by the signature vrs tuple
-    /// @param _preimage        The sha256 preimage of the digest
-    /// @return                 True if successful, otherwise revert
+    /// @notice                 Anyone can provide a signature that was not requested to prove fraud during funding.
+    /// @dev                    ECDSA is NOT SECURE unless you verify the digest.
```
Shadowfiend: This isn't really actionable. What does that mean to me as a dev who is calling `provideFundingECDSAFraudProof`? Is that an issue for me as the dev caller, or is it an issue for the keep implementor?

implementation/contracts/deposit/DepositRedemption.sol
```diff
@@ -319,14 +319,14 @@ library DepositRedemption {
     }
 
     /// @notice                 Check the redemption transaction input and output vector to ensure the transaction spends
-    ///                         the correct UTXO and sends value to the appropreate public key hash
+    ///                         the correct UTXO and sends value to the appropreate public key hash.
```
Shadowfiend: appropriate*

implementation/contracts/deposit/OutsourceDepositLogging.sol
```diff
@@ -6,23 +6,23 @@ import {DepositUtils} from "./DepositUtils.sol";
 library OutsourceDepositLogging {
 
 
-    /// @notice               Fires a Created event
-    /// @dev                  We append the sender, which is the deposit contract that called
-    /// @param  _keepAddress  The address of the associated keep
+    /// @notice               Fires a Created event.
+    /// @dev                  We append the sender, which is the deposit contract that called.
```
Shadowfiend: Who is we? Append it to what? Definitely need to be clearer here.

implementation/contracts/deposit/OutsourceDepositLogging.sol
```diff
@@ -63,8 +63,8 @@ library OutsourceDepositLogging {
         );
     }
 
-    /// @notice     Fires a RegisteredPubkey event
-    /// @dev        The logger is on a system contract, so all logs from all deposits are from the smae addres
+    /// @notice     Fires a RegisteredPubkey event.
+    /// @dev        The logger is on a system contract, so all logs from all deposits are from the same addres.
```
Shadowfiend: All of these have an `addres`->`address` typo.

implementation/contracts/price-feed/MockMedianizer.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity ^0.5.10;
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "../external/IMedianizer.sol";
 
-/// @notice A mock implementation of a medianizer price oracle
+/// @notice A mock implementation of a medianizer price oracle.
```
Shadowfiend: This should probably be `@title`.

implementation/contracts/proxy/DepositFactory.sol
```diff
@@ -63,12 +63,12 @@ contract DepositFactory is CloneFactory, TBTCSystemAuthority{
...
-    /// @dev                   Calls createNewDeposit from deposit contract as init method.
-    ///                        We don't offer pure createClone, meaning that the only way
-    ///                        to create a clone is by also calling createNewDeposit()
-    ///                        Deposits created this way will never pass by state 0 (START)
-    /// @return                True if successful, otherwise revert
+    /// @notice                Creates a new deposit instance and mints a TDT.    
+    ///                        This function is currently the only way to create a new deposit. 
+    /// @dev                   Calls Deposit.createNewDeposit to initialize the instance.
+    ///                        Mints the TDT to the function caller. 
+    //                         See (TBTCDepositToken for more info on TDTs).
```
Shadowfiend: 👌

Keep in mind you can put backticks around code to get proper formatting in the generated docs.

implementation/contracts/system/FeeRebateToken.sol
```diff
@@ -3,6 +3,14 @@ pragma solidity ^0.5.10;
 import "openzeppelin-solidity/contracts/token/ERC721/ERC721Metadata.sol";
 import "./VendingMachineAuthority.sol";
 
+/// @title  Fee Rebate Token
+/// @notice The Fee Rebate Token (FRT) is a non fungible token (ERC721)
+///         the ID of which corresponds to a given deposit address. 
+///         If the corresponding deposit is still active, ownership of this token
+///         could result in reimbursement of the singer fee payed to open the deposit.
```
Shadowfiend: Probably worth going into a little more detail about when it will and won't result in that, but this is a great start.

implementation/contracts/system/FeeRebateToken.sol
```diff
@@ -3,6 +3,14 @@ pragma solidity ^0.5.10;
...
+/// @dev    This token is minted automatically when a TDT (TBTCDepositToken.sol)
+///         is exchanged for TBTC (TBTCToken.sol) via the 
+///         Vending Machine (VendingMachine.sol).
```
Shadowfiend: Rather than referencing filenames here, I think you should just reference the contract names.

implementation/contracts/system/TBTCDepositToken.sol
```diff
@@ -11,11 +11,13 @@ import "./DepositFactoryAuthority.sol";
...
-//          TBTC tokens whose value is backed 1-to-1 by the corresponding
-//          deposit's BTC.
+///         TBTC tokens whose value is backed 1-to-1 by the corresponding
+///         deposit's BTC.
 /// @dev Currently, TDTs are minted using the uint256 casting of the
 ///      corresponding deposit contract's address. That is, the TDTs id is
-///      convertible to the deposit's address and vice versa.
+///      convertible to the deposit's address and vice versa. TDTs are minted
+///      automatically by the factory during each deposit's initialization. See
+///      DepositFactory.createNewDeposit() for more info on how the TDT is minted.
```
Shadowfiend: 👌

implementation/contracts/system/TBTCDepositToken.sol
```diff
@@ -33,13 +35,14 @@ contract TBTCDepositToken is ERC721Metadata, DepositFactoryAuthority {
...
-    /// @param _tokenId uint256 ID of the token to query the existence of
-    /// @return bool whether the token exists
+    /// @param _tokenId uint256 ID of the token to query the existence of.
+    /// @return bool whether the token exists.
     function exists(uint256 _tokenId) public view returns (bool) {
         return _exists(_tokenId);
     }
 
-    /// @notice           Set allowance for other address and notify.
+    /// @notice           Allow another contract to spend on the caller's behalf.
```
Shadowfiend: Another address, really---no one knows if it's a contract or not (though it may be worth calling out “(generally another smart contract)”, for example).

implementation/contracts/system/TBTCDepositToken.sol
```diff
@@ -53,5 +56,7 @@ contract TBTCDepositToken is ERC721Metadata, DepositFactoryAuthority {
         spender.receiveApproval(msg.sender, _tdtId, address(this), _extraData);
     }
 }
+
+/// @notice Interface of recipient contract for approveAndCall pattern. This is implemented in FundingScript.sol
```
Shadowfiend: I wouldn't explicitly tie this to `FundingScript.sol`. Maybe the thing to say here is “for an example, see the `FundingScript` contract”?

implementation/contracts/system/TBTCSystem.sol
```diff
@@ -50,6 +54,17 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         relay = _relay;
     }
 
+    /// @notice        Initialize contracts 
+    /// @dev           Only the Deposit factory should call this.
```
Shadowfiend: Also it can only be called once, right?

implementation/contracts/system/TBTCSystem.sol
```diff
@@ -50,6 +54,17 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
 
+    /// @notice        Initialize contracts 
+    /// @dev           Only the Deposit factory should call this. 
+    /// @param _keepVendor        ECDSA keep vendor address. 
+    /// @param _depositFactory    Deposit Factory address. More info in DepositFactory.sol.
+    /// @param _masterDepositAddress  Master Deposit address. More info in Deposit.sol.
+    /// @param _tbtcToken         TBTCToken address. More info in TBTCToken.sol.
+    /// @param _tbtcDepositToken  TBTCDepositToken (TDT) address. More info in TBTCDepositToken.sol.
+    /// @param _feeRebateToken    FeeRebateToken (FRT) address. More info in FeeRebateToken.sol.
+    /// @param _vendingMachine    Vending Machine address. More info in VendingMachine.sol
```
Shadowfiend: Same regarding references to contracts rather than files.

implementation/contracts/system/TBTCToken.sol
```diff
@@ -4,11 +4,9 @@ import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
...
 import "./VendingMachineAuthority.sol";
 
-/**
- @dev Interface of recipient contract for approveAndCall pattern.
-*/
-interface tokenRecipient { function receiveApproval(address _from, uint256 _value, address _token, bytes calldata _extraData) external; }
-
+/// @title  TBTC Token.
+/// @notice This is the TBTC ERC20 contract.
+/// @dev    Tokens can only be minted by the Vending Machine (VendingMachine.sol).
```
Shadowfiend: Same here, “by the `VendingMachine` contract”.

implementation/contracts/system/TBTCToken.sol
```diff
@@ -30,7 +28,7 @@ contract TBTCToken is ERC20Detailed, ERC20, VendingMachineAuthority {
         return true;
     }
 
-    /// @dev             Burns an amount of the token of a given account
+    /// @dev             Burns an amount of the token of a given account.
```
Shadowfiend: I would say “from the given account's balance”, maybe? This phrasing is a little weird.

implementation/contracts/system/TBTCToken.sol
```diff
@@ -63,3 +61,6 @@ contract TBTCToken is ERC20Detailed, ERC20, VendingMachineAuthority {
         }
     }
 }
+
+/// @notice Interface of recipient contract for approveAndCall pattern. This is implemented in RedemptionScript.sol
```
Shadowfiend: Same note re: contract vs filename + not implying that's the only place it can be implemented.

As a side note, 80-character wrapping would ideally still apply here.

implementation/contracts/deposit/DepositUtils.sol
```diff
@@ -68,26 +68,26 @@ library DepositUtils {
         mapping (bytes32 => uint256) approvedDigests;
     }
 
-    /// @notice         Gets the current block difficulty
-    /// @dev            Calls the light relay and gets the current block difficulty
-    /// @return         The difficulty
+    /// @notice         Gets the current block difficulty.
+    /// @dev            Calls the light relay and gets the current block.
```
Shadowfiend: Why'd we lose the word `difficulty` here? This is the only fix I want to land before we merge.

implementation/contracts/deposit/DepositFunding.sol
NicholasDotSol: agree. I think we should remove this entirely. We verify on keep, but it's not pertinent to developers using tbtc. Might be worth adding it in-line on ECDSAKeep if at all.

implementation/contracts/system/FeeRebateToken.sol
NicholasDotSol: maybe remove file names from everywhere?

implementation/contracts/deposit/Deposit.sol
```diff
@@ -12,8 +12,8 @@ import "../system/DepositFactoryAuthority.sol";
 /// that (through various libraries) handles bitcoin funding,
 /// bitcoin-spv proofs, redemption, liquidation, and fraud logic. 
 /// @dev This is the execution context for libraries:
-/// DepositFunding, DepositLiquidaton, DepositRedemption,
-/// DepositSTates, depositUtils, OutsourceDepositLogging, and TBTCConstants.
+/// `DepositFunding`, `DepositLiquidaton`, `DepositRedemption`,
+/// `DepositSTates`, `depositUtils`, `OutsourceDepositLogging`, and `TBTCConstants`.
```
Shadowfiend: Caps are off here, States and DepositUtils.

implementation/contracts/deposit/Deposit.sol
```diff
@@ -88,11 +88,11 @@ contract Deposit is DepositFactoryAuthority {
     // THIS IS THE INIT FUNCTION
     /// @notice        The Deposit Factory can spin up a new deposit.
     /// @dev           Only the Deposit factory can call this. 
-    /// @param _TBTCSystem        TBTCSystem address. More info in VendingMachine.sol.
-    /// @param _TBTCToken         TBTCToken address. More info in TBTCToken.sol.
-    /// @param _TBTCDepositToken  TBTCDepositToken (TDT) address. More info in TBTCDepositToken.sol.
-    /// @param _FeeRebateToken    FeeRebateToken (FRT) address. More info in FeeRebateToken.sol.
-    /// @param _VendingMachine    VendingMachine address. More info in VendingMachine.sol.
+    /// @param _TBTCSystem        TBTCSystem address. More info in `VendingMachine`.
+    /// @param _TBTCToken         TBTCToken address. More info in TBTCToken`.
```
Shadowfiend: Missed the opening backtick.

implementation/contracts/deposit/DepositUtils.sol
```diff
@@ -116,7 +116,7 @@ library DepositUtils {
     }
 
     /// @notice                 Syntactically check an SPV proof for a bitcoin transaction with its hash (ID).
-    /// @dev                    Stateless SPV Proof verification documented elsewhere (see github.com/summa-tx/bitcoin-spv).
+    /// @dev                    Stateless SPV Proof verification documented elsewhere (see `github.com/summa-tx/bitcoin-spv`).
```
Shadowfiend: Don't think we need backticks here; without them, most processors of the text will link it. May even want to drop the backticks + add an `https://` before it.

implementation/contracts/external/IMedianizer.sol
```diff
@@ -1,7 +1,7 @@
 pragma solidity ^0.5.10;
 
 /// @notice A medianizer price feed.
-/// @dev Based off the MakerDAO medianizer (https://github.com/makerdao/median)
+/// @dev Based off the MakerDAO medianizer (`https://github.com/makerdao/median`)
```
Shadowfiend: Same note on the link front.

implementation/contracts/system/FeeRebateToken.sol
```diff
@@ -7,10 +7,13 @@ import "./VendingMachineAuthority.sol";
 /// @notice The Fee Rebate Token (FRT) is a non fungible token (ERC721)
 ///         the ID of which corresponds to a given deposit address. 
 ///         If the corresponding deposit is still active, ownership of this token
-///         could result in reimbursement of the singer fee payed to open the deposit.
-/// @dev    This token is minted automatically when a TDT (TBTCDepositToken.sol)
-///         is exchanged for TBTC (TBTCToken.sol) via the 
-///         Vending Machine (VendingMachine.sol).
+///         could result in reimbursement of the singer fee paid to open the deposit.
```
Shadowfiend: singer -> signer just noticed 😬

implementation/contracts/system/FeeRebateToken.sol
```diff
@@ -7,10 +7,13 @@ import "./VendingMachineAuthority.sol";
...
+/// @dev    This token is minted automatically when a TDT (TBTCDepositToken)
+///         is exchanged for TBTC (TBTCToken) via the 
+///         Vending Machine (VendingMachine).
```
Shadowfiend: These could use backticks, and probably rewrapping.

implementation/contracts/deposit/DepositFunding.sol
Shadowfiend: Agree, //cc @nkuba on whether it's necessary on the Keep side.

It *may* be worth noting here that we expect the Keep to verify it, but could see that going either way.

implementation/contracts/system/VendingMachine.sol
```diff
@@ -9,6 +9,11 @@ import {DepositUtils} from "../deposit/DepositUtils.sol";
 import "../deposit/Deposit.sol";
 import "./TBTCSystemAuthority.sol";
 
+/// @title  Vending Machine
+/// @notice The Vending Machine swaps TDTs (`TBTCDepositToken`)
```
Shadowfiend: I actually prefer to express this in terms of responsibilities, fwiw, but I'm fine with where we landed.

NicholasDotSol:
@Shadowfiend Should be good to go here

PR #465, opened 2020-02-14
Skip remainingTerm subtest that fails intermittently

### LINKED ISSUE #464: Fix `remainingTerm` off-by-one in unit test
This test fails intermittently:

`AssertionError: expected 15551997 to equal 15551998`

I suspect it's something to do with `evm_increaseTime` mining a block thereafter, which may be n+1 on non-local environments (ie. in our keep-dev testnet in CI). A fix might look like:

```js
    it('returns 0 if deposit is at term', async () => {
      const block = await web3.eth.getBlock('latest')
      await testDeposit.setUTXOInfo(prevoutValueBytes, block.timestamp, outpoint)
      const depositExpirationTime = (new BN(block.timestamp)).add(depositTerm)
      await increaseTimeTo(depositExpirationTime.toNumber())

      const remainingTerm = await testDeposit.remainingTerm.call()
      expect(remainingTerm).to.eq.BN(0)
    })
```

liamzebedee:
Adds an error margin to fix #464.


implementation/test/DepositUtilsTest.js
```diff
@@ -499,7 +499,8 @@ contract('DepositUtils', (accounts) => {
       expect(remainingTerm).to.eq.BN(expectedRemainder)
     })
 
-    it('returns 0 if deposit is at term', async () => {
+    // TODO: Fix this test intermittently failing. See https://github.com/keep-network/tbtc/issues/464
+    it.skip('returns 0 if deposit is at term', async () => {
```
Shadowfiend: Hmmm… What is this test actually testing?

In https://github.com/keep-network/tbtc/commit/6972a3b3d7db4c26863fd306e06b9410a8c1b720#diff-f25476486707efaaf257e6d1b77e013e, @NicholasDotSol fixed the other test in this group by pulling the remaining time from the block timestamp. This test I guess is meant to test that we only return 0 once we're past term? Feels like we should do `depositTerm.toNumber() + <some amount>` and then check that the remaining term is still 0, maybe.

liamzebedee: I see what you're saying - the test that @NicholasDotSol fixed doesn't test when the deposit term ends, it only tests that the value returned is depositTerm seconds into the future (with some custom remainder logic for off-by-one's). increaseTime is only called in the second.

Shadowfiend: That's the first bit of what I'm saying yeah. The second bit is that we may want to just reconsider how this test works such that it's not as subjected to these aberrations, rather than skip it (e.g. by adding a margin).

liamzebedee: Sure - I wasn't sure if we would be comfortable with that style of solution. I can add a margin, and a quick comment.

Shadowfiend:
Seminal error indeed; thanks for the midair adjustment :)

PR #467, opened 2020-02-18
implement approvedToLog check

NicholasDotSol:
closes: #477 
Checks that a caller is a Deposit address by calling `TBTCDepositToken.exists`. 

`setTbtcDepositToken` sets up the `tbtcDepositToken` contract. This setup is performed during `TBTCSystem.initialize`

Another approach could be to use a constructor and handle deployment ordering, but the Deposit is not operational until TBTCSystem.initialize has been called, and no events are expected before that call.

-----------------

Waiting on:

- [x] #481 merge.


Shadowfiend:
Also, let's file an issue here so we can flag it for the Diligence folks.

NicholasDotSol:
> My only concern here: are we sure we've covered all events by tests, to make sure we don't suddenly lose a bunch of events with this access control?

@Shadowfiend 
#481 is open. Includes testing of the last event.

implementation/contracts/DepositLog.sol
```diff
@@ -99,15 +105,26 @@ contract DepositLog {
     //
 
     /// @notice             Checks if an address is an allowed logger
-    /// @dev                Calls the system to check if the caller is a Deposit
+    /// @dev                checks tbtcDepositToken to see if the caller represents
+    ///                     an existing deposit.
     ///                     We don't require this, so deposits are not bricked if the system borks
     /// @param  _caller     The address of the calling contract
     /// @return             True if approved, otherwise false
     /* solium-disable-next-line no-empty-blocks */
```
tintinweb: now obsolete 🙌

implementation/contracts/DepositLog.sol
```diff
@@ -10,6 +12,10 @@ contract DepositLog {
       Everyone should be able to ENTIRELY rely on log messages
     */
 
+    // `TBTCDepositToken` mints a token for every new Deposit.
+    // If a token exists for a given ID, we know it is a valid Deposit address.
+    TBTCDepositToken tbtcDepositToken;
```
tintinweb: 🤔shadows `TBTCSystem.tbtcDepositToken`. will raise a `DeclarationError` with solc >= 0.6 due to re-declaration in same class.

implementation/contracts/DepositLog.sol
```diff
@@ -99,15 +105,26 @@ contract DepositLog {
...
-        return true;
+    function approvedToLog(address _caller) public view returns (bool) {
+        return tbtcDepositToken.exists(uint256(_caller));
+    }
+
+    /// @notice               Sets the tbtcDepositToken contract.
+    /// @dev                  The contract is used by `approvedToLog` to check if the
+    ///                       caller is a Deposit contract. This should only be called once.
+    /// @param  _tbtcDepositTokenAddress  The address of the tbtcDepositToken.
+    function setTbtcDepositToken(address _tbtcDepositTokenAddress) public {
```
tintinweb: this should be internal - otherwise anyone could call this before `TBTCSystem.initialize` is called. 
it is not very elegant though that this is needed in the first place, see comment on shadowed variable - the token is also available in the system contract.

tintinweb:
- the return value of log calls does not seem to be used -> might be removed completely.
- thinking about whether it would be cheaper to throw instead of checking `approvedToLog` return value if the return value of the log call is indeed never used. (it only throws if an unknown address tries to emit an event so that should be safe)

implementation/contracts/DepositLog.sol
NicholasDotSol: > this should be internal

indeed >.<

> it is not very elegant though that this is needed in the first place

Agree here, There's definitely a cleaner way to do this

Issue #469, opened 2020-02-18
Update lotsize to bond conversion


nkuba:
We currently convert lot size (sat) to bond (wei) incorrectly. We just multiply a lot size by 1,5.
e.g.
```
lotSize = 1 BTC = 100000000 sat => bondRequirement = 1,5 * 100000000 sat = 150000000 sat
150000000 wei = 0.00000000015 ETH
```
We pass sat value to request new keep and later to keep factory open new keep without calculating the correct value of an ETH bond in wei.

https://github.com/keep-network/tbtc/blob/51379d7ae488a7d97a2092c6297218e538ecd77a/implementation/contracts/deposit/DepositFunding.sol#L62-L65

PR #470, opened 2020-02-19
Bond requirement fix

NicholasDotSol:
closes: #469 

The bond requirement should be the ETH value of the `lot size` * `initialCollateralizedPercent`

extras:
- Move price feed result rejection statement to TBTCSystem. This is to avoid future reliance on individual checks. 
- Introduce  `MockBTCETHPriceRelay.sol` to set value for the price oracle if the main `TBTCSystem` contract needs to be used instead of `tbtcSystemStub` and price logic is required.


implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -60,9 +60,11 @@ library DepositFunding {
         // TODO: Whole value is stored as funder bond in the deposit, but part
         // of it should be transferred to keep: https://github.com/keep-network/tbtc/issues/297
         _d.lotSizeSatoshis = _lotSize;
-        uint256 _bondRequirement = _lotSize.mul(_system.getInitialCollateralizedPercent()).div(100);
+        uint256 _bondRequirementSatoshi = _lotSize.mul(_system.getInitialCollateralizedPercent()).div(100);
```
Shadowfiend: Time to go ahead and change `_lotSize` to specify its unit as well. Let's change it to `_lotSizeSatoshis`, and let's also go through the code and make sure anywhere we talk about denominated numbers we're including the unit in the respective variables.

PR #479, opened 2020-02-21
Tidy up the implementation README

mhluongo:
Removed old Uniswap references and a couple extra linting commands.

### DIFF
--- a/implementation/README.md
+++ b/implementation/README.md
@@ -9,17 +9,6 @@ Contains the on-chain smart contracts and tests.
 npm install
 ```
 
-#### Migrating externals
-
-tBTC interacts with external systems deployed on-chain, such as [Uniswap](https://uniswap.exchange).
-
-During development, these must be deployed to your local blockchain (ie. Ganache).
-
-```sh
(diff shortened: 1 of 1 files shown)


implementation/README.md
```diff
@@ -30,14 +19,12 @@ npm run compile
...
+npm run lint:fix
 ```
 
-Eslint errors can be disabled using a comment on the previous line. For example, 
-to disable linter errors for the 'no-unused-vars' rule: 
-`// eslint-disable-next-line no-unused-vars`.
\ No newline at end of file
+Eslint errors can be disabled using a comment on the previous line. For example,
+to disable linter errors for the 'no-unused-vars' rule:
+`// eslint-disable-next-line no-unused-vars`.
```
Shadowfiend: I would love to delete this altogether. Though occasionally necessary, lint skips are IMO not a great idea, so recommending them in the README seems like an anti-pattern.

mhluongo: Agreed, all I did was remove trailing whitespace though 😂

I'll snip it

PR #481, opened 2020-02-24
Implement remaining tests

NicholasDotSol:
No more `.skip()`
All functions and events should now be covered by tests.


Shadowfiend:
🤔 why do we keep timing out here? Something feels off. `TBTCSystem` is deploying a much-reduced system from a complexity standpoint than the deposit tests, right? Why is it taking so long?

NicholasDotSol:
> 🤔 why do we keep timing out here? Something feels off. TBTCSystem is deploying a much-reduced system from a complexity standpoint than the deposit tests, right? Why is it taking so long?

I assumed it was just random congestion + the before hooks for deploying everything can take some time.

implementation/test/DepositFraudTest.js
```diff
@@ -341,6 +344,102 @@ describe("DepositFraud", async function() {
...
+
+    it("executes and emits StartedLiquidation event", async () => {
+      const block = await web3.eth.getBlock("latest")
+
+      await testDeposit.startSignerFraudLiquidation({from: owner})
+
+      const events = await tbtcSystemStub.getPastEvents("StartedLiquidation", {
+        fromBlock: block.number,
+        toBlock: "latest",
+      })
```
Shadowfiend: In digging at truffle-contract for tbtc.js, I found that you can directly pull events from the transaction receipt:

https://github.com/keep-network/tbtc.js/blob/5e1f35a459e95ff0d8f33286f37ea0a006173d34/src/EthereumHelpers.js#L1-L25

May be worth considering this for our tests, rather than `getPastEvents`. No need to do that in this PR though, just a thought/flag. That may also allow us to use e.g. `buidler`, which would improve our dev experience more as well. It would also remove a block wait from the start of these tests. Also not sure if that applies to OpenZeppelin's setup as well.

implementation/test/DepositFraudTest.js
```diff
@@ -341,6 +344,102 @@ describe("DepositFraud", async function() {
...
+
+      expect(events[0].returnValues[0]).to.equal(testDeposit.address)
+      expect(events[0].returnValues[1]).to.be.true
+      expect(events[0].returnValues[2]).to.eq.BN(block.timestamp)
+
+      expect(initiator).to.equal(owner)
+      expect(initiated).to.eq.BN(block.timestamp)
+    })
+
+    it("completes liquidation if we came from the redemption flow", async () => {
```
Shadowfiend: Is this really, “it liquidates immediately with bonds going to the redeemer if we came from the redemption flow”?

Shadowfiend: Also… Is this correct behavior? Shouldn't we require fraud proof before liquidating for signer fraud?

implementation/test/DepositFraudTest.js
```diff
@@ -425,8 +524,6 @@ describe("DepositFraud", async function() {
         "Signature is not fraud",
       )
     })
-
-    it.skip("TODO: full test for startSignerFraudLiquidation", async () => {})
```
Shadowfiend: Aligned with the above, it seems like the two TODOs need to be filled in, not replaced with the `start` flows. We should be validating that the completion of the liquidation flows is correct, and if we're in a place where *starting* liquidation insta-liquidates that is… Incorrect? <_<

implementation/test/DepositUtilsTest.js
```diff
@@ -376,7 +382,61 @@ describe("DepositUtils", async function() {
...
+      await restoreSnapshot()
+    })
+
+    it("returns base value if no time has elapsed", async () => {
+      testDeposit.send(auctionValue, {from: accounts[0]})
+      const block = await web3.eth.getBlock("latest")
+
+      await testDeposit.setLiquidationAndCourtesyInitated(block.timestamp, 0)
+
+      const value = await testDeposit.auctionValue.call()
```
Shadowfiend: Almost feels like we need a contract helper that sets the timestamp and immediately returns the auction value here… What if a block is mined between the `setLiquidationAnd…` call above and this call?

implementation/test/DepositUtilsTest.js
```diff
@@ -376,7 +382,61 @@ describe("DepositUtils", async function() {
   })
 
   describe("auctionValue()", async () => {
-    it.skip("is TODO")
+    let duration
+    let basePercentage
+    before(async () => {
+      duration = await tbtcConstants.getAuctionDuration.call()
+      basePercentage = await tbtcConstants.getAuctionBasePercentage.call()
```
Shadowfiend: Almost would be nice to make this `baseFactor` and do the division by 100 here, instead of having to do it manually in each test case.

implementation/test/DepositUtilsTest.js
```diff
@@ -376,7 +382,61 @@ describe("DepositUtils", async function() {
...
+      const value = await testDeposit.auctionValue.call()
+      expect(value).to.eq.BN(auctionValue.mul(basePercentage).div(new BN(100)))
+    })
+
+    it("returns full value if auction Duration has elapsed ", async () => {
+      testDeposit.send(auctionValue, {from: accounts[0]})
+      const block = await web3.eth.getBlock("latest")
+
+      await testDeposit.setLiquidationAndCourtesyInitated(block.timestamp, 0)
+      await increaseTime(duration.toNumber())
```
Shadowfiend: Is this potentially subject to the issue we just fixed in #465 ? Should we add some extra factor here to avoid intermittent failure?

implementation/test/DepositUtilsTest.js
```diff
@@ -376,7 +382,61 @@ describe("DepositUtils", async function() {
...
+
+      const value = await testDeposit.auctionValue.call()
+      expect(value).to.eq.BN(auctionValue)
+    })
+
+    it("scales auction value currectly", async () => {
```
Shadowfiend: correctly*

implementation/test/DepositUtilsTest.js
```diff
@@ -376,7 +382,61 @@ describe("DepositUtils", async function() {
...
+        .sub(basePercentage)
+        .mul(elapsedTime)
+        .div(duration)
+      const percentage = basePercentage.add(elapsedPercent)
+
+      // elapse half the auction time
+      await increaseTime(elapsedTime.toNumber())
+
+      const value = await testDeposit.auctionValue.call()
+      expect(value).to.eq.BN(auctionValue.mul(percentage).div(new BN(100)))
```
Shadowfiend: I wonder if it would make sense to make this some multi-value test, e.g. 10 random increments from 0 to 100, testing that things are right along the way.

implementation/testHelpers/helpers/assertBalance.js
```diff
@@ -2,6 +2,7 @@ const BN = require('bn.js')
 const chai = require('chai')
 const expect = chai.expect
 const bnChai = require('bn-chai')
+const {web3} = require("@openzeppelin/test-environment")
```
Shadowfiend: Did this get used anywhere?

implementation/test/DepositFraudTest.js
NicholasDotSol: > Also… Is this correct behavior? Shouldn't we require fraud proof before liquidating for signer fraud?

We do in contracts. `startSignerFraudLiquidation` is only called at the end of ECDSA/SPV fraud proof.

implementation/testHelpers/helpers/assertBalance.js
NicholasDotSol: used in `increaseTime()`

implementation/test/DepositUtilsTest.js
NicholasDotSol: Maybe. I'm not sure if the benefit is worth the extra tests. Testing for half-value at half-time is a cheap way to test assuming linear scaling.

implementation/test/DepositFraudTest.js
NicholasDotSol: Are you suggesting a full test, starting from deposit cretion and reaching `startSignerFraudLiquidation`?

implementation/test/DepositUtilsTest.js
NicholasDotSol: I've seen that a grace of +1 second can be used.  think it's okay to test within that margin. ie: value is either as expected, or what would be expected given we're off by 1 second.

implementation/test/DepositUtilsTest.js
NicholasDotSol: yes, potentially. We could test margin here too

implementation/test/DepositUtilsTest.js
Shadowfiend: In general, yes, but making a habit of multi-value tests is long-term more beneficial.

That said, no need to block on it.

implementation/test/DepositFraudTest.js
Shadowfiend: I think what confused me here is that the guard to make sure this is allowed to be called happens at a higher level. Makes sense with the way that these contracts fit together though.

implementation/test/DepositFraudTest.js
Shadowfiend: We definitely need one, but not in this PR. This makes more sense to me now after your response to the other comment.

implementation/test/DepositUtilsTest.js
Shadowfiend: I'm mostly thinking about what happens if we're running tests in parallel. Realize we're not doing this right now, but I want to get there at some point as a performance optimization.

We can deal with it then, but having a helper as I described above feels like a more correct way to express the test we're trying to express here.

implementation/testHelpers/helpers/assertBalance.js
Shadowfiend: That's not called in this file, right?

implementation/test/DepositUtilsTest.js
NicholasDotSol: we do it twice, using different values. We also use `basePercentage` directly. Not sure I see why to change this

implementation/testHelpers/helpers/assertBalance.js
NicholasDotSol: no, but it's called in some test files

implementation/test/DepositFraudTest.js
NicholasDotSol: This will be easy for events that don't go through the Deposit logging. Can use openzeppelin `expectEvent`.
As you probably know from tbtc.js events that are executed outside the context of the called function don't get included in the logs and are rather encoded in rawLogs. It's less direct to retrieve them from there. Still very possible and probably better.

implementation/testHelpers/helpers/assertBalance.js
Shadowfiend: … I guess I'm asking, why are we importing it in this file?

NicholasDotSol: Tests using `increaseTime` fail with `web3 is not defined' error

PR #483, opened 2020-02-25
Use SafeMath wherever possible

NicholasDotSol:
closes: #484 
Use SafeMath instead of conventional operators.

### DIFF
--- a/implementation/contracts/deposit/Deposit.sol
+++ b/implementation/contracts/deposit/Deposit.sol
@@ -275,7 +275,7 @@ contract Deposit is DepositFactoryAuthority {
     }
 
     /// @notice                 Anyone can provide a signature that was not requested to prove fraud during funding.
-    /// @dev                    ECDSA is NOT SECURE unless you verify the digest.
+    /// @dev                    Calls out to the keep to verify if there was fraud.
     /// @param  _v              Signature recovery value.
     /// @param  _r              Signature R value.
     /// @param  _s              Signature S value.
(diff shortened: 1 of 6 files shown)


implementation/contracts/deposit/Deposit.sol
```diff
@@ -275,7 +275,7 @@ contract Deposit is DepositFactoryAuthority {
     }
 
     /// @notice                 Anyone can provide a signature that was not requested to prove fraud during funding.
-    /// @dev                    ECDSA is NOT SECURE unless you verify the digest.
```
mhluongo: I kind of like this blaring reminder everywhere, is there another spot to drop it?

Shadowfiend: As-is it's not particularly actionable anyway (what does “verify” mean?); moreover, *I believe* keep-ecdsa does the verification in question, so to both consumers and developers this isn't a relevant concern when interacting with tBTC.

Worth a double-check, of course, but this is why we removed some other instances of the warning.

Shadowfiend:
Are we worried about increasing gas costs for numbers that we already know will be safe, say because they're constants in the system or because they are governable but can have limits imposed on the governance side?

I do like safety by default, just want to make sure we've considered this.

Shadowfiend:
Also, let's capture an issue here so we can let Diligence know, in case they've got something already in their issue pile.

implementation/contracts/deposit/Deposit.sol
NicholasDotSol: yeah this was just cleaning up what was left from the original PR [comment](https://github.com/keep-network/tbtc/pull/459#discussion_r380381916)
Warning in ECDSAKeep should be good

NicholasDotSol:
> Are we worried about increasing gas costs for numbers that we already know will be safe, say because they're constants in the system or because they are governable but can have limits imposed on the governance side?

the gas cost increase is there although really small. I still stand by using SafeMath everywhere. There is less overhead if a value changes. Also constants today might be different tomorrow.

PR #486, opened 2020-02-27
Fixed external address for keep vendor

nkuba:
Due to broken line by code formatted the address was not being set correctly on contracts migrations. `sed` command we use expects one line to update.
In this PR we:
- inline the `BondedECDSAKeepVendorAddress` definition
- ignore the line for the formatter
- add logging to initialization script to catch problems when debugging
- replace sample address to invalid value so it gets rejected if not updated

### DIFF
--- a/implementation/migrations/3_initialize.js
+++ b/implementation/migrations/3_initialize.js
@@ -22,7 +22,7 @@ module.exports = async function(deployer, network) {
   if (process.env.NODE_ENV == "test" && !process.env.INTEGRATION_TEST) return
 
   const keepThreshold = 3
-  const groupSize = 3
+  const keepGroupSize = 3
 
   console.debug(
     `Initializing TBTCSystem [${TBTCSystem.address}] with:\n` +
@@ -34,7 +34,7 @@ module.exports = async function(deployer, network) {
       `  feeRebateToken: ${FeeRebateToken.address}\n` +
       `  vendingMachine: ${VendingMachine.address}\n` +
       `  keepThreshold: ${keepThreshold}\n` +
-      `  keepSize: ${groupSize}`,
+      `  keepSize: ${keepGroupSize}`,
   )
 
   // System.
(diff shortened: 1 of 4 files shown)


implementation/migrations/3_initialize.js
```diff
@@ -21,6 +21,22 @@ module.exports = async function(deployer, network) {
   // Don't enact this setup during unit testing.
   if (process.env.NODE_ENV == "test" && !process.env.INTEGRATION_TEST) return
 
+  const keepThreshold = 3
+  const groupSize = 3
```
pdyraga: If we have `keepThreshold` this should probably be `keepGroupSize`

implementation/scripts/circleci-provision-external-contracts.sh
```diff
@@ -12,7 +12,7 @@ function fetch_bonded_keep_vendor_address() {
 }
 
 function set_bonded_keep_vendor_address() {
-  sed -i -e "/BondedECDSAKeepVendorAddress/s/0x[a-fA-F0-9]\{0,40\}/${BONDED_ECDSA_KEEP_VENDOR_ADDRESS}/" ./implementation/migrations/externals.js
+  sed -i -e "/BondedECDSAKeepVendorAddress/s/0x[a-zA-Z0-9]\{0,40\}/${BONDED_ECDSA_KEEP_VENDOR_ADDRESS}/" ./implementation/migrations/externals.js
```
pdyraga: Do you think it's worth dropping a comment here why it's `a-zA-Z` instead of `a-fA-F`? Looking at this PR it's obvious, but I am not sure if this will be obvious for future-us looking just at this file.

nkuba: I would treat it as just a temporary solution. We should keep addresses in a format safer to parse and replace (e.g. `json`), so we can change `sed` and regexp to e.g. `jq` we use in this script anyways.

pdyraga: Then maybe a TODO?

nkuba:
Migrations passed in https://circleci.com/workflow-run/4de8a60d-4319-41c7-8c25-b26995884b0a

PR #487, opened 2020-02-28
tbtc-maintainers initialization container updates

nkuba:
In this PR we update tbtc-maintainers initialization script to use `MockRelay` instead of `TBTCSystem` as we expect maintainers to submit difficulty to the relay contract on test environments.

We also cleaned up the deployment configuration a little bit to align it with improvements introduced in our other projects.

Refs: https://github.com/keep-network/tbtc-maintainers/pull/23


sthompson22:
Looks like we're still creating an ETH account on deployment here.  Do we want to remove this to bring config inline with the rest of the apps?

If we don't that's fine but we'll have differing configurations between keep-dev and keep-test, since keep-test is hooked up to Ropsten and we deal exclusively  in static accounts there.

If you want me to pick up those modifications, I can.

nkuba:
Yes, I had this idea too. Please do tweak it!

nkuba:
LGTM, thanks @sthompson22 for updates. Leaving the final approval to you. Please merge at will.

Issue #473, opened 2020-02-20
Close keep when not needed or before seizing signers bonds


nkuba:
Before calling seize signer bonds keep should be closed by calling `closeKeep` function.
The close function should be also called when the keep won't be needed anymore, e.g. after redemption.

Depends on the close keep function to be exposed in the `IBondedECDSAKeep` interface in https://github.com/keep-network/keep-tecdsa/issues/237.

PR #490, opened 2020-02-29
Keep API updates: closeKeep, returnPartialSignerBonds, distributeERC20ToMembers

pdyraga:
Closes: #473 

See https://github.com/keep-network/keep-tecdsa/pull/262
See https://github.com/keep-network/keep-tecdsa/pull/263

This is WIP - we first need to merge PRs from `keep-ecdsa` but I decided to open it early as a draft to make sure we don't lose the changes from our radar.

In https://github.com/keep-network/keep-tecdsa/pull/263, we expose `closeKeep` function on `IBondedECDSAKeep` interface. This function should be called at the end of redemption flow when the keep is no longer needed and the signing group can disband.

In https://github.com/keep-network/keep-tecdsa/pull/262 we introduce `returnPartialSignerBonds`. Currently, when tBTC seizes signer bonds for liquidation, a part of them is returned using `distributeETHToMembers`. This behaviour is wrong as `distributeETHToMembers` distributes ETH to beneficiaries and ETH coming from the liquidation should be returned to the pool as an unbonded value. This is fixed in this PR and instead of using `distributeETHToMembers` we use `returnPartialSignerBonds`. 

Also, in https://github.com/keep-network/keep-tecdsa/pull/262 we rename `distributeETHToMembers` to `distributeETHReward` and `distributeERC20ToMembers` to `distributeERC20Reward` to make it clear the distributed ETH/ERC20 goes to beneficiaries as a reward. This change is reflected in this PR as a simple rename of the function called.


implementation/contracts/deposit/DepositUtils.sol
```diff
@@ -373,6 +373,7 @@ library DepositUtils {
     function seizeSignerBonds(Deposit storage _d) internal returns (uint256) {
         uint256 _preCallBalance = address(this).balance;
         IBondedECDSAKeep _keep = IBondedECDSAKeep(_d.keepAddress);
+        _keep.closeKeep();
         _keep.seizeSignerBonds();
```
tintinweb: calls `markAsClosed()` twice (closeKeep and in seizeSignerBonds)

this should be the other way round:
- first seizeSignerBonds (requires OnlyWhenActive and would bail if closeKeep is called first; there's nothing to seize after closekeep free'd the bonds)
- then closeKeep - this might bail on freeBond if a bond was completely seized before

is this covered by unit-tests?

tintinweb:
appears to be incomplete if a keep should always be closed at least when the TDT enters an end-state:

a keep should be closed when TDT reaches an end-state:

- failed_setup: 
  - notifySignerSetupFailure ✖️ does not close the keep. 
  - notifyFundingTimeout ✖️ does not close the keep. 
  - provideFundingECDSAFraudProof, ✖️slashes but does not close keep
  - provideFraudBTCFundingProof ✖️seizes funds, does not close keep
  - notifyFraudFundingTimeout  ✖️partially slashes, does not close keep
- liquidated: 
  - provideSPVFraudProof ✅removed
  - purchaseSignerBondsAtAuction  ✅ via startSignerAbortLiquidation, ✅ via startSignerFraudLiquidation (implicitly via seizebonds)
- redeemed: 
  - provideRedemptionProof  ✅

also see inline comment.

implementation/contracts/deposit/DepositUtils.sol
Shadowfiend: It is, but they're unit tests and it's really an integration test we need.

This was a slight miscommunication between the `seizeSignerBonds` update and this addition here; `seizeSignerBonds` is the right place (as this allows the keep to own its own closing in that scenario) and we're working on removing this `closeKeep` call.

Shadowfiend:
Per the above reply, we expect `seizeSignerBonds` to handle closing the keep. With that, these paths *are* covered:
- `provideFraudBTCFundingProof`: via `distributeSignerBondsToFunder` -> but this is going away as we refactor to remove fraud BTC funding proof (per y'all's suggestion).
- `notifyFraudFundingTimeout`: via `partiallySlashForFraudInFunding`
- I'm looking more closely at `provideFundingECDSAFraudProof`, though again we're refactoring this.

`notifySignerSetupFailure` will move into this state with the fix to #495 in #520, since it will start seizing signer bonds to refund the funder.

@NicholasDotSol can we capture `notifyFundingTimeout` needing a `closeKeep` call? Let's also make sure we're handling this piece correctly in the refactor @ #534.

tintinweb:
- failed_setup: 
  - notifySignerSetupFailure ✅closed by seizing funds with #520 
  - notifyFundingTimeout ✖️ does not close the keep.  --> I believe this should release signer bonds and close the keep. funder failed to provide funding proof.
  - provideFundingECDSAFraudProof, ✅slashes stake, distributes signer bonds to funder (push payment -> should be pull or funder may block), closes keep.
  - provideFraudBTCFundingProof ✅ removed with https://github.com/keep-network/tbtc/pull/534
  - notifyFraudFundingTimeout  ✅ removed with https://github.com/keep-network/tbtc/pull/534
- liquidated: 
  - provideSPVFraudProof ✅removed
  - purchaseSignerBondsAtAuction  ✅ via startSignerAbortLiquidation, ✅ via startSignerFraudLiquidation (implicitly via seizebonds)
- redeemed: 
  - provideRedemptionProof  ✅

Issue #511, opened 2020-03-03
Variable shadowing in TBTCDepositToken constructor


NicholasDotSol:
`TBTCDepositToken` inherits from `DepositFactoryAuthority`, which has a single state variable, `_depositFactory`. This variable is shadowed in the `TBTCDepositToken` constructor.

- [x] Rename the parameter or state variable.

Issue #508, opened 2020-03-03
Variable shadowing in DepositFactory


NicholasDotSol:
DepositFactory inherits from TBTCSystemAuthority. Both contracts declare a state variable with the same name, `tbtcSystem`.

- [x] Remove the shadowed variable.

NicholasDotSol:
closed by #512

PR #512, opened 2020-03-04
Fix value shadowing

NicholasDotSol:
closes: #511, #508 

Rename variables to avoid shadowing.


implementation/contracts/system/TBTCSystemAuthority.sol
```diff
@@ -5,16 +5,16 @@ pragma solidity 0.5.10;
 /// @dev    The `TBTCSystem` contract address is passed as a constructor parameter.
 contract TBTCSystemAuthority {
 
-    address internal tbtcSystem;
+    address internal tbtcSystemAddress;
 
     /// @notice Set the address of the System contract on contract initialization.
     constructor(address _tbtcSystem) public {
```
nkuba: Should we also rename `_tbtcSystem` to `_tbtcSystemAddress` to be consistent?

nkuba:
@NicholasDotSol LGTM. Can we just update this branch with master?

NicholasDotSol:
@nkuba g2g on green

Issue #510, opened 2020-03-03
Where possible, use constant rather than state variables


NicholasDotSol:
`TBTCSystem` uses a state variable for `pausedDuration`, but this value is never changed.

- [ ] Consider using the constant keyword.

PR #513, opened 2020-03-04
Make pausedDuration constant

NicholasDotSol:
closes: #510 

Pause duration will not change, therefore making this a read-only
constant makes sense


implementation/contracts/system/TBTCSystem.sol
```diff
@@ -35,7 +35,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
 
     bool _initialized = false;
     uint256 pausedTimestamp;
-    uint256 pausedDuration = 10 days;
+    uint256 constant pausedDuration = 10 days;
```
nkuba: Thoughts on moving it to `TBTCConstants` library?

NicholasDotSol: >  moving it to TBTCConstants library

Leaning towards no. It's not really a TBTCConstant, it's a System constant that will only really be used once.

nkuba:
@NicholasDotSol can we update this branch with master?

NicholasDotSol:
done

nkuba:
@NicholasDotSol we need to update it again 😢

NicholasDotSol:
@nkuba updated

Issue #492, opened 2020-03-03
RequestNewKeep ACL


NicholasDotSol:
`TBTCSystem.requestNewKeep` is used by each new Deposit contract on creation. It calls BondedECDSAKeepFactory.openKeep, which sets the Deposit contract as the “owner,” a permissioned role within the created keep. openKeep also automatically allocates bonds from members registered to the application. The “application” from which member bonds are allocated is the tbtc system itself.

Because requestNewKeep has no access controls, anyone can request that a keep be opened with msg.sender as the “owner,” and arbitrary signing threshold values

- [ ] Add access control to requestNewKeep, so that it can only be called as a part of the Deposit creation and initialization process.

PR #514, opened 2020-03-05
ACL for RequestNewKeep

NicholasDotSol:
closes: #492 492

Only `Deposit` instances should be able to call `RequestNewKeep`


implementation/test/TBTCSystemTest.js
```diff
@@ -9,30 +9,41 @@ const {BN, expectRevert} = require("@openzeppelin/test-helpers")
 const {expect} = require("chai")
 
 const TBTCSystem = contract.fromArtifact("TBTCSystem")
+const TestTBTCDepositToken = contract.fromArtifact("TestTBTCDepositToken")
 
 describe("TBTCSystem", async function() {
   let tbtcSystem
   let ecdsaKeepFactory
+  let tdt
```
nkuba: Can we name it `tbtcDepositToken`?

implementation/test/TBTCSystemTest.js
```diff
@@ -9,30 +9,41 @@ const {BN, expectRevert} = require("@openzeppelin/test-helpers")
...
+    const {
+      tbtcSystemStub,
+      ecdsaKeepFactoryStub,
+      tbtcDepositToken,
+    } = await deployAndLinkAll(
       [],
       // Though deployTestDeposit deploys a TBTCSystemStub for us, we want to
       // test TBTCSystem itself.
-      {TBTCSystemStub: TBTCSystem},
+      {TBTCSystemStub: TBTCSystem, tbtcDepositToken: TestTBTCDepositToken},
```
nkuba: Do we need to add `tbtcDepositToken: TestTBTCDepositToken` here? It's already initialized in the helper function:
https://github.com/keep-network/tbtc/blob/83559cb26805b5c94106e0a3a99a68ca5c4a4428/implementation/testHelpers/testDeployer.js#L23

implementation/test/TBTCSystemTest.js
```diff
@@ -50,6 +61,7 @@ describe("TBTCSystem", async function() {
 
       const result = await tbtcSystem.requestNewKeep.call(5, 10, 0, {
         value: openKeepFee,
+        from: accounts[0],
```
nkuba: Shouldn't it be called from `keepOwner`?

implementation/test/TBTCSystemTest.js
```diff
@@ -58,7 +70,10 @@ describe("TBTCSystem", async function() {
     it("forwards value to keep factory", async () => {
       const initialBalance = await web3.eth.getBalance(ecdsaKeepFactory.address)
 
-      await tbtcSystem.requestNewKeep(5, 10, 0, {value: openKeepFee})
+      await tbtcSystem.requestNewKeep(5, 10, 0, {
+        value: openKeepFee,
+        from: accounts[0],
```
nkuba: Shouldn't it be called from `keepOwner`?

implementation/test/TBTCSystemTest.js
```diff
@@ -9,30 +9,41 @@ const {BN, expectRevert} = require("@openzeppelin/test-helpers")
...
     let openKeepFee
+
     before(async () => {
       openKeepFee = await ecdsaKeepFactory.openKeepFeeEstimate.call()
+      await tdt.forceMint(accounts[0], web3.utils.toBN(accounts[0]))
     })
+
     it("sends caller as owner to open new keep", async () => {
       const expectedKeepOwner = accounts[2]
+      await tdt.forceMint(accounts[0], web3.utils.toBN(expectedKeepOwner))
```
nkuba: We already have it in `before` block, do we need to duplicate it here?

implementation/test/TBTCSystemTest.js
```diff
@@ -9,30 +9,41 @@ const {BN, expectRevert} = require("@openzeppelin/test-helpers")
...
     )
     // Refer to this correctly throughout the rest of the test.
     tbtcSystem = tbtcSystemStub
     ecdsaKeepFactory = ecdsaKeepFactoryStub
+    tdt = tbtcDepositToken
   })
 
   describe("requestNewKeep()", async () => {
     let openKeepFee
+
```
nkuba: We could define two constants here:
```js
const depositOwner = accounts[1]
const keepOwner = accounts[2] // deposit
```
And then in `before` block mint a token:
```js
await tdt.forceMint(depositOwner, web3.utils.toBN(keepOwner))
```

implementation/test/TBTCSystemTest.js
```diff
@@ -9,30 +9,41 @@ const {BN, expectRevert} = require("@openzeppelin/test-helpers")
...
     before(async () => {
       openKeepFee = await ecdsaKeepFactory.openKeepFeeEstimate.call()
+      await tdt.forceMint(accounts[0], web3.utils.toBN(accounts[0]))
     })
+
     it("sends caller as owner to open new keep", async () => {
       const expectedKeepOwner = accounts[2]
```
nkuba: Can we extract it to `keepOwner` as in my comment above?

implementation/test/TBTCSystemTest.js
```diff
@@ -67,6 +82,16 @@ describe("TBTCSystem", async function() {
         "TBTCSystem did not correctly forward value to keep factory",
       ).to.eq.BN(openKeepFee)
     })
+
+    it("reverts if caller does not match a valid TDT", async () => {
```
nkuba: `reverts if the caller is not a deposit`?

implementation/test/TBTCSystemTest.js
```diff
@@ -67,6 +82,16 @@ describe("TBTCSystem", async function() {
...
+      await expectRevert(
+        tbtcSystem.requestNewKeep(5, 10, 0, {
+          value: openKeepFee,
+          from: accounts[1],
```
nkuba: We should use at least two cases to validate the revert:
1. When a caller is a TDT owner.
2. When a caller is `accounts[0]`, that is neither the TDT owner nor the deposit (TDT ID).

implementation/contracts/system/TBTCSystem.sol
```diff
@@ -266,6 +268,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         payable
         returns (address)
     {
+        require(tbtcDepositToken.exists(uint256(msg.sender)), "Caller must be a Deposit contract");
```
nkuba: It's a nice solution! 👌 I'm glad you took this issue, as I'm way behind the latest changes to deposit tokens.

nkuba:
Please remember about updating this branch with master.

implementation/test/TBTCSystemTest.js
NicholasDotSol: tbtcDepositToken is in use by deployAndLinkAll

implementation/test/TBTCSystemTest.js
NicholasDotSol: The utility will replace the `tbtcDepositToken` used to initialize the `Deposit` and all related contracts. So `tbtcDepositToken` refers to `TestTBTCDepositToken`. We need this because the `tbtcDepositToken` doens't have a `forceMint`

implementation/test/TBTCSystemTest.js
NicholasDotSol: it doesn't matter here, but can change for consistency

implementation/test/TBTCSystemTest.js
NicholasDotSol: it this specific case yes, the before block was minting the token for accounts[0], but 
 expectedKeepOwner  = accounts[2] 

This will be changed because of your comment [here](https://github.com/keep-network/tbtc/pull/514#discussion_r388831194) though

implementation/test/TBTCSystemTest.js
NicholasDotSol: This is how I had it initially, but it didn't really feel accurate. We aren't really checking that it is a Deposit, we are checking that it is a TDT. Also in the test we just mint the token 🤷‍♂ 

I can go either way

implementation/test/TBTCSystemTest.js
NicholasDotSol: I'm not really clear on this, can you explain a little bit further?

> When a caller is a TDT owner

this should not revert.

NicholasDotSol: > We should use at least two cases to validate the revert:

Why? the test does not care about who holds the TDT, only if it exists.

implementation/test/TBTCSystemTest.js
nkuba: > The utility will replace the tbtcDepositToken used to initialize the Deposit and all related contracts. So tbtcDepositToken refers to TestTBTCDepositToken.

But I believe we don't need to replace it as it's already configured to `estTBTCDepositToken` [there](https://github.com/keep-network/tbtc/blob/83559cb26805b5c94106e0a3a99a68ca5c4a4428/implementation/testHelpers/testDeployer.js#L23), right?

implementation/test/TBTCSystemTest.js
nkuba: As I understand only Deposit contract can call `requestNewKeep` which is determined by TDT token ID. But TDT owner is a different account and we should fail to call `requestNewKeep` by the TDT owner. Am I incorrect?

implementation/test/TBTCSystemTest.js
nkuba: 🤷‍♂

implementation/test/TBTCSystemTest.js
```diff
@@ -9,39 +9,48 @@ const {BN, expectRevert} = require("@openzeppelin/test-helpers")
...
+    tdt = tbtcDepositToken
   })
 
   describe("requestNewKeep()", async () => {
     let openKeepFee
+    const keepOwner = accounts[2]
+
     before(async () => {
       openKeepFee = await ecdsaKeepFactory.openKeepFeeEstimate.call()
+      await tdt.forceMint(accounts[0], web3.utils.toBN(keepOwner))
```
nkuba: Can we extract `accounts[0]` to a constant as suggested in https://github.com/keep-network/tbtc/pull/514/files#r388833426. If you prefer we can name it `tdtOwner` instead of `depositOwner`, but let's make it different than `accounts[0]` which is a default account used by truffle.

We can use `tdtOwner` defined on a higher level to test if the request gets rejected when called by them as proposed in https://github.com/keep-network/tbtc/pull/514/files#r388835478

implementation/test/TBTCSystemTest.js
NicholasDotSol: ah i see, you're right then no need to substitute!

implementation/test/TBTCSystemTest.js
NicholasDotSol: you're correct, but the test will not be checking any logic. We don't crae who the owner is and there is no owner related logic. We just care that the sender is addres(tokenId)

implementation/test/TBTCSystemTest.js
NicholasDotSol: we can name it, but the address here is not significant. we just need to mint the token and need to provide an address to do so.

implementation/test/TBTCSystemTest.js
nkuba: Token owner is not used in the logic right now but it's very much related to the token and verification we perform. Having such a case can help us catching a regression bug in the future when someone updates e.g. token implementation in a way that impacts this verification.

implementation/test/TBTCSystemTest.js
NicholasDotSol: extracted TDT owner to accounts[1]

implementation/test/TBTCSystemTest.js
NicholasDotSol: Added test to check if it will revert given a caller is the owner of a valid TDT.

NicholasDotSol:
@nkuba ready here

implementation/contracts/system/TBTCSystem.sol
Shadowfiend: This is a delightfully elegant solution---I know we'd already talked about it, but it's still awesome to see it work so cleanly :)

implementation/test/TBTCSystemTest.js
NicholasDotSol: done

implementation/test/TBTCSystemTest.js
```diff
@@ -24,24 +29,28 @@ describe("TBTCSystem", async function() {
     // Refer to this correctly throughout the rest of the test.
     tbtcSystem = tbtcSystemStub
     ecdsaKeepFactory = ecdsaKeepFactoryStub
+    tdt = tbtcDepositToken
   })
 
   describe("requestNewKeep()", async () => {
     let openKeepFee
+    const keepOwner = accounts[2]
+    const tdtOwner = accounts[1]
```
nkuba: I'm sorry, the order `2`, `1` makes me anxious. Can we switch these?
```js
const tdtOwner = accounts[1]
const keepOwner = accounts[2] 
```

NicholasDotSol: done

Issue #493, opened 2020-03-03
Waiting period for system changes


NicholasDotSol:
Changing/upgrading things in the System contract without warning can lead to unpredictable function call behavior. 

- [x]  Make all upgrades require two steps with a mandatory time window between them. The first step merely broadcasts to users that a particular change is coming, and the second step commits that change after a suitable waiting period.

PR #515, opened 2020-03-05
Two-step change for governable values

NicholasDotSol:
closes: #493 

Changing/upgrading things in the System contract without warning can lead to unpredictable function call behavior.

 Make all upgrades require two steps with a mandatory time window between them. The first step merely broadcasts to users that a particular change is coming, and the second step commits that change after a suitable waiting period.

- Governance Time delay 1 hour, set arbitrarily.
- Currently, the initial update functions are unlocked. This means that if a change is incoming and is halfway through the waiting period, the initial update function can be called again, the timer resets, an event is fired and the new change is now pending.


implementation/contracts/system/TBTCSystem.sol
```diff
@@ -217,6 +308,36 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
     }
 
+    /// @notice Get the time remaining until the collateralization thresholds can be updated.
+    function getRemainingCollateralizationUpdateTime() external view returns (uint256) {
+        require(collateralizationThresholdsChangeInitiated > 0, "Update not initiated");
+
+        uint256 elapsed = block.timestamp.sub(collateralizationThresholdsChangeInitiated);
+        return (elapsed >= governanceTimeDelay)?
+        0:
+        governanceTimeDelay.sub(elapsed);
```
nkuba: Can we extract common code to an internal function?

implementation/test/TBTCSystemTest.js
```diff
@@ -69,55 +69,181 @@ describe("TBTCSystem", async function() {
     })
   })
 
-  describe("setSignerFeeDivisor", async () => {
-    it("sets the signer fee", async () => {
-      await tbtcSystem.setSignerFeeDivisor(new BN("201"))
+  describe("update Signer fee", async () => {
```
nkuba: `update Signer fee` -> `update signer fee`?

implementation/test/TBTCSystemTest.js
```diff
@@ -69,55 +69,181 @@ describe("TBTCSystem", async function() {
...
+    describe("beginSignerFeeDivisorUpdate", async () => {
+      const newFee = new BN("201")
+      it("executes and fires SignerFeeDivisorUpdateStarted event", async () => {
```
nkuba: Can we add a test case where we call begin update with different value once again after it has already been called?

implementation/test/TBTCSystemTest.js
```diff
@@ -69,55 +69,181 @@ describe("TBTCSystem", async function() {
...
 
-    it("reverts if msg.sender != owner", async () => {
-      await expectRevert.unspecified(
-        tbtcSystem.setSignerFeeDivisor(new BN("201"), {
-          from: accounts[1],
-        }),
-        "",
-      )
+    describe("finalizeSignerFeeDivisorUpdate", async () => {
+      it("reverts if the governance timer has not elapsed", async () => {
```
nkuba: In this test case we verify the case where update was not initiated. Can we rename it and add a case where delay time has not elapsed (`await increaseTime(remainingTime)`)?

implementation/test/TBTCSystemTest.js
```diff
@@ -69,55 +69,181 @@ describe("TBTCSystem", async function() {
...
+      })
+
+      it("updates signer fee and fires SignerFeeDivisorUpdated event", async () => {
+        const remainingTime = await tbtcSystem.geRemainingSignerFeeDivisorUpdateTime()
+
+        await increaseTime(remainingTime.toNumber() + 1)
+
+        receipt = await tbtcSystem.finalizeSignerFeeDivisorUpdate()
+
+        expectEvent(receipt, "SignerFeeDivisorUpdated", {
```
nkuba: We should also check the value returned by `getSignerFeeDivisor`.

implementation/test/TBTCSystemTest.js
```diff
@@ -69,55 +69,181 @@ describe("TBTCSystem", async function() {
...
+
+        receipt = await tbtcSystem.finalizeCollateralizationThresholdsUpdate()
+
+        expectEvent(receipt, "CollateralizationThresholdsUpdated", {
+          _initialCollateralizedPercent: initialPercent,
+          _undercollateralizedThresholdPercent: undercollateralizedPercent,
+          _severelyUndercollateralizedThresholdPercent: severelyUndercollateralizedPercent,
+        })
+      })
     })
```
nkuba: Can we add test cases for `getRemaining*UpdateTime`?

implementation/contracts/system/TBTCSystem.sol
```diff
@@ -192,16 +206,93 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+        external onlyOwner
+    {
+        require(
+            block.timestamp.sub(signerFeeDivisorChangeInitiated) >= governanceTimeDelay,
+            "Timer not elapsed, or change not initiated"
+        );
+        signerFeeDivisor = newSignerFeeDivisor;
+        emit SignerFeeDivisorUpdated(newSignerFeeDivisor);
+        newSignerFeeDivisor = 0;
+        signerFeeDivisorChangeInitiated = 0;
```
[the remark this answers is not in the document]
NicholasDotSol: yes, this we need

implementation/contracts/system/TBTCSystem.sol
NicholasDotSol: sure, i'm ok with that because this isn't a user-facing function so the slightly extra cost doesn't matter much. also deployment should be cheaper

implementation/test/TBTCSystemTest.js
NicholasDotSol: I think we should hold on this. 
from PR description:
```
Currently, the initial update functions are unlocked. This means that if a change is incoming and is halfway through the waiting period, the initial update function can be called again, the timer resets, an event is fired and the new change is now pending.
```

I'm not set on this functinality. Maybe we want to lock the function after an upgrade has been initiated, but there are drawbacks on both sides:

if we allow re-initiation at any time:
- function can be called all the time, emitting an event each time. 
if we don't allow re-initiation at any time:
- the risk of adding a time delay on sensitive System variables is amplified, since the time needed to change can be up to 2x timeDelay, or minimum of 2x if fixing a bad input.

implementation/test/TBTCSystemTest.js
NicholasDotSol: yes indeed :)

implementation/test/TBTCSystemTest.js
nkuba: I'm not saying that we should block subsequent calls, I just wanted to have a test case that verifies we're able to do it.

NicholasDotSol: Alright, I'm good with testing it as is for now.

implementation/contracts/system/TBTCSystem.sol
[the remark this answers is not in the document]
nkuba: @NicholasDotSol have you pushed any updates around that?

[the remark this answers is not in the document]
NicholasDotSol: yep, we have 2 separate requirements in `onlyAfterDelay` modifier now

implementation/test/TBTCSystemTest.js
```diff
@@ -103,7 +103,9 @@ describe("TBTCSystem", async function() {
       await increaseTime(totalDelay.toNumber() - expectedRemaining)
 
       const remaining = await tbtcSystem.geRemainingSignerFeeDivisorUpdateTime.call()
-      expect(remaining).to.eq.BN(expectedRemaining)
+      expect([expectedRemaining, expectedRemaining + 1]).to.include.toString(
```
nkuba: Do we need to check the range here? Why won't we get specific `expectedRemaining` value as a result?

NicholasDotSol: It's just a way to avoid potentially non-deterministic tests. If a second is incremented locally before the end of the computation/promise resolution there will be an unexpected block timestamp increase.

Issue #497, opened 2020-03-03
Unreachable LIQUIDATION_IN_PROGRESS state


NicholasDotSol:
There should be a distinction between `LIQUIDATION_IN_PROGRESS` and `FRAUD_LIQUIDATION_IN_PROGRESS`
However, the former unreachable and instead, `FRAUD_LIQUIDATION_IN_PROGRESS` is always called. This means that all non-fraud state transitions end up in the fraud liquidation path and will perform actions as if fraud was detected even though it might be caused by an undercollateralized notification or courtesy timeout.

 - [ ]  Fix the state transitions for non-fraud liquidations.

PR #517, opened 2020-03-06
Liquidation can be non-fraud

NicholasDotSol:
closes: #497

`LIQUIDATION_IN_PROGRESS` was never used.

set state to `FRAUD_LIQUIDATION IN PROGRESS` when fraud,
and `LIQUIDATION_IN_PROGRESS` when not fraud.

### DIFF
--- a/implementation/contracts/deposit/DepositLiquidation.sol
+++ b/implementation/contracts/deposit/DepositLiquidation.sol
@@ -95,7 +95,7 @@ library DepositLiquidation {
 
         _d.liquidationInitiated = block.timestamp;  // Store the timestamp for auction
         _d.liquidationInitiator = msg.sender;
-        _d.setFraudLiquidationInProgress();
+        _d.setLiquidationInProgress();
     }
 
     /// @notice                 Anyone can provide a signature that was not requested to prove fraud.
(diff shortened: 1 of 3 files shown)


implementation/contracts/deposit/DepositLiquidation.sol
```diff
@@ -95,7 +95,7 @@ library DepositLiquidation {
 
         _d.liquidationInitiated = block.timestamp;  // Store the timestamp for auction
         _d.liquidationInitiator = msg.sender;
-        _d.setFraudLiquidationInProgress();
+        _d.setLiquidationInProgress();
```
nkuba: Doesn't this change require an update in unit tests? Can we make sure we verify state in tests after calling this function (for `notifyUndercollateralizedLiquidation` and `notifyCourtesyTimeout`)?

NicholasDotSol: > Doesn't this change require an update in unit tests?

nope, we didn't verify state before.

> Can we make sure we verify state in tests after calling this function

Yes! adding state verification for `notifyUndercollateralizedLiquidation` `notifyCourtesyTimeout` `provideECDSAFraudProof` and `provideSPVFraudProof`

nkuba:
@NicholasDotSol LGTM. Can we just update this branch with master?

NicholasDotSol:
@nkuba ready here

Issue #499, opened 2020-03-03
Disallow signatures with high-s values


NicholasDotSol:
DepositRedemption.provideRedemptionSignature is used by signers to publish a signature that can be used to redeem a deposit on Bitcoin. The function accepts a signature s value in the upper half of the secp256k1 curve
Although `ecrecover` accepts signatures with these `s` values, they are no longer used in Bitcoin. As such, the signature will appear to be valid to the Ethereum smart contract, but will likely not be accepted on Bitcoin. If no users watching malleate the signature, the redemption process will likely enter a fee increase loop, incurring a cost on the deposit owner.

- [ ] Ensure the passed-in s value is restricted to the lower half of the secp256k1 curve
`_s <= 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0`

PR #518, opened 2020-03-06
Restrict _s to lower half of secp256k1 curve

NicholasDotSol:
closes: #499 

Although ecrecover accepts signatures with these s values, they are no longer used in Bitcoin.
As such, the signature will appear to be valid to the
Ethereum smart contract, but will likely not be accepted on Bitcoin.


implementation/contracts/deposit/DepositRedemption.sol
```diff
@@ -192,6 +192,15 @@ library DepositRedemption {
         // Instead, we consider this a no-harm-no-foul situation.
         // The signers have not stolen funds. Most likely they've just inconvenienced someone
 
+        // Validate `s` value for a malleability concern described in EIP-2.
```
nkuba: We should also mention it in function's docs that we restrict `_s` value.

implementation/test/DepositRedemptionTest.js
```diff
@@ -743,6 +743,15 @@ describe("DepositRedemption", async function() {
         "Invalid signature",
       )
     })
+
+    it("reverts if S value is on the upper half of the secp256k1 curve's order", async () => {
+      const s =
+        "0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A1"
```
wadeAlexC: Just a note - a better unit test would take a valid signature, flip the 's' value, check that the signature passes `ecrecover`, then submit to `provideRedemptionSignature` and expect the "Malleable signature" failure.

Issue #495, opened 2020-03-03
Funder loses payment if Keep creation fails


NicholasDotSol:
If a signing group fails to be established, payment for the keep is not returned even though one could assume that the signing group tried to play unfairly. The signing group might intentionally try to cause this scenario to interfere with the system.

- [ ] ensure that a keep group is always established or otherwise the funder is refunded the fee for the keep

refs: #495 
additional refs: #494

pdyraga:
One possible option is to seize signer bonds after the timeout, return payment to the customer from seized bonds and then return partial signer bonds to signers.

NicholasDotSol:
@pdyraga If the keep creation fails we should get a TX revert.
What might fails is the signer setup and return of the pub key. 
At any time after `createNewKeep` is called we can seise Signer bonds, right?

pdyraga:
> @pdyraga If the keep creation fails we should get a TX revert.

Yes, this is how it works.

> What might fails is the signer setup and return of the pub key.

Yes, key is submitted once generated and [there is a fixed timeout for that](https://github.com/keep-network/keep-tecdsa/blob/v0.9.0/solidity/contracts/BondedECDSAKeep.sol#L48)

> At any time after `createNewKeep` is called we can seise Signer bonds, right?

Yes. Any time after creating new keep and before closing it or seizing bonds (seizing bonds closes keep).

PR #520, opened 2020-03-12
Refund funder if signer setup fails

NicholasDotSol:
closes: #495 

Is signer setup fails and `notifySignerSetupFailure` is called successfully, the signer bonds are seized and the funder is refunded. The remainder is returned. 

- the refund is the cost to create a new keep at current price levels. This might e less than the original. As a solution, the original amount can be stored for reference or the refund amount can be a set to a value higher than an expected fee amount.


implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -65,6 +65,7 @@ library DepositFunding {
         uint256 _bondRequirementSatoshi = _lotSizeSatoshis.mul(_system.getInitialCollateralizedPercent()).div(100);
         uint256 _bondRequirementWei = _d.fetchBitcoinPrice().mul(_bondRequirementSatoshi);
 
+        _d.keepSetupFee = _system.createNewDepositFeeEstimate();
```
Shadowfiend: This feels like it should be `= msg.value`, no? That's what we actually forward to the keep, should be what we capture here also.

NicholasDotSol: Thought so initially, but `msg.value` is unbounded and that's a potential hole. It's not very realistic so if you think it's not worth covering, we record msg.value

Shadowfiend: I see… So the depositor could send a greater `msg.value` than the bond amount, and then when we try and refund them we have insufficient seized bond to do it, and things go sideways. We also give the depositor the ability to take more from signers than they are really due in the case where signer setup fails.

I think the thing we need to do here is use the lesser of `msg.sender` and `_system.createNewDepositFeeEstimate`. This protects against situations where the depositor sends more or less than the deposit fee estimate and then gets repaid. Thoughts?

NicholasDotSol: if `createNewDepositFeeEstimate ` is less than `msg.value` does the tx still go through? I'm not sure how this works on the Keep side

Shadowfiend: I believe it goes through + surplus is captured in the system to subsidize future entries.

Shadowfiend: Let's chat about this tomorrow, I think. It feels like just a minor griefing vector, since the depositor can only regain as much as they sent… It only works to specifically target signers and take bond from them, not as a way to gain any value.

NicholasDotSol: > surplus is captured in the system to subsidize future entries.

My q was with msg.value < createNewDepositFeeEstimate case, Don't know if this fails on Keep side. 

> Let's chat about this tomorrow, I think. It feels like just a minor griefing vector

Agree, I's have a bigger issue with it if the Depositor has a way to intentionally trigger it. I would still like to avoid it though

nkuba: > My q was with msg.value < createNewDepositFeeEstimate case, Don't know if this fails on Keep side.

`msg.value < openKeepFeeEstimate()` will revert on `openKeep` call.

Shadowfiend: Okay, let's aim to capture `msg.value` and use that for refunding.

nkuba:
I just noticed that my last comments were sitting in pendning state and I hadn't submitted them before. Sorry for the delay.

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -103,6 +104,14 @@ library DepositFunding {
             block.timestamp > _d.signingGroupRequestedAt.add(TBTCConstants.getSigningGroupFormationTimeout()),
             "Signing group formation timeout not yet elapsed"
         );
+
+        // refund the deposit owner the cost to create a new keep at current price levels.
```
nkuba: I believe the comment from this function docs `/// @dev We rely on the keep system to punish the signers in this case.` is no longer valid.

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -103,6 +104,14 @@ library DepositFunding {
             block.timestamp > _d.signingGroupRequestedAt.add(TBTCConstants.getSigningGroupFormationTimeout()),
             "Signing group formation timeout not yet elapsed"
         );
+
```
nkuba: Are we validating that the group formation actually failed? We should probably check if the public key is set on the keep or something. If I understand correctly right now anyone can call `notifySignerSetupFailure` function even though the key is ready but `retrieveSignerPubkey` has not been called yet.

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -103,6 +104,14 @@ library DepositFunding {
...
+        // refund the deposit owner the cost to create a new keep at current price levels.
+        uint256 _seized = _d.seizeSignerBonds();
+
+        /* solium-disable-next-line security/no-send */
+        _d.depositOwner().send(_d.keepSetupFee);
```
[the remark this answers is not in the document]
NicholasDotSol: We should be protected against reentrancy because of our state checks, so this change makes sense. I think I'll pull it to a separate PR though since there are quite a few instances and they all need to be checked against reentrancy.

implementation/contracts/deposit/DepositFunding.sol
NicholasDotSol: What is the rationale for `msg.value` over `openKeepFeeEstimate`?
It protects users from receiving less if they overfunded, but I don't think that should be at the signers' expense

implementation/contracts/deposit/DepositFunding.sol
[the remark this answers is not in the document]
NicholasDotSol: Will updated once resolved, but still a small chance it resolved the other way

implementation/contracts/deposit/DepositFunding.sol
NicholasDotSol: We could add a check:
```
bytes memory _publicKey = IBondedECDSAKeep(_d.keepAddress).getPublicKey();
require(_publicKey != 64, "public key not set or set incorrectly");
```

Worth pointing out that this slightly changes calling incentives, removing any incentive for signers to call `retrieveSignerPubkey` since they can't be punished for not setting it within the timeout. 
@Shadowfiend Thoughts?

implementation/contracts/deposit/DepositFunding.sol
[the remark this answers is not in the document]
nkuba: I believe it's gas cost optimization when switching to `call.value`

> Any smart contract that uses transfer() or send() is taking a hard dependency on gas costs by forwarding a fixed amount of gas: 2300.
>
> Our recommendation is to stop using transfer() and send() in your code and switch to using call() instead

https://diligence.consensys.net/blog/2019/09/stop-using-soliditys-transfer-now/

implementation/contracts/deposit/DepositFunding.sol
Shadowfiend: Do keeps do no punishment on DKG failure? Guessing not because it's not attributable.

implementation/contracts/deposit/DepositFunding.sol
Shadowfiend: > retrieveSignerPubkey

This method is one of my two least favorite parts of our protocol :rage4:

implementation/contracts/deposit/DepositFunding.sol
nkuba: Currently not, signers are punished on proved signature fraud.

In the case of DKG failure signers' bonds can be seized by the keep owner. Punishing the clients and exposing dedicated function for fee refund will be considered in https://github.com/keep-network/keep-ecdsa/issues/324.

implementation/contracts/deposit/DepositFunding.sol
Shadowfiend: Yeah I don't see a great way out of this.

If we refund based on the estimate, and the estimate is less than the provided value, the funder loses the spread without doing anything wrong. If everyone is behaving correctly and honestly, the value loss is small. If someone is misbehaving, the value loss is still bounded and small, since the keep estimates are part of the system rather than controlled by any of the parties in this particular interaction.

If we refund based on the value, the funder gets fully refunded, but may take more than they're really due from the signers (though notably, the signers have failed to do what they were supposed to at this point). Assuming everyone is behaving correctly and honestly, the value loss would once again be small. If someone is misbehaving, there is potentially high value loss (up to the full signer bond). The misbehavior required is sizeable, in that the funder has to also be able to prevent the signing group from forming; however, if the lot size is high enough and the bonds are correspondingly large enough, this may be enticing.

If we do the lesser of the two, we're really not getting any improvement over always refunding based on the current estimate. If we do the greater of the two, then we're not getting any improvement over refunding the funder value. So those two options aren't really worth it.

Given this, I think staying with the current fee estimate feels like the right next move. I still want to talk about this tomorrow (which we didn't do last time I said it hah) -- @NicholasDotSol do you mind making sure we have this conversation?

implementation/contracts/deposit/DepositFunding.sol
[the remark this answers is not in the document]
Shadowfiend: I talked to Martin and Alex today, there's a broader recommendation here that we heavily consider withdrawal vs transfer in several cases. Let's talk about it tomorrow.

implementation/contracts/deposit/DepositFunding.sol
NicholasDotSol: > This method is one of my two least favorite parts of our protocol :rage4:

I hear you

Shadowfiend: I'm 50-50 here. I think what we've got is likely sufficient, and am fine with a signer incentive to update this (the funder is also incentivized, since they'll be getting their funding back). This was one of the conclusions from the audit: even in imperfect cases, as long as players are properly incentivized, we're fine here.

I think this would be nice to have, but am wary of the incentive shift. But, I also don't love that it introduces a signer incentive that is a non-keep behavior other than rewards management. In particular, I don't like that fulfilling this incentive would require Go code that interacts with non-keep contracts.

Then again, I also despise that we have to actively pull the key from the keep to the deposit. I'm thinking we accept the imperfections and work to eliminate these bits in v2, though…

NicholasDotSol: I think we're on the same page here @Shadowfiend

solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -96,14 +97,21 @@ library DepositFunding {
...
         require(
             block.timestamp > _d.signingGroupRequestedAt.add(TBTCConstants.getSigningGroupFormationTimeout()),
             "Signing group formation timeout not yet elapsed"
         );
+
+        // refund the deposit owner the cost to create a new keep at current price levels.
+        uint256 _seized = _d.seizeSignerBonds();
+
+        /* solium-disable-next-line security/no-send */
+        _d.depositOwner().send(_d.keepSetupFee);
```
wadeAlexC: If possible, this should be a pull payment.

Shadowfiend: I'm going to merge this branch and have us consider the pull payment aspect holistically across the codebase next week.

implementation/contracts/deposit/DepositFunding.sol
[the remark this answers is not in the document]
NicholasDotSol: Updated

solidity/contracts/deposit/DepositFunding.sol
Shadowfiend: This landed in #570.

Issue #503, opened 2020-03-03
liquidationInitiator fallback revert can block purchaseSignerBondsAtAuction


NicholasDotSol:
during purchaseSignerBondsAtAuction, Instead of receiving a share of the funds the liquidationInitiator can decide to intentionally reject the funds by raising an exception causing `initiator.transfer(contractEthBalance)` to throw, blocking the auction and forcing the liquidation to fail. The deposit will stay in one of the `*_liquidation_in_progress` states.

- [ ] use `address.send` instead instead of `address.transfer`

Issue #504, opened 2020-03-03
purchaseSignerBondsAtAuction possible 1 wei remainder


NicholasDotSol:
purchaseSignerBondsAtAuction might leave 1 `wei` in the contract if the contract balance is odd. This is because the division will floor the values. 

- [ ] transfer the full contract balance as the last transfer action in `purchaseSignerBondsAtAuction`

NicholasDotSol:
closed with 524

PR #524, opened 2020-03-13
Block liquidationInitiator fallback revert on purchaseSignerBondsAtAuction

NicholasDotSol:
closes: #503, #504 

- `Send` does not raise a revert, so if ether is sent
to a reverting fallback the system won't halt.

- the remainder of the contract balance is cent after the first half to avoid a 1-wei remainder on odd initial balances.


implementation/contracts/deposit/DepositLiquidation.sol
```diff
@@ -256,12 +256,14 @@ library DepositLiquidation {
...
-                initiator.transfer(contractEthBalance);
+                /* solium-disable-next-line security/no-send */
+                initiator.send(contractEthBalance);
             } else {
                 // There will always be a liquidation initiator.
                 uint256 split = contractEthBalance.div(2);
                 _d.pushFundsToKeepGroup(split);
-                initiator.transfer(split);
+                /* solium-disable-next-line security/no-send */
+                initiator.send(address(this).balance);
```
Shadowfiend: No loss of functionality here so I'm good, but a couple of questions:
- Do we want to leave space for the initiator to pull their funds manually in case this fails? This could use a raw `.call()` since the initiator would be paying for their withdrawal.
- Do we want to globally disable the security/no-send rule?

No need for further action here until we figure out how we want to tackle the inherent issues with the initiator split, though.

tintinweb: As @Shadowfiend noted, the method could also bit split in to parts: (1) first part liquidating the deposit and pushing funds to keep (secure) (2) a method that allows the initiator to pull the deposit contracts balance. this would avoid a case where eth is locked in the contract because `send` fails (which should be unlikely).

PR #526, opened 2020-03-16
Restrict Deposit fallback function

### LINKED ISSUE #506: Restrict access to fallback function in Deposit.sol
Deposit.sol has an empty, payable fallback function. It is unused except when seizing signer bonds from BondedECDSAKeep.

- [ ] Have the fallback revert if the sender is not the BondedECDSAKeep. (So that Ether is not accidentally sent to a Deposit)

NicholasDotSol:
closes #506 

Deposit.sol has an empty, payable fallback function. It is unused except when seizing signer bonds from BondedECDSAKeep.

- Have the fallback revert if the sender is not the BondedECDSAKeep. (So that Ether is not accidentally sent to a Deposit)


implementation/contracts/deposit/Deposit.sol
```diff
@@ -28,7 +28,9 @@ contract Deposit is DepositFactoryAuthority {
     /* solium-disable-next-line no-empty-blocks */
     constructor () public {}
 
-    function () external payable {}
+    function () external payable {
+        require(msg.sender == self.keepAddress, "Deposit contract can only receive ETH from bondedECDSAKeep");
```
Shadowfiend: How about just “from underlying keep”? Doesn't seem like we use `bondedECDSAKeep` specifically anywhere, so it's a weird single-use term.

Won't block on this obv.

Shadowfiend:
Looks like tests failed here @NicholasDotSol -.-

Shadowfiend:
Guessing this is the new tests from #481 that need to be updated the same way other tests were updated here, maybe.

Issue #505, opened 2020-03-03
approveAndCall unused return parameter


NicholasDotSol:
`approveAndCall` always returns false because the return value bool success is never set.

- [ ] Return the correct success state.

PR #527, opened 2020-03-16
ApproveAndCall return values

NicholasDotSol:
closes: #505

`approveAndCall` always returns false because the return value bool success is never set.


implementation/test/VendingMachineTest.js
```diff
@@ -423,6 +423,15 @@ describe("VendingMachine", async function() {
...
 
+        const success = await tbtcToken.approveAndCall.call(
+          redemptionScript.address,
+          depositValue.add(signerFee),
+          calldata,
+          {from: owner},
+        )
+
+        expect(success).to.equal(true)
+
```
Shadowfiend: If we add this, do we still need the `approveAndCall` below?

NicholasDotSol: Yeah, this one is .call so we don't get any state change

Shadowfiend: So that we can observe the success, I guess…

These feel like separate tests. Any reason we've glommed them on to the existing ones rather than firing up a separate test that handles `approveAndCall` specifically?

implementation/contracts/system/TBTCToken.sol
```diff
@@ -53,12 +53,13 @@ contract TBTCToken is ERC20Detailed, ERC20, VendingMachineAuthority {
...
     /// @param _value     The max amount they can spend.
     /// @param _extraData Extra information to send to the approved contract.
-    function approveAndCall(address _spender, uint256 _value, bytes memory _extraData) public returns (bool success) {
+    function approveAndCall(address _spender, uint256 _value, bytes memory _extraData) public returns (bool) {
         tokenRecipient spender = tokenRecipient(_spender);
         if (approve(_spender, _value)) {
             spender.receiveApproval(msg.sender, _value, address(this), _extraData);
             return true;
         }
+        return false;
```
Shadowfiend: This structure seems different---do we need to introduce the same `if (approve...)` guard to the TDT?

implementation/test/VendingMachineTest.js
```diff
@@ -27,7 +27,8 @@ const _expectedUTXOoutpoint =
   "0x5f40bccf997d221cd0e9cb6564643f9808a89a5e1c65ea5e6530c0b51c18487c00000000"
 const _outValueBytes = "0x2040351d00000000"
 
-describe("VendingMachine", async function() {
+// eslint-disable-next-line no-only-tests/no-only-tests
```
Shadowfiend: Nope!

implementation/test/VendingMachineTest.js
```diff
@@ -443,6 +444,31 @@ describe("VendingMachine", async function() {
         expect(eventList[0].returnValues._digest).to.equal(sighash)
       })
 
+      it("returns true on success", async () => {
```
Shadowfiend: Couple of things here:
- This block describes `RedemptionScript`, but as far as I can see, you're not actually returning true on success for `RedemptionScript`?
- This is really a test of `TBTCToken`, not `RedemptionScript`, right?

Same note below for `TBTCDepositToken` vs `FundingScript`.

implementation/contracts/system/TBTCToken.sol
NicholasDotSol: `approve()` does not `return bool` for ERC721

implementation/test/VendingMachineTest.js
NicholasDotSol: 🤦‍♂

implementation/test/VendingMachineTest.js
NicholasDotSol: > This is really a test of TBTCToken 

Yes In a way. In order to test `approveAndCall` specifically, we would still have to go through `receiveApproval`, and this is implemented in our funding/redemption scripts. I can see how it's not ideal. 

The tests themselves are testing `approveAndCall` though.
`returns true on success` -> `approveAndCall returns true on success`?

implementation/contracts/system/TBTCToken.sol
tintinweb: 👍

implementation/test/VendingMachineTest.js
Shadowfiend: Sorry, I should have made my point clearer: this test is entirely a test of `TBTCToken`, so it feels like it should live in `TBTCTokenTest`. Similarly, the TDT one should live in `TBTCDepositTokenTest`. It feels fine to wrap these up into one `TokenTest` file or similar, but putting them in unrelated tests is not ideal.

In the interest of keeping things moving I'm going to merge this as-is, since it implements the underlying functionality we're looking for---but let's strive to make the changes quickly unless there was a good reason to have it in the current structure.

PR #528, opened 2020-03-16
testing keep-test Circle workflow

sthompson22:
There's a couple things we needed to address here.

1. When https://github.com/keep-network/tbtc/pull/485 was merged I failed to account for the job `setup_github_package_registry` being required in `migrate_contracts`.  When we merged the linked PR npm auth on contract migrations from the utility-box started failing as a result. This was observed in both `keep-dev` and `keep-test` migrations.  To address this we are now mounting the workspace that has `setup_github_package_registry` output to `migrate_contracts` and added a new job too remove the problematic `npm_modules` dir on scp to the utility-box.

2. Include git tag filters on the `compile_contracts` step.  This one is a bit perplexing but the job chain doesn't get produced unless the filter is present.

3. Updated the `keep-tecdsa` package to `0.10.0-rc`, and updated `package-lock.json` via `npm i`

This was tested against a `keep-dev` [deployment](https://app.circleci.com/pipelines/github/keep-network/tbtc/2090/workflows/d1f365f2-1c38-4aa5-8290-472d2a716e26/jobs/17070)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,9 +180,16 @@ jobs:
   migrate_contracts:
     executor: docker-thesis-buildpack
     steps:
+      - checkout
       - setup_remote_docker:
           docker_layer_caching: true
-      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Remove Installed node_module
+          working_directory: ~/project/implementation
+          command: |
+            rm -rf node_modules
       - run:
           name: Provision External Contract Addresses
           command: |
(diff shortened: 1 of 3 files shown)


sthompson22:
[here](https://app.circleci.com/pipelines/github/keep-network/tbtc/2073/workflows/9876166a-8a1a-45d7-8d5a-69de986b028d) is an example build of what happens with the current state of the `keep_test` workflow.

sthompson22:
This isn't ready yet, sorry.  Testing a proper fix against `keep-dev`, now.

.circleci/config.yml
```diff
@@ -180,9 +180,16 @@ jobs:
...
       - setup_remote_docker:
           docker_layer_caching: true
-      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Remove Installed node_module
+          working_directory: ~/project/implementation
+          command: |
+            rm -rf node_modules
```
Shadowfiend: Could maybe use a comment here?

Not a huge deal, anyway.

Issue #494, opened 2020-03-03
Enforce state transitions and incentivize calls


NicholasDotSol:
State transitions from one deposit state to another require someone calling the corresponding transition method on the deposit and actually spend gas on it. This issue assumes that participants are not always pushing forward through the state machine as soon as a new state becomes available, opening up the possibility of having multiple state transitions being a valid option for a deposit.

Ensure that there are no competing interests between participants of the system to favor one transition over the other, causing race conditions, front-running opportunities or stale deposits that are not pushed to end-states.

issues to fix: 

- [x] A TDT holder can choose not to call out notifySignerSetupFailure hoping that the signing group still forms after the signer setup timeout passes.

- ~The deposit can be pushed to active state even after `notifySignerSetupFailure`, notifyFundingTimeout have passed but nobody called it out.~ -> If both parties want to let the deposit activate later, that's their prerogative.

- ~Members of the signing group might decide to call notifyFraudFundingTimeout in a race to avoid late submissions for` provideFraudBTCFundingProof` to succeed in order to contain funds lost due to fraud.~ -> `provideFraudBTCFundingProof` is no longer a state.

- [x] A malicious signing group observes BTC funding on the bitcoin chain in an attempt to commit fraud at the time the provideBTCFundingProof transition becomes available to front-run `provideFundingECDSAFraudProof` forcing the deposit into active state.

- [x] If oracle price slippage occurs for one block (flash-crash type of event) someone could call an undercollateralization transition.

- [x] A deposit term expiration courtesy call can be exit in the rare case where `_d.fundedAt + TBTCConstants.getDepositTerm() == block.timestamp`

Shadowfiend:
@NicholasDotSol do you mind linking the PRs that have fixed the fixed issues?

Shadowfiend:
Closing this as we've handled all the ones that were a problem.

PR #529, opened 2020-03-17
Exit undercollateralized liquidation 

NicholasDotSol:
refs: https://github.com/keep-network/tbtc/issues/494

given an oracle price slip flash crash event, someone can initiate undercollateralized liquidation. This PR introduces a way to exit this liquidation flow back to the `Active` state.

- The base auction bond percentage is reduced to give time for price recovery given a flash-crash event. 
- A new State was introduced. `UNDERCOLLATERALIZED_LIQUIDATION_IN_PROGRESS`. exiting the liquidation state can only happen from this new state. Liquidation should not be exited if there a signer abort or fraud.
- `redemptionTeardown` gas savings was moved to `purchaseSignerBondsAtAuction` to avoid complications while exiting the liquidation state.  

unresolved:
- 70% auction base percentage reduced semi-arbitrarily.


Shadowfiend:
@NicholasDotSol let's pull https://github.com/keep-network/tbtc/pull/529/commits/ee4cf3e24dc83f15e40c5fee120b2e05c0e46864 out to its own PR so we can merge it. Still not convinced I want the rest of this, but I also don't want to close the PR yet… Just need to think about it some more, and don't want to block the liquidation percentage change.

Shadowfiend:
Commit in question was pulled out to #539. Holding on this PR for now.

Shadowfiend:
I'm going to close this PR without deleting the branch. We may revisit, but at this point it's highly unlikely this'll land before mainnet. We can reconsider this possibility for v2, however.

PR #531, opened 2020-03-17
v0.10.1 deploy - will close

sthompson22:
drafting for build...something is still up with the keep-test workflow, not firing properly on tag.


sthompson22:
Deployed, closing.  Will open a separate PR for the truffle config update

Issue #401, opened 2020-01-17
Integrate with keep-ecdsa v 0.10.0


nkuba:
We need to update tBTC contracts to use the latest released keep-ecdsa interfaces.

- [ ] bump up dependency
- [x] fix seize signer bonds problem with already closed keep
- [x] implement stub functions

From https://github.com/keep-network/keep-tecdsa/issues/128
- ~~[] Integrate [`seizeSignerBonds`](https://github.com/keep-network/tbtc/blob/46a38aeec76c1707279b44cd3da78d16dc8a88ed/implementation/contracts/external/IBondedECDSAKeep.sol#L25) and [`checkBondAmount`](https://github.com/keep-network/tbtc/blob/46a38aeec76c1707279b44cd3da78d16dc8a88ed/implementation/contracts/external/IBondedECDSAKeep.sol#L20) with Bonded ECDSA Keep contract functions~~ already done

From  https://github.com/keep-network/keep-tecdsa/pull/262
- ~~[] split `pushFundsToKeepGroup` to functions using `returnPartialSignerBonds` and `distributeETHToMembers`~~ - we won't use `distributeETHReward` function
- [ ] replace `distributeERC20ToMembers` with `distributeERC20Reward`

From https://github.com/keep-network/keep-tecdsa/pull/259
- ~~[x] call `closeKeep` function when done~~ - already done https://github.com/keep-network/tbtc/pull/490

PR #532, opened 2020-03-17
Bump keep-ecdsa to 0.10.0-rc and fix seize bonds problem

nkuba:
This PR contains a couple of improvements to integration with keep-ecdsa:
- fixed problem with seizing signer bonds being called after keep has already been closed (https://github.com/keep-network/tbtc/commit/f5075ec11ff1877d9fec61083e61e9539fc356fa),
- updated ECDSAKeepStub to be closer to the actual implementation of IBondedECDSAKeep interface, with this we are more likely to catch integration problems like the one above in unit tests,
- updated unit tests to reset ECDSAKeepStub used in testing,
- bumped up version of `@keep-network/keep-ecdsa` to `0.10.0-rc`.

Closes: https://github.com/keep-network/tbtc/issues/401

### DIFF
--- a/implementation/contracts/test/keep/ECDSAKeepStub.sol
+++ b/implementation/contracts/test/keep/ECDSAKeepStub.sol
@@ -4,17 +4,11 @@ import {
     IBondedECDSAKeep
 } from "@keep-network/keep-ecdsa/contracts/api/IBondedECDSAKeep.sol";
 
-import "openzeppelin-solidity/contracts/math/SafeMath.sol";
-
 /// @notice Implementation of ECDSAKeep interface used in tests only
 /// @dev This is a stub used in tests, so we don't have to call actual ECDSAKeep
 contract ECDSAKeepStub is IBondedECDSAKeep {
-    using SafeMath for uint256;
-
     bytes publicKey;
     bool success;
-    uint256 memberCount = 3;
-    bool isActive = true;
     uint256 bondAmount = 10000;
 
     // Notification that the keep was requested to sign a digest.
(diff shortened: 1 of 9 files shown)


implementation/contracts/test/keep/ECDSAKeepStub.sol
```diff
@@ -1,74 +1,96 @@
...
+    function sign(bytes32 _digest) external onlyWhenActive {
+        emit SignatureRequested(_digest);
     }
 
-    function distributeETHToMembers() external payable {
-    // solium-disable-previous-line no-empty-blocks
+    function distributeETHReward() external payable {
+        uint256 dividend = msg.value.div(memberCount);
+
+        require(dividend > 0, "Dividend value must be non-zero");
```
Shadowfiend: Hmmm… I'm 50-50 on this change. At this point this ceases to be a stub. Plus we don't end up correctly testing the interface; instead, we start integration testing the stub :/

The current situation is imperfect, but I think we should stick with what we have and work on system and integration tests out of band to catch the broader issue.

Thoughts?

nkuba: I'm not a huge fan of this solution either. I wanted to have anything to catch possible integration problems sooner than later. This is definiately a case for integration tests.

Shadowfiend: Yeah, I think I'm interested in just leaving this. We would fail to catch the next integration problem unless the change made in keep-ecdsa was made to this stub as well, which is roughly equivalent to having someone on keep-ecdsa also reviewing tbtc ecdsa-related changes, which is what we should have done in the first place when this originally happened (and which I bypassed).

I think I'd rather leave the stub as-is and pull in just the fix, tbh :/ 

Also I imagine since the diff didn't include it that this was clear, but we shouldn't bump the keep-ecdsa dependency to -rc in a PR to master :)

nkuba: > Also I imagine since the diff didn't include it that this was clear, but we shouldn't bump the keep-ecdsa dependency to -rc in a PR to master :)

We didn't release the `0.10.0` version, but we probably should - CC @pdyraga.

pdyraga: We did. Please see:
https://github.com/keep-network/keep-ecdsa/packages/29040?version=0.10.0-rc
https://github.com/keep-network/keep-ecdsa/pull/310

Shadowfiend:
Ah ah ah, you didn't update your pre-commit :grimacing:

Looks like you need to run `prettier` on the updated test files.

nkuba:
> Ah ah ah, you didn't update your pre-commit 😬

To be honest, I did. I was just scared of the number of changes from prettier and disabled it for this PR, not to cause any confusions and handle it in dedicated formatting update PR haha

> Looks like you need to run prettier on the updated test files.

I'll do that!

nkuba:
Actually, it wasn't prettier who changes so much formatting, it was something wrong with my VSCode settings. Updated the code, I'm sorry for the inconvenience 😉

implementation/contracts/test/keep/ECDSAKeepStub.sol
nkuba: We did release `0.10.0-rc` but I believe @Shadowfiend expected `0.10.0` to be used in `tbtc` dependencies. Am I correct?

Shadowfiend: We should reference `0.10.0-pre`, as here: https://github.com/keep-network/tbtc.js/blob/master/package.json#L26

However, I just realized that until I finish all my packaging stuff we may not be publishing `-pre` versions, so… Whatever works, works.

Shadowfiend:
No worries! Let's rock and roll 😎

PR #534, opened 2020-03-18
Funding fraud refactor

NicholasDotSol:
refs: #494

Members of the signing group might decide to call notifyFraudFundingTimeout in a race to avoid late submissions for provideFraudBTCFundingProof to succeed in order to contain funds lost due to fraud.

- Fraud during funding ends on `provideFundingECDSAFraudProof` with the signers
always losing their full bind as long as the funding timeout has not elapsed.


implementation/test/DepositFraudTest.js
```diff
@@ -62,9 +43,11 @@ describe("DepositFraud", async function() {
 
   describe("provideFundingECDSAFraudProof", async () => {
     let timer
+    let bond
 
     before(async () => {
       timer = await tbtcConstants.getFundingTimeout.call()
+      bond = 1000000
```
Shadowfiend: Why not make `bond` a `const` above and initialize it outside of `before`?

implementation/test/helpers/utils.js
```diff
@@ -14,16 +14,15 @@ const states = {
...
-  LIQUIDATED: new BN(12),
+  FAILED_SETUP: new BN(3),
+  ACTIVE: new BN(4),
+  AWAITING_WITHDRAWAL_SIGNATURE: new BN(5),
+  AWAITING_WITHDRAWAL_PROOF: new BN(6),
+  REDEEMED: new BN(7),
+  COURTESY_CALL: new BN(8),
+  FRAUD_LIQUIDATION_IN_PROGRESS: new BN(9),
+  LIQUIDATION_IN_PROGRESS: new BN(10),
+  LIQUIDATED: new BN(11),
```
Shadowfiend: Well this is about to require an dApp/tbtc.js update :sob:

implementation/test/DepositFraudTest.js
NicholasDotSol: changed it, no reason I can think of other than "the other one was like that" 😄

implementation/contracts/deposit/DepositFunding.sol
```diff
@@ -173,83 +164,18 @@ library DepositFunding {
...
         if (block.timestamp > _d.fundingProofTimerStart.add(TBTCConstants.getFundingTimeout())) {
+            uint256 _seized = _d.seizeSignerBonds();
+            address(0).transfer(_seized);
             _d.setFailedSetup();
+            _d.logSetupFailed();
         } else {
-            /* NB: This is reuse of the variable */
-            _d.fundingProofTimerStart = block.timestamp;
-            _d.setFraudAwaitingBTCFundingProof();
+            distributeSignerBondsToFunder(_d);
```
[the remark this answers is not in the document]
NicholasDotSol: > Thoughts?

Agree. As a note:

> f we remove the distinction, we allow the funder to gain the ETH bond even if they haven't funded the deposit, but only if they convince the funders to collude.

This shouldn't be a concern, making the change even better

tintinweb:
just dropping a quick summary of changes here:

removes state `fraud_awaiting_btc_funding_proof` and therefore transitions:
- `notifyFraudFundingTimeout`
- `provideFraudBTCFundingProof`

changes:
- `provideFundingECDSAFraudProof` slashes member stakes, seizes signer bonds and distributes them to the funder (`distributeSignerBondsToFunder`) - does this implicitly close the keep or will it stay open?
- notifyFundingTimeout closes the keep (implicitly releasing the signer bond)

Shadowfiend:
> does this implicitly close the keep

It does! Keep closing [is now](https://github.com/keep-network/keep-ecdsa/pull/259) part of `seizeSignerBonds`:

https://github.com/keep-network/keep-ecdsa/blob/050cc064dd679ce0494521df3785581aaed090d5/solidity/contracts/BondedECDSAKeep.sol#L254

solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -168,84 +160,10 @@ library DepositFunding {
...
-        );
-        require(
-            block.timestamp > _d.fundingProofTimerStart.add(TBTCConstants.getFraudFundingTimeout()),
-            "Fraud funding proof timeout has not elapsed"
-        );
-        _d.setFailedSetup();
-        _d.logSetupFailed();
-
-        partiallySlashForFraudInFunding(_d);
+        distributeSignerBondsToFunder(_d);
```
wadeAlexC: Note: funder can cause this to fail and potentially push a fraudulent deposit into an active state.

Shadowfiend: Indeed, this feeds into the overall “push vs pull” question noted in a few other places as well:
- https://github.com/keep-network/tbtc/pull/524#discussion_r392771211 (liquidation)
- https://github.com/keep-network/tbtc/pull/520#discussion_r399326416 (also funding, but a different scenario)
- I believe a couple of other places we haven't called out explicitly but that we're taking stock of at the moment.

PR #535, opened 2020-03-18
Follow the Script: Add scripts for provisioning local external contracts

liamzebedee:
`tbtc` depends on the `BondedECDSAKeepVendor` external from keep-ecdsa for deployment. Development for a newcomer involves deploying keep-ecdsa and manually copying the address into the [externals.js](https://github.com/keep-network/tbtc/blob/master/solidity/migrations/externals.js#L3) configuration.

In keep-core and keep-ecdsa, our developer experience is simpler. Running an [install script](https://github.com/keep-network/keep-ecdsa/blob/master/scripts/install.sh) automatically fetches and sets the address from the external repo. Previously I've been using some scripts at [local-tbtc-env](https://github.com/keep-network/local-tbtc-env) to do this.


Shadowfiend:
Would appreciate a more detailed PR description here on what problems we're solving and when this script would be applicable.

liamzebedee:
Good point @Shadowfiend, updated.

Shadowfiend:
@nkuba could you give this some review? I feel like you've got the best cross-cutting familiarity here.

solidity/scripts/install.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+set -ex
```
nkuba: We can remove `x` option when we're ready with the script. We won't need this additional output unless we're debugging some problem.

solidity/scripts/lcl-provision-external-contracts.sh
```diff
@@ -0,0 +1,46 @@
...
+set -e
+
+# Fetch addresses of contacts migrated from keep-network/keep-core project.
+# The `keep-core` contracts have to be migrated before running this script.
+# It requires `KEEP_ECDSA_SOL_ARTIFACTS_PATH` variable to pointing to a directory where
+# contracts artifacts after migrations are located. It also expects NETWORK_ID
+# variable to be set to the ID of the network where contract were deployed.
+# 
+# Sample command:
+# KEEP_ECDSA_SOL_ARTIFACTS_PATH=~/go/src/github.com/keep-network/keep-tecdsa/solidity/build/contracts \
```
nkuba: `keep-tecdsa` -> `keep-ecdsa`

solidity/scripts/lcl-provision-external-contracts.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -e
+
+# Fetch addresses of contacts migrated from keep-network/keep-core project.
+# The `keep-core` contracts have to be migrated before running this script.
```
nkuba: (`keep-core` -> `keep-ecdsa`) x2

solidity/scripts/install.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
nkuba: @Shadowfiend thoughts on placing this installation script under repo's root directory in `/scripts/install.sh`?

solidity/scripts/lcl-provision-external-contracts.sh
```diff
@@ -0,0 +1,46 @@
...
+  echo "Fetching value for ${BONDED_ECDSA_KEEP_VENDOR_PROPERTY}..."
+  local contractDataPath=$(realpath $KEEP_ECDSA_SOL_ARTIFACTS_PATH/$BONDED_ECDSA_KEEP_VENDOR_CONTRACT_DATA)
+  local ADDRESS=$(cat ${contractDataPath} | jq "${JSON_QUERY}" | tr -d '"')
+
+  if [[ !($ADDRESS =~ $ADDRESS_REGEXP) ]]; then
+    echo "Invalid address: ${ADDRESS}"
+    FAILED=true
+  else
+    echo "Found value for ${BONDED_ECDSA_KEEP_VENDOR_PROPERTY} = ${ADDRESS}"
+    sed -i -e "/${BONDED_ECDSA_KEEP_VENDOR_PROPERTY}/s/\"[a-zA-Z0-9]*\"/\"${ADDRESS}\"/" $DESTINATION_FILE
```
nkuba: Let's cover two types of quotation marks:
```suggestion
    sed -i -e "/${BONDED_ECDSA_KEEP_VENDOR_PROPERTY}/s/\['\"][a-zA-Z0-9]*['\"]/\"${ADDRESS}\"/" $DESTINATION_FILE
```

solidity/scripts/install.sh
Shadowfiend: I'm down if that's where it is in other repos.

nkuba: Yep, that's why I proposed to move it. Cool, cool!

Shadowfiend:
There's a thing here where we use the new package relationships to handle this for us, I think? Updating `package.json` to point to your local directory for `keep-ecdsa`, and using JS to reference the artifacts via import the way we do in e.g. tbtc.js.

liamzebedee:
So much simpler. @Shadowfiend 👍

PR #537, opened 2020-03-23
DepositLog refactor

NicholasDotSol:
refs: https://github.com/keep-network/tbtc/pull/467

- Fix `TBTCDepositToken` shadow by removing `TBTCSystem`'s copy
   - Now `TBTCSystem` uses the one inherited from `DepositLog`.
- Remove return value from logging functions.
   - Return values were not used. 
- Logging functions now revert if `approvedToLog` fails. 
   - Returning true/false allowed for non-halting logging fail. Events would not be emitted but the `Deposit` contract would continue to function. It seems that this is not a real benefit. If `approvedToLog` fails, the logging function is not called by the `Deposit` contract and it makes sense to revert.


tintinweb:
`setTbtcToken` should be removed https://github.com/keep-network/tbtc/pull/467

NicholasDotSol:
> setTbtcToken should be removed #467

@tintinweb, Not sure I follow.

Shadowfiend:
@tintinweb I assume https://github.com/keep-network/tbtc/pull/467#discussion_r394401567 is the specific bit you're referencing. I think we're not currently aiming for the bigger refactor and just looking to keep this internal---is that right @NicholasDotSol ?

NicholasDotSol:
>I think we're not currently aiming for the bigger refactor and just looking to keep this internal---is that right @NicholasDotSol ?

yeah, I don't see a reason for a larger refactor now. We successfully de-duplicate by using `DepositLog`'s copy of `tbtcDepositToken` in `TBTCSystem`.
The look and functionality seem clean to me.

Open to discussion of course

tintinweb:
Sorry I missed that conversation!

The only point I had here is that `setTbtcToken()` is not really needed because `tbtcDepositToken ` it can be set directly from `TBTCSystem` (with the appropriate checks). However, the current implementation of using an internal setter is even cleaner. 🙌 LGTM

PR #538, opened 2020-03-23
The Big Move: implementation -> solidity

Shadowfiend:
As promised. Update references in scripts and docs as well.

--------

Missing:
- [x] Passing build.
- [ ] Signoff from @sthompson22 on the infra bits.


Shadowfiend:
Those may exist locally for you because they were untracked, but they shouldn't be in the repo. Branch view tracks: https://github.com/keep-network/tbtc/tree/the-big-move

sthompson22:
> Those may exist locally for you because they were untracked, but they shouldn't be in the repo. Branch view tracks: https://github.com/keep-network/tbtc/tree/the-big-move

Gah! Thanks.

PR #539, opened 2020-03-23
Dynamic base liquidation percentage

NicholasDotSol:
Replace `AUCTION_BASE_PERCENTAGE` constant with a function that returns base percentage given the `initialCollateralizationPercent` from `tbtcSystem`

The base collateralization percentage should be the percentage of the `InitialCollateralizationPercent` that will result in a 100% bond value base auction assuming perfect collateralization.

- Note, this breaks down if the recorded collateralization percentage is below 100%

### DIFF
--- a/implementation/contracts/deposit/DepositUtils.sol
+++ b/implementation/contracts/deposit/DepositUtils.sol
@@ -33,6 +33,7 @@ library DepositUtils {
         uint256 lotSizeSatoshis;
         uint8 currentState;
         uint256 signerFeeDivisor;
+        uint128 initialCollateralizedPercent;
         uint128 undercollateralizedThresholdPercent;
         uint128 severelyUndercollateralizedThresholdPercent;
 
@@ -373,9 +374,7 @@ library DepositUtils {
     /// @return     The percentage of the InitialCollateralizationPercent that will result
     ///             in a 100% bond value base auction given perfect collateralization.
     function getAuctionBasePercentage(Deposit storage _d) internal view returns (uint256) {
-        ITBTCSystem _sys = ITBTCSystem(_d.TBTCSystem);
-        uint256 initialCollateralizedPercent = _sys.getInitialCollateralizedPercent();
-        return (10000 / initialCollateralizedPercent);
+        return uint256(10000).div(_d.initialCollateralizedPercent);
     }
 
     /// @notice     Seize the signer bond from the keep contract.
--- a/implementation/contracts/deposit/DepositUtils.sol
+++ b/implementation/contracts/deposit/DepositUtils.sol
@@ -230,7 +230,7 @@ library DepositUtils {
         }
 
         // This should make a smooth flow from base% to 100%
-        uint256 _basePercentage = TBTCConstants.getAuctionBasePercentage();
+        uint256 _basePercentage = getAuctionBasePercentage(_d);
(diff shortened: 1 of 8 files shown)


implementation/contracts/deposit/DepositUtils.sol
```diff
@@ -367,6 +367,17 @@ library DepositUtils {
...
     }
 
+
+    /// @notice     Get the starting percentage of the bond at auction.
+    /// @dev        This will return the same value regardless of collateral price.
+    /// @return     The percentage of the InitialCollateralizationPercent that will result
+    ///             in a 100% bond value base auction given perfect collateralization.
+    function getAuctionBasePercentage(Deposit storage _d) internal view returns (uint256) {
+        ITBTCSystem _sys = ITBTCSystem(_d.TBTCSystem);
+        uint256 initialCollateralizedPercent = _sys.getInitialCollateralizedPercent();
```
Shadowfiend: We have to store this at deposit creation, otherwise the existing deposit's behavior changes based on governance actions :/

implementation/test/DepositUtilsTest.js
```diff
@@ -383,12 +383,24 @@ describe("DepositUtils", async function() {
     })
   })
 
+  describe("getAuctionBasePercentage()", async () => {
+    it("returns correct base percentage", async () => {
+      const basePercentage = await testDeposit.getAuctionBasePercentage.call()
+
+      const initialCollateralization = await tbtcSystemStub.getInitialCollateralizedPercent()
+
+      // 10000 to avoid losing value to truncating is solidity.
```
Shadowfiend: in?

implementation/test/DepositUtilsTest.js
```diff
@@ -383,12 +383,24 @@ describe("DepositUtils", async function() {
...
+      const expecged = new BN(10000).div(initialCollateralization)
```
Shadowfiend: sp

implementation/contracts/deposit/DepositUtils.sol
NicholasDotSol: derp.

NicholasDotSol:
@Shadowfiend, ready here

implementation/contracts/deposit/DepositUtils.sol
```diff
@@ -33,6 +33,7 @@ library DepositUtils {
         uint256 lotSizeSatoshis;
         uint8 currentState;
         uint256 signerFeeDivisor;
+        uint128 initialCollateralizedPercent;
```
Shadowfiend: This and the two below seem like they can be smaller than `uint128`, perhaps? Could even be a `uint8`? Maybe a `uint16` if we want to allow ourselves maximum flexibility, but our percentages are expressed as percentages, so `uint8` gives us up to 250% as a representation 🤔 

Not sure if the Solidity compiler is good enough to pack things together and save us gas in that case though.

NicholasDotSol: `Uint16` for some peace of mind. Also, the enforced `initialCollateralizedPercent` max is 300, which is over the 255 uint8 limit. 
Solidity will automatically pack where possible, given the variables fit into a single 32-byte increment. two `uint128`s should be packed

Shadowfiend:
Putting a hold here as @mhluongo wants to leave <100% collateralization as an available initial percentage, so we need to see how that affects decisions here.

Shadowfiend:
Enforcement is in #541.

PR #541, opened 2020-03-24
Reduce uint sizes for better bit packing

NicholasDotSol:
Some values such as collateralization percentages and divisors don't need 256 or 128 bits of storage. This PR assigns more reasonable sizes.

These changes should reduce gas needed for Deposit creation by ~17% (75000)

Note:
` keepThreshold` and `keepSize` updated to `uint16` internally from `uint256`. 
External dependencies expecting to receive a `uint256` are not impacted.


solidity/contracts/deposit/Deposit.sol
```diff
@@ -67,7 +67,7 @@ contract Deposit is DepositFactoryAuthority {
 
     /// @notice     Get the deposit's BTC lot size in satoshi.
     /// @return     uint256 lot size in satoshi.
```
Shadowfiend: This is no longer true.

solidity/contracts/deposit/Deposit.sol
```diff
@@ -109,9 +109,9 @@ contract Deposit is DepositFactoryAuthority {
         address _TBTCDepositToken,
         address _FeeRebateToken,
         address _VendingMachine,
-        uint256 _m,
-        uint256 _n,
-        uint256 _lotSizeSatoshis
+        uint16 _m,
+        uint16 _n,
```
Shadowfiend: We do need to make sure either keep-ecdsa is updated to this, or we consider not making the change here. We may not be willing to make this change in keep-ecdsa right now.

solidity/test/TBTCSystemTest.js
```diff
@@ -329,7 +329,7 @@ describe("TBTCSystem", async function() {
         expect(receipt.logs[0].args[0][1]).to.eq.BN(lotSizes[1])
         expect([
           receipt.logs[0].args[1].toString(),
-          receipt.logs[0].args[1].toString() - 1,
+          (receipt.logs[0].args[1] - 1).toString(),
```
Shadowfiend: Oh JavaScript.

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -50,7 +51,7 @@ library DepositLiquidation {
 
         // Determine value of the lot in wei
         uint256 _satoshiPrice = _d.fetchBitcoinPrice();
-        uint256 _lotSizeSatoshis = _d.lotSizeSatoshis;
+        uint64 _lotSizeSatoshis = _d.lotSizeSatoshis;
         uint256 _lotValue = _lotSizeSatoshis.mul(_satoshiPrice);
```
Shadowfiend: Do we need to cast here?

solidity/contracts/deposit/Deposit.sol
NicholasDotSol: I have a note in the PR description. We don't need to update `keep-ecdsa`.
`TBTCSystem.requestNewKeep` calls out to `_keepFactory.openKeep`, but `requestNewKeep` still takes `uint256`.
`function requestNewKeep(uint256 _m, uint256 _n, uint256 _bond)`
`uint16` params are passed into it from `depositFunding`, but they should get implicitly casted.

solidity/contracts/deposit/DepositLiquidation.sol
NicholasDotSol: no need, it should be implicit

solidity/contracts/deposit/Deposit.sol
Shadowfiend: Sorry, I'm not worried that we'll pass bad data to Keep, but rather that it feels like we should reflect Keep's capabilities in this interface. If Keep uses uint256, we should probably expose the same; if Keep uses uint16, we should do that instead. I think having Keep use uint16 makes sense, the top limit there feels reasonable… But we should discuss.

solidity/contracts/deposit/DepositLiquidation.sol
Shadowfiend: So `_lotSizeSatoshis` will automatically be upcast so that it doesn't overflow when multiplied? Is that something `SafeMath` gives us?

solidity/contracts/deposit/Deposit.sol
NicholasDotSol: Not having divergent assumptions makes sense. The change saves about 2k gas

solidity/contracts/deposit/DepositLiquidation.sol
NicholasDotSol: `mul(uint256 a, uint256 b)`, so  `_lotSizeSatoshis` should be upcast here

solidity/contracts/deposit/Deposit.sol
Shadowfiend: We still need to work through this on the Keep side; for now, we're going to keep what's here, here. Even if Keep supports larger groups, I think it's survivable for tBTC has a smaller limitation.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -55,23 +55,23 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
     uint256 private lotSizesChangeInitiated;
     uint256 private collateralizationThresholdsChangeInitiated;
 
-    uint256 private newSignerFeeDivisor;
-    uint256[] newLotSizesSatoshis;
-    uint128 private newInitialCollateralizedPercent;
-    uint128 private newUndercollateralizedThresholdPercent;
-    uint128 private newSeverelyUndercollateralizedThresholdPercent;
+    uint16 private newSignerFeeDivisor;
+    uint64[] newLotSizesSatoshis;
```
Shadowfiend: Unrelated: should this be private?

Issue #507, opened 2020-03-03
Use specific contract type instead of address


NicholasDotSol:
Rather than storing addresses and then casting to the known contract type, it’s better to use the best type available so the compiler can check for type safety.

- [ ] Where possible, use more specific types instead of address. This goes for parameter types as well as state variable types.

PR #542, opened 2020-03-25
Use specific contract types

NicholasDotSol:
closes: #507 

Rather than storing addresses and then casting to the known contract type, it’s better to use the best type available so the compiler can check for type safety.

### DIFF
--- a/solidity/contracts/deposit/Deposit.sol
+++ b/solidity/contracts/deposit/Deposit.sol
@@ -97,11 +97,11 @@ contract Deposit is DepositFactoryAuthority {
     // THIS IS THE INIT FUNCTION
     /// @notice        The Deposit Factory can spin up a new deposit.
     /// @dev           Only the Deposit factory can call this.
-    /// @param _TBTCSystem        `TBTCSystem` contract. More info in `VendingMachine`.
-    /// @param _TBTCToken         `TBTCToken` contract. More info in TBTCToken`.
-    /// @param _TBTCDepositToken  `TBTCDepositToken` (TDT) contract. More info in `TBTCDepositToken`.
-    /// @param _FeeRebateToken    `FeeRebateToken` (FRT) contract. More info in `FeeRebateToken`.
-    /// @param _VendingMachineAddress    `VendingMachine` address. More info in `VendingMachine`.
+    /// @param _tbtcSystem        `TBTCSystem` contract. More info in `VendingMachine`.
+    /// @param _tbtcToken         `TBTCToken` contract. More info in TBTCToken`.
+    /// @param _tbtcDepositToken  `TBTCDepositToken` (TDT) contract. More info in `TBTCDepositToken`.
+    /// @param _feeRebateToken    `FeeRebateToken` (FRT) contract. More info in `FeeRebateToken`.
+    /// @param _vendingMachineAddress    `VendingMachine` address. More info in `VendingMachine`.
     /// @param _m           Signing group honesty threshold.
     /// @param _n           Signing group size.
     /// @param _lotSizeSatoshis The minimum amount of satoshi the funder is required to send.
@@ -109,20 +109,20 @@ contract Deposit is DepositFactoryAuthority {
     ///                         (10**7 satoshi == 0.1 BTC == 0.1 TBTC).
     /// @return             True if successful, otherwise revert.
     function createNewDeposit(
-        ITBTCSystem _TBTCSystem,
-        TBTCToken _TBTCToken,
-        IERC721 _TBTCDepositToken,
-        FeeRebateToken _FeeRebateToken,
-        address _VendingMachineAddress,
+        ITBTCSystem _tbtcSystem,
+        TBTCToken _tbtcToken,
+        IERC721 _tbtcDepositToken,
+        FeeRebateToken _feeRebateToken,
+        address _vendingMachineAddress,
         uint256 _m,
         uint256 _n,
         uint256 _lotSizeSatoshis
     ) public onlyFactory payable returns (bool) {
-        self.tbtcSystem = _TBTCSystem;
-        self.tbtcToken = _TBTCToken;
-        self.tbtcDepositToken = _TBTCDepositToken;
-        self.feeRebateToken = _FeeRebateToken;
-        self.VendingMachineAddress = _VendingMachineAddress;
+        self.tbtcSystem = _tbtcSystem;
+        self.tbtcToken = _tbtcToken;
+        self.tbtcDepositToken = _tbtcDepositToken;
+        self.feeRebateToken = _feeRebateToken;
+        self.vendingMachineAddress = _vendingMachineAddress;
         self.createNewDeposit(_m, _n, _lotSizeSatoshis);
         return true;
     }
--- a/solidity/contracts/deposit/Deposit.sol
+++ b/solidity/contracts/deposit/Deposit.sol
@@ -5,6 +5,11 @@ import {DepositUtils} from "./DepositUtils.sol";
 import {DepositFunding} from "./DepositFunding.sol";
 import {DepositRedemption} from "./DepositRedemption.sol";
 import {DepositStates} from "./DepositStates.sol";
+import {ITBTCSystem} from "../interfaces/ITBTCSystem.sol";
+import {IERC721} from "openzeppelin-solidity/contracts/token/ERC721/IERC721.sol";
+import {TBTCToken} from "../system/TBTCToken.sol";
+import {FeeRebateToken} from "../system/FeeRebateToken.sol";
+
 import "../system/DepositFactoryAuthority.sol";
 
 /// @title  Deposit.
@@ -113,10 +118,10 @@ contract Deposit is DepositFactoryAuthority {
         uint256 _n,
         uint256 _lotSizeSatoshis
     ) public onlyFactory payable returns (bool) {
-        self.TBTCSystem = _TBTCSystem;
-        self.TBTCToken = _TBTCToken;
-        self.TBTCDepositToken = _TBTCDepositToken;
-        self.FeeRebateToken = _FeeRebateToken;
+        self.tbtcSystem = ITBTCSystem(_TBTCSystem);
+        self.tbtcToken = TBTCToken(_TBTCToken);
+        self.tbtcDepositToken = IERC721(_TBTCDepositToken);
+        self.feeRebateToken = FeeRebateToken(_FeeRebateToken);
         self.VendingMachine = _VendingMachine;
         self.createNewDeposit(_m, _n, _lotSizeSatoshis);
         return true;
--- a/solidity/contracts/deposit/Deposit.sol
+++ b/solidity/contracts/deposit/Deposit.sol
@@ -101,7 +101,7 @@ contract Deposit is DepositFactoryAuthority {
     /// @param _TBTCToken         `TBTCToken` contract. More info in TBTCToken`.
     /// @param _TBTCDepositToken  `TBTCDepositToken` (TDT) contract. More info in `TBTCDepositToken`.
     /// @param _FeeRebateToken    `FeeRebateToken` (FRT) contract. More info in `FeeRebateToken`.
-    /// @param _VendingMachine    `VendingMachine` address. More info in `VendingMachine`.
+    /// @param _VendingMachineAddress    `VendingMachine` address. More info in `VendingMachine`.
     /// @param _m           Signing group honesty threshold.
     /// @param _n           Signing group size.
     /// @param _lotSizeSatoshis The minimum amount of satoshi the funder is required to send.
@@ -113,7 +113,7 @@ contract Deposit is DepositFactoryAuthority {
         TBTCToken _TBTCToken,
         IERC721 _TBTCDepositToken,
         FeeRebateToken _FeeRebateToken,
-        address _VendingMachine,
+        address _VendingMachineAddress,
         uint256 _m,
         uint256 _n,
         uint256 _lotSizeSatoshis
@@ -122,7 +122,7 @@ contract Deposit is DepositFactoryAuthority {
         self.tbtcToken = _TBTCToken;
         self.tbtcDepositToken = _TBTCDepositToken;
         self.feeRebateToken = _FeeRebateToken;
-        self.VendingMachine = _VendingMachine;
+        self.VendingMachineAddress = _VendingMachineAddress;
         self.createNewDeposit(_m, _n, _lotSizeSatoshis);
         return true;
     }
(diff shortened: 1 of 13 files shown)


solidity/contracts/deposit/Deposit.sol
```diff
@@ -113,10 +118,10 @@ contract Deposit is DepositFactoryAuthority {
...
         uint256 _lotSizeSatoshis
     ) public onlyFactory payable returns (bool) {
-        self.TBTCSystem = _TBTCSystem;
-        self.TBTCToken = _TBTCToken;
-        self.TBTCDepositToken = _TBTCDepositToken;
-        self.FeeRebateToken = _FeeRebateToken;
+        self.tbtcSystem = ITBTCSystem(_TBTCSystem);
+        self.tbtcToken = TBTCToken(_TBTCToken);
+        self.tbtcDepositToken = IERC721(_TBTCDepositToken);
+        self.feeRebateToken = FeeRebateToken(_FeeRebateToken);
```
Shadowfiend: Can we declare the actual function parameters to have these types?

solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -55,7 +55,7 @@ library DepositFunding {
         uint256 _n,
         uint256 _lotSizeSatoshis
     ) public returns (bool) {
-        TBTCSystem _system = TBTCSystem(_d.TBTCSystem);
+        TBTCSystem _system = TBTCSystem(address(_d.tbtcSystem));
```
Shadowfiend: Why can't we just reference `_d.tbtcSystem` here? It's already properly typed, no?

solidity/contracts/deposit/DepositRedemption.sol
```diff
@@ -31,13 +31,10 @@ library DepositRedemption {
...
     function distributeSignerFee(DepositUtils.Deposit storage _d) internal {
-        address _tbtcTokenAddress = _d.TBTCToken;
-        TBTCToken _tbtcToken = TBTCToken(_tbtcTokenAddress);
-
         IBondedECDSAKeep _keep = IBondedECDSAKeep(_d.keepAddress);
 
-        _tbtcToken.approve(_d.keepAddress, _d.signerFee());
-        _keep.distributeERC20Reward(_tbtcTokenAddress, _d.signerFee());
+        _d.tbtcToken.approve(_d.keepAddress, _d.signerFee());
+        _keep.distributeERC20Reward(address(_d.tbtcToken), _d.signerFee());
```
Shadowfiend: This feels like it should propagate to keep, insofar as `distributeERC20Reward` should take an `IERC20`?

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -354,8 +350,7 @@ library DepositUtils {
     /// @dev            We cast the address to a uint256 to match the 721 standard.
     /// @return         The current deposit beneficiary.
     function depositOwner(Deposit storage _d) public view returns (address payable) {
-        IERC721 _tbtcDepositToken = IERC721(_d.TBTCDepositToken);
-        return address(uint160(_tbtcDepositToken.ownerOf(uint256(address(this)))));
+        return address(uint160(_d.tbtcDepositToken.ownerOf(uint256(address(this)))));
```
Shadowfiend: Is there room here for an overload of `ownerAddressOf` that takes a `Deposit` and returns an `address` and does all of the weird type conversions for us as needed?

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -48,9 +48,9 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     uint256 constant pausedDuration = 10 days;
 
     TBTCDepositToken tbtcDepositToken;
-    address public keepVendor;
-    address public priceFeed;
-    address public relay;
+    IBTCETHPriceFeed priceFeed;
+    IBondedECDSAKeepVendor keepVendor;
+    IRelay relay;
```
Shadowfiend: These were declared `public` before, we should probably keep that explicit declaration? Or does leaving off `public` do something else? (Sorry, occasional Solidity gaps exist.)

solidity/contracts/deposit/Deposit.sol
NicholasDotSol: Yes we can, It doesn't really make a difference though

NicholasDotSol: Will change anyway, might look a little cleaner

solidity/contracts/deposit/DepositFunding.sol
NicholasDotSol: _d.tbtcSystem is an `ITBTCSystem` instance. It doesn't have functions like `getAllowNewDeposits()`

NicholasDotSol: It's worth adding necessary functions to the interface actually. In this PR if that's okay with you @Shadowfiend

solidity/contracts/deposit/DepositRedemption.sol
NicholasDotSol: Don't think I follow :s

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: Yeah that might be a good idea.

solidity/contracts/system/TBTCSystem.sol
NicholasDotSol: > Or does leaving off public do something else

Nope, it's just public by default. I'll add the visibility back for consistency.

solidity/contracts/deposit/DepositFunding.sol
Shadowfiend: Yes please, awesome plan!

solidity/contracts/deposit/DepositRedemption.sol
Shadowfiend: Sorry, I mean `distributeERC20Reward` should declare its first parameter as an `IERC20` rather than an `address`, and then we wouldn't need this cast, right? Obviously we'd have to make that change in keep-ecdsa as well.

solidity/contracts/deposit/DepositUtils.sol
Shadowfiend: Probably not worth cycles unless you're motivated, but yeah, seems cleaner.

solidity/contracts/proxy/DepositFactory.sol
```diff
@@ -51,9 +53,9 @@ contract DepositFactory is CloneFactory, TBTCSystemAuthority{
     ) public onlyTbtcSystem {
         masterDepositAddress = _masterDepositAddress;
         tbtcDepositToken = TBTCDepositToken(_tbtcDepositToken);
-        tbtcSystem = _tbtcSystem;
-        tbtcToken = _tbtcToken;
-        feeRebateToken = _feeRebateToken;
+        tbtcSystem = TBTCSystem(_tbtcSystem);
```
Shadowfiend: Sort of in the same vein, can we bubble this typing up to the parameters for `setExternalDependencies` as well?

solidity/contracts/proxy/DepositFactory.sol
```diff
@@ -19,9 +21,9 @@ contract DepositFactory is CloneFactory, TBTCSystemAuthority{
     // which will be used as a master contract for cloning.
     address payable public masterDepositAddress;
     TBTCDepositToken tbtcDepositToken;
-    address public tbtcSystem;
-    address public tbtcToken;
-    address public feeRebateToken;
+    TBTCSystem public tbtcSystem;
+    TBTCToken public tbtcToken;
+    FeeRebateToken public feeRebateToken;
     address public vendingMachine;
```
Shadowfiend: Should we hit this one?

solidity/contracts/interfaces/ITBTCSystem.sol
```diff
@@ -16,4 +16,10 @@ interface ITBTCSystem {
     function fetchRelayPreviousDifficulty() external view returns (uint256);
 
     function getInitialCollateralizedPercent() external view returns (uint128);
+    function getAllowNewDeposits() external view returns (bool);
+    function isAllowedLotSize(uint256 _lotSizeSatoshis) external view returns (bool);
+    function requestNewKeep(uint256 _m, uint256 _n, uint256 _bond) external payable returns (address);
+    function getSignerFeeDivisor() external view returns (uint256);
+    function getUndercollateralizedThresholdPercent() external view returns (uint128);
+    function getSeverelyUndercollateralizedThresholdPercent() external view returns (uint128);
```
Shadowfiend: Arguably our docs should be here rather than on `TBTCSystem`, but I need to see how our documentation generator is operating.

solidity/contracts/proxy/DepositFactory.sol
NicholasDotSol: We only use the address and never call any `VendingMachine` contract functions.
Hitting this would be counterproductive since we would just have to cast it back to use it.

Shadowfiend: Why do we use the address? `createNewDeposit` should likewise take a `VendingMachine`, right? Generally eschewing the type check is a bad idea, and if we have to cast to a looser type at the use point, that's fine---otherwise you're leaving on the table the capacity of the compiler to bonk you when you mishandle some point in the initialization chain.

NicholasDotSol: > Why do we use the address? 

It's just used for redemption logic:
` if(tdtHolder == vendingMachine){...}`

Worth noting, It's introduced into the system as a contract type via tbtcSystem.initialize().
Then it's passed along and stored as an address. Might be worth renaming it to `vendingMachineAddress` to avoid confusion?

solidity/contracts/deposit/DepositRedemption.sol
NicholasDotSol: I'm not sure about this. At this point we're losing any compiler advantages anyway, no?

solidity/contracts/proxy/DepositFactory.sol
Shadowfiend: I like this. Makes it clear that we're really only interested in the address.

Also makes it clear this is a dirty dirty hack and in another life where we had more time we'd try to get around it haha.

solidity/contracts/deposit/DepositRedemption.sol
Shadowfiend: Hmmm… How so? The interface would require `IERC20`, and the Keep side also implements that interface. Even though there isn't a build of the whole system where the compiler will catch stuff, the interfaces allow us to catch those same issues at compile time.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -79,41 +81,41 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+        _vendingMachine.setExternalAddresses(
             _tbtcToken,
             _tbtcDepositToken,
             _feeRebateToken
         );
-        DepositFactory(_depositFactory).setExternalDependencies(
+        _depositFactory.setExternalDependencies(
             _masterDepositAddress,
-            address(this),
+            TBTCSystem(address(this)),
```
Shadowfiend: Can this not just be `this`?

solidity/contracts/proxy/DepositFactory.sol
NicholasDotSol: ok awesome

solidity/contracts/deposit/DepositRedemption.sol
NicholasDotSol: Alright, let's change it. Needs `keep-ecdsa` PR.

solidity/contracts/system/TBTCSystem.sol
NicholasDotSol: it can, and it should

solidity/contracts/deposit/Deposit.sol
```diff
@@ -122,7 +122,7 @@ contract Deposit is DepositFactoryAuthority {
         self.tbtcToken = _TBTCToken;
         self.tbtcDepositToken = _TBTCDepositToken;
         self.feeRebateToken = _FeeRebateToken;
-        self.VendingMachine = _VendingMachine;
+        self.VendingMachineAddress = _VendingMachineAddress;
```
Shadowfiend: … Feels like all of these should start with a lowercase? And does that maybe also include all the `_feeRebateToken`, `_tbtcDepositToken`, etc?

solidity/contracts/test/deposit/TestDeposit.sol
```diff
@@ -36,13 +36,13 @@ contract TestDeposit is Deposit {
...
         address _feeRebateToken,
-        address _vendingMachine
+        address _vendingMachineAddress
     ) public {
         self.tbtcSystem = ITBTCSystem(_sys);
         self.tbtcToken = TBTCToken(_token);
         self.tbtcDepositToken = IERC721(_tbtcDepositToken);
         self.feeRebateToken = FeeRebateToken(_feeRebateToken);
-        self.VendingMachine = _vendingMachine;
+        self.VendingMachineAddress = _vendingMachineAddress;
```
Shadowfiend: Ditto on the caps.

solidity/contracts/test/keep/ECDSAKeepStub.sol
```diff
@@ -35,7 +36,7 @@ contract ECDSAKeepStub is IBondedECDSAKeep {
         // solium-disable-previous-line no-empty-blocks
     }
 
-    function distributeERC20Reward(address _asset, uint256 _value) external {
+    function distributeERC20Reward(IERC20 _asset, uint256 _value) external {
```
Shadowfiend: This won't be enough, we have to play this out over both repos. Let's do that in a separate PR.

NicholasDotSol: > This won't be enough

We just need an interface change from `keep-ecdsa`.  This should be enough for TBTC

solidity/contracts/deposit/Deposit.sol
NicholasDotSol: I can go either way, as long as we're consistent. 
Changing to lowercase for consistency

solidity/contracts/test/keep/ECDSAKeepStub.sol
NicholasDotSol: > we have to play this out over both repos

Okay with moving to separate PR to unblock

Shadowfiend: Well, the interface change will also require changing the implementors of the interface, etc. It's not going to be huge, just needs a little more finesse :)

solidity/contracts/deposit/DepositRedemption.sol
NicholasDotSol: moving to separate PR

PR #543, opened 2020-03-26
Re-introduce the introduction to tBTC in README

liamzebedee:
I somehow lost this during the MD->AsciiDoc conversion of the README.

It doesn't look like AsciiDoc supports relative links(?), so I've made it absolute for now.


Shadowfiend:
> It doesn't look like AsciiDoc supports relative links

It should. I'll have a look today.

Shadowfiend:
Ran out of time here, but my guess is you want `link:./path/to/place[link text]` for a local path. Another option if the other file is asciidoc is to use the regular cross-reference syntax `<<./path/thing.adoc,link text>>`, or you can leave off the link text if you want it inferred (can't recall if that works for cross-file references).

PR #545, opened 2020-04-01
Price feed governable addition

NicholasDotSol:
Instead of holding two feeds, ETH/USD and BTC/USD, hold an array
of direct `ETH/BTC` feeds. 

When calculating price, look for the first active price feed from the 
`ETH/BTC` array and use that one to calculate the wei/sat price. Then
use this to calculate lot size in wei where needed.

Feeds can only be added to this array by `TBTCSystem`; `TBTCSystem`
has a 90-day timer before a price feed addition can be finalized.

Also includes #576 .


solidity/contracts/price-feed/BTCETHPriceFeed.sol
```diff
@@ -11,9 +11,10 @@ contract BTCETHPriceFeed is Ownable, IBTCETHPriceFeed {
     using SafeMath for uint256;
 
     bool private _initialized = false;
+    address internal tbtcSystemAddress;
 
-    IMedianizer btcPriceFeed;
-    IMedianizer ethPriceFeed;
+    IMedianizer[] private btcUsdFeeds;
+    IMedianizer[] private ethUsdFeeds;
```
Shadowfiend, line 17: Really it's two changes here… We'll have a single direct BTCETH price feed, in addition to this way of dealing with failures.

mhluongo: Yep, we just need one array

Shadowfiend:
> Piling on 😆

but why?

solidity/migrations/prices.json
```diff
@@ -1,4 +1,3 @@
 {
-    "BTCUSD": "7152.55",
-    "ETHUSD": "142.28"
+    "BTCETH": "502709446162"
```
Shadowfiend, line 2: Is this made up or correct? Keep in mind this was originally set via https://github.com/keep-network/tbtc/blob/61822b0fcea66b4f0d3b51d4d54a7c8301f90222/solidity/scripts/update-prices.sh . May make sense to update that to work for the one value want here and set it to a recent value for completeness.

NicholasDotSol: it's correct given   `"BTCUSD": "7152.55"` and` "ETHUSD": "142.28"`
updating the script makes sense

Shadowfiend: … Is it?

```
  1 BTC     $142.28     
--------- x ------- = 0.0198 BTC / ETH = 50.270 ETH / BTC
$7152.55     1 ETH  
```

Is the issue here that BTCETH is really “ETHsats”?

Shadowfiend: No, I got my denominations wrong, sorry. Need to just do this math for a week or two instead of coming back to it once every 3 months hah. This is “weiBTC”, yah?

NicholasDotSol: yeah. The math is correct, it's just `x 10^10`

NicholasDotSol: Wait, sorry. should be `weiSat`, not `weiBtc`. 
`x 10^10` is the result of simplifying `x 10^8`(for sat) and `x 10^18` (for wei)

Shadowfiend: Ok. I'm double-checking to see if we'll have weiBTC, weisats, or something else.

liamzebedee: I too made this "units conversion" mistake in https://github.com/keep-network/tbtc/pull/393. I'd definitely like to avoid the confusion where we can. Maybe add a doc if we're changing from decimal point ETH to wei?

Shadowfiend: I would love for this to be referred to as `weisat` if that's what we've got, rather than `ETHBTC` (or `BTCETH` which is the wrong way around).

I haven't heard back from Maker yet on the denomination they've chosen to use.

NicholasDotSol: > `BTCETH` which is the wrong way around

Why is this the wrong way around?

shouldn't `BTCETH` return `BTC` value in terms of `ETH`
same way `ETHUSD` returns `ETH` in terms of `USD`?

Shadowfiend: Oh hey you're right. I keep reading these as “ETH per USD” rather than “ETH in USD”. I'll unbreak my brain, my bad.

Looks like the [Kovan ETHBTC Medianizer](https://kovan.etherscan.io/address/0xf60df9b138a00ae8dbd07f55fd2305cc791e68cc#events) is publishing BTC in wei. So, quick take: `weiBTC` is what we call our thing, adjust accordingly throughout.

NicholasDotSol: unfortunately, we'll have a naming split between weiBTC and BTCwei. since we're given one and need the other

NicholasDotSol: `satwei` actually

Shadowfiend: So, we need weisat (number of wei in 1 sat), since lot size is in sats and we'll want bond amount in wei, right? And the feed will give us BTCwei. And `10**26/BTCwei` will give us `weisat`, ya?

Shadowfiend: I'm a dumb and got this backwards again. Feed is ETHBTC, correctly---BTC in 1 ETH, 18 decimals. We want satwei---wei in 1 sat, also 18 decimals. `10**26/ETHBTC` gives `satwei`.

solidity/contracts/price-feed/SatWeiPriceFeed.sol
```diff
@@ -0,0 +1,88 @@
+pragma solidity 0.5.17;
+
+import {SafeMath} from "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "../external/IMedianizer.sol";
+import "../interfaces/ISatWeiPriceFeed.sol";
+
+/// @notice Bitcoin-Ether price feed.
```
Shadowfiend, line 8: I mean :grimacing:

solidity/contracts/price-feed/SatWeiPriceFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+/// @dev Based on the ratio of BTC/USD and ETH/USD.
```
Shadowfiend, line 9: 'Tain't though.

solidity/contracts/price-feed/SatWeiPriceFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+    bool private _initialized = false;
+    address internal tbtcSystemAddress;
+
+    IMedianizer[] private ethBtcFeeds;
+
+    constructor() public {
+    // solium-disable-previous-line no-empty-blocks
+    }
+
+    /// @notice Initialises the addresses of the BTC/ETH price feeds.
```
Shadowfiend, line 22: Does itttttt

solidity/contracts/price-feed/SatWeiPriceFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+    /// @param _ETHBTCPriceFeed The ETHBTC price feed address.
```
Shadowfiend, line 23: Got an undoced param here.

solidity/contracts/price-feed/SatWeiPriceFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+    )
+        external onlyOwner
+    {
+        require(!_initialized, "Already initialized.");
+        tbtcSystemAddress = _tbtcSystemAddress;
+        ethBtcFeeds.push(_ETHBTCPriceFeed);
+        _initialized = true;
+    }
+
+    /// @notice Get the current price of satoshi in wei.
```
Shadowfiend, line 36: Worth explicitly saying “1 satoshi” yeah?

solidity/contracts/price-feed/SatWeiPriceFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+        bool ethBtcActive;
+
+        for(uint i = 0; i < ethBtcFeeds.length; i++){
+            (, ethBtcActive) = ethBtcFeeds[i].peek();
+            if(ethBtcActive) {
+                return address(ethBtcFeeds[i]);
+            }
+        }
+        return address(0);
+    }
```
Shadowfiend, line 72: Why not use this from `getPrice`?

solidity/contracts/interfaces/ISatWeiPriceFeed.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity 0.5.17;
+
+import "../external/IMedianizer.sol";
+
+interface ISatWeiPriceFeed {
```
Shadowfiend, line 5: An `@notice` wouldn't go amiss here.

solidity/contracts/interfaces/ISatWeiPriceFeed.sol
```diff
@@ -0,0 +1,13 @@
...
+    /// @notice Get the current price of satoshi in wei.
```
Shadowfiend, line 6: Worth `1 satoshi` explicitly?

docs/introduction-to-tbtc.md
```diff
@@ -87,7 +87,7 @@ The redemption flow is as follows:
 
 The "happy paths" have been covered, but there hasn't been discussion of when *things fall apart*. 
 
-To disincentivise signers from stealing bitcoin, deposits are overcollateralised with ETH. The collateral is priced using an on-chain [ETH:BTC price feed](https://github.com/keep-network/tbtc/blob/master/solidity/contracts/price-feed/BTCETHPriceFeed.sol), which is operated by MakerDAO. This collateral also ensures  guarantees of signer availability within the protocol.
+To disincentivise signers from stealing bitcoin, deposits are overcollateralised with ETH. The collateral is priced using an on-chain [ETH:BTC price feed](https://github.com/keep-network/tbtc/blob/master/solidity/contracts/price-feed/SatWeiPriceFeed.sol), which is operated by MakerDAO. This collateral also ensures  guarantees of signer availability within the protocol.
```
Shadowfiend, line 90: Feels awkward that we're linking to the thing we built, but saying it's operated by Maker. May want to point to IMedianizer, or follow the asciidoc and point to the Maker OSM docs.

solidity/contracts/price-feed/SatWeiPriceFeed.sol
NicholasDotSol: because I'd like to keep this function `external`

docs/introduction-to-tbtc.md
NicholasDotSol: I like pointing at our think here, what if we change the phrasing:

> The collateral is priced using an on-chain [ETH:BTC price feed](https://github.com/keep-network/tbtc/blob/master/solidity/contracts/price-feed/SatWeiPriceFeed.sol), which works using price-feeds operated by MakerDAO.

solidity/contracts/price-feed/SatWeiPriceFeed.sol
Shadowfiend: Feels like we should put an internal shell in place and reuse it from the external call and from `getPrice` tbh…

mhluongo:
Can we get an updated / accurate PR description before merge? Dismissing my review for now, happy to take a second look when ready

solidity/contracts/price-feed/SatWeiPriceFeed.sol
NicholasDotSol: we could do that, but the functionality is not identical. For one we don't need the price value in the `getWorkingEthBtcFeed` call, we would need more code if we keep the internal function small.

Shadowfiend:
Description is accurate already as far as I can tell.

docs/introduction-to-tbtc.md
NicholasDotSol: note: `price-feeds` -> `a price feed`

solidity/contracts/price-feed/SatWeiPriceFeed.sol
Shadowfiend: Mmm… Fair, fair.

PR #547, opened 2020-04-03
Update default value transfer method

NicholasDotSol:
`.call.value()` should not revert in case of failure, but there is also no directly enforced gas limit.
It is important to ensure against reentrance for all value-handling functions.

- `send` also does not revert in case of failure, and there is an enforced gas limit that should limit reentrancy vectors, but it is not entirely reentrancy-secure due to the proven non-constant nature of op-code gas costs. Therefore instead of relying on `send` to restrict reentrancy, we explicitly ensure our contract is non-reentrant and allow a user's fallback to perform calculations beyond `send`'s limit since there is arguable no reason to enforce the gas limit.

### DIFF
--- a/solidity/contracts/deposit/DepositFunding.sol
+++ b/solidity/contracts/deposit/DepositFunding.sol
@@ -82,7 +82,7 @@ library DepositFunding {
     /// @dev        This is only called as part of funding fraud flow.
     function distributeSignerBondsToFunder(DepositUtils.Deposit storage _d) internal {
         uint256 _seized = _d.seizeSignerBonds();
-        _d.depositOwner().transfer(_seized);  // Transfer whole amount
+        _d.depositOwner().call.value(_seized)("");  // Transfer whole amount
     }
 
     /// @notice     Anyone may notify the contract that signing group setup has timed out.
@@ -97,12 +97,12 @@ library DepositFunding {
         // refund the deposit owner the cost to create a new Deposit at the time the Deposit was opened.
         uint256 _seized = _d.seizeSignerBonds();
 
-        /* solium-disable-next-line security/no-send */
-        _d.depositOwner().send(_d.keepSetupFee);
-        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
-
         _d.setFailedSetup();
         _d.logSetupFailed();
+
+        /* solium-disable-next-line security/no-send */
+        _d.depositOwner().call.value(_d.keepSetupFee)("");
+        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
 
         fundingTeardown(_d);
     }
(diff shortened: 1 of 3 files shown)


solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -97,13 +97,13 @@ library DepositFunding {
...
-        _d.depositOwner().send(_d.keepSetupFee);
-        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
-
         _d.setFailedSetup();
         _d.logSetupFailed();
 
+        /* solium-disable-next-line security/no-send */
+        _d.depositOwner().call.value(_d.keepSetupFee)("");
+        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
+
```
Shadowfiend, line 106: We may want to leave a note as to why we're doing the state change in `fundingTeardown` after we do the above transfers.

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -74,7 +74,7 @@ library DepositLiquidation {
         if (_d.auctionTBTCAmount() == 0) {
             // we came from the redemption flow
             _d.setLiquidated();
-            _d.redeemerAddress.transfer(_seized);
+            _d.redeemerAddress.call.value(_seized)("");
```
Shadowfiend, line 77: Feels like we should do this at the end of the function, per checks-effects-interactions, no?

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -180,7 +180,7 @@ library DepositLiquidation {
 
         // Distribute funds to auction buyer
         uint256 _valueToDistribute = _d.auctionValue();
-        msg.sender.transfer(_valueToDistribute);
+        msg.sender.call.value(_valueToDistribute)("");
```
Shadowfiend, line 183: As above re: checks-effects-interactions.

solidity/contracts/deposit/DepositLiquidation.sol
NicholasDotSol: we `setLiquidated` before transferring, preventing reentrancy. 

>  at the end of the function

Assuming you mean above `setFraudLiquidationInProgress`. I'm fine with that, although the values are unrelated.

solidity/contracts/deposit/DepositFunding.sol
NicholasDotSol: there's no reason here. We can do it after as long as we ensure state change comes first.

solidity/contracts/deposit/DepositLiquidation.sol
NicholasDotSol: `_d.setLiquidated();` is at the top of the function, preventing reentrancy.

Not really much more to move up aside from the initiator check.

NicholasDotSol:
Closed in favor of #570
moving to a pull model

PR #565, opened 2020-04-06
Solid Ice: Freeze Solidity version for mainnet release

### LINKED ISSUE #558: Use OpenZeppelin SafeMath instead of bitcoin-spv SafeMath
Future versions will let us access any improvements or fixes made in OpenZeppelin.

### LINKED ISSUE #559: Share tokenRecipient interface used in token contracts
Right now `tokenRecipient` is defined in two contracts; we should share that interface to minimize potential for unintended future drift.

### LINKED ISSUE #560: Pin solidity version to a final version for mainnet
Let's take the latest 0.5 version of Solidity and pin to exactly that version for mainnet release.

Shadowfiend:
Nothing too crazy here, all pragmas are bumped to Solidity 0.5.15,
and the dependency is made fixed rather than floating with `^`.

Bonus in this PR: our SafeMath imports were split between OpenZeppelin
and bitcoin-spv; now, they are all OpenZeppelin's version.

And as a bonus bonus, a double encore if you will… A few dependencies
were either outdated, needed to be moved to `devDependencies`, or
were in both `dependencies` and `devDependencies` and only needed
to be in the latter. These are fixed.

Triple encore because why not, `tokenRecipient` from the `TBTCToken`
and `TBTCDepositToken` contracts is pulled into a common
`interfaces/ITokenRecipient.sol`.

Closes #560, closes #558, closes #559.


solidity/contracts/interfaces/ITokenRecipient.sol
```diff
@@ -0,0 +1,37 @@
+pragma solidity 0.5.15;
+
+/// @title Interface of recipient contract for `approveAndCall` pattern.
+///        Implementors will be able to be used in an `approveAndCall`
+///        interaction with a supporting contract, such that a token approval
+///        can call the contract acting on that approval in a single
+///        transaction.
+///
+///        See the `FundingScript` and `RedemptionScript` contracts as examples.
+interface ITokenRecipient {
```
NicholasDotSol, line 10: yes please

solidity/contracts/system/VendingMachine.sol
```diff
@@ -1,6 +1,6 @@
-pragma solidity ^0.5.10;
+pragma solidity 0.5.15;
 
-import {SafeMath} from "@summa-tx/bitcoin-spv-sol/contracts/SafeMath.sol";
+import {SafeMath} from "openzeppelin-solidity/contracts/math/SafeMath.sol";
```
NicholasDotSol, line 3: 💯

solidity/contracts/system/TBTCDepositToken.sol
```diff
@@ -45,27 +45,15 @@ contract TBTCDepositToken is ERC721Metadata, DepositFactoryAuthority {
...
+    ///                   interface below to receive approval notifications.
+    /// @param _spender   `ITokenRecipient`-conforming contract authorized to
+    ///        operate on the approved token.
     /// @param _tdtId     The TDT they can spend.
     /// @param _extraData Extra information to send to the approved contract.
-    function approveAndCall(address _spender, uint256 _tdtId, bytes memory _extraData) public returns (bool) {
-        tokenRecipient spender = tokenRecipient(_spender);
-        approve(_spender, _tdtId);
-        spender.receiveApproval(msg.sender, _tdtId, address(this), _extraData);
+    function approveAndCall(ITokenRecipient _spender, uint256 _tdtId, bytes memory _extraData) public returns (bool) {
```
NicholasDotSol, line 54: 👍

PR #566, opened 2020-04-06
Pass stake lock duration to BondedECDSAKeepFactory

nkuba:
BondedECDSAKeepFactory factory requires a stake lock duration for opening a new keep. The value is used to lock signer KEEP token stakes for the period of the keep activity. It is a protection against staked token being undelegated before the keep is closed.

In this PR we update deposit funding to pass a deposit term as the value for the stake lock duration.

This PR is a draft until https://github.com/keep-network/keep-ecdsa/issues/360 is closed. We will need to update a dependency lock to the new version of keep-ecdsa.

Refs: https://github.com/keep-network/keep-ecdsa/issues/360

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -393,13 +393,13 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     /// @notice Request a new keep opening.
     /// @param _m Minimum number of honest keep members required to sign.
     /// @param _n Number of members in the keep.
-    /// @param _stakeLockDuration Duration of stake lock in seconds.
+    /// @param _maxSecuredLifetime Duration of stake lock in seconds.
     /// @return Address of a new keep.
     function requestNewKeep(
         uint256 _m,
         uint256 _n,
         uint256 _bond,
-        uint256 _stakeLockDuration
+        uint256 _maxSecuredLifetime
     )
         external
         payable
@@ -407,6 +407,6 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     {
         require(tbtcDepositToken.exists(uint256(msg.sender)), "Caller must be a Deposit contract");
         IBondedECDSAKeepFactory _keepFactory = IBondedECDSAKeepFactory(keepVendor.selectFactory());
-        return _keepFactory.openKeep.value(msg.value)(_n, _m, msg.sender, _bond, _stakeLockDuration);
+        return _keepFactory.openKeep.value(msg.value)(_n, _m, msg.sender, _bond, _maxSecuredLifetime);
     }
 }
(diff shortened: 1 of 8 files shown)


solidity/test/TBTCSystemTest.js
```diff
@@ -33,14 +33,15 @@ describe("TBTCSystem", async function() {
     let openKeepFee
     const tdtOwner = accounts[1]
     const keepOwner = accounts[2]
+    const tokenStakeDuration = 123
```
NicholasDotSol, line 36: can we rename this to make it clear this is a lock?
maybe `stakeLockDuration`, like before

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -393,14 +393,20 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
     /// @param _m Minimum number of honest keep members required to sign.
     /// @param _n Number of members in the keep.
+    /// @param _stakeLockDuration Duration of stake lock in seconds.
     /// @return Address of a new keep.
-    function requestNewKeep(uint256 _m, uint256 _n, uint256 _bond)
+    function requestNewKeep(
+        uint256 _m,
+        uint256 _n,
+        uint256 _bond,
+        uint256 _stakeLockDuration
```
Shadowfiend, line 402: I feel like we're actually making this interface too specific without actually making it specific. This isn't a stake lock duration, that's just what the Keep uses it for. It's really the max secured lifetime of the Keep, and the Keep uses this to lock the stake. Yah?

nkuba: Yes, you're 💯 correct. How about `maxSecuredLifetime`?

Shadowfiend: I'm down.

solidity/test/TBTCSystemTest.js
nkuba: Renamed it to `maxSecuredLifetime` as per https://github.com/keep-network/tbtc/pull/566#discussion_r406559607

solidity/contracts/system/TBTCSystem.sol
nkuba: I updated the name in `TBTCSystem` contract context to `maxSecuredLifetime`.

But in `ECDSAKeepFactoryStub` I left it the same as is defined on the keep-ecdsa side in `IBondedECDSAKeepFactory` interface - `stakeLockDuration`. If you feel it has to be updated in this interface as well, please let me know.

Shadowfiend: I think that makes sense on the Keep boundary, since it's very deliberately the stake being locked, not the bond.

Shadowfiend: Eh now I'm not so sure, but I don't want to bikeshed it anymore.

PR #569, opened 2020-04-08
Use a pull model to withdraw deposit ETH balances

NicholasDotSol:
Push models have security and stability concerns, a pull model allows for better security, as well as more flexibility and freedom to third-party contract fallbacks given `.call.value()`'s gas-forwarding.

- Accumulated Eth balance can only be withdrawn after the contract reaches an end-state

### DIFF
--- a/solidity/.soliumrc.json
+++ b/solidity/.soliumrc.json
@@ -5,6 +5,7 @@
     ],
     "rules": {
         "security/no-block-members": "off",
-        "security/no-inline-assembly": "off"
+        "security/no-inline-assembly": "off",
+        "security/no-call-value": "off"
     }
 }
--- a/solidity/contracts/deposit/DepositFunding.sol
+++ b/solidity/contracts/deposit/DepositFunding.sol
@@ -82,7 +82,7 @@ library DepositFunding {
     /// @dev        This is only called as part of funding fraud flow.
     function distributeSignerBondsToFunder(DepositUtils.Deposit storage _d) internal {
         uint256 _seized = _d.seizeSignerBonds();
-        _d.depositOwner().transfer(_seized);  // Transfer whole amount
+        _d.depositOwner().call.value(_seized)("");  // Transfer whole amount
     }
 
     /// @notice     Anyone may notify the contract that signing group setup has timed out.
@@ -97,12 +97,12 @@ library DepositFunding {
         // refund the deposit owner the cost to create a new Deposit at the time the Deposit was opened.
         uint256 _seized = _d.seizeSignerBonds();
 
-        /* solium-disable-next-line security/no-send */
-        _d.depositOwner().send(_d.keepSetupFee);
-        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
-
         _d.setFailedSetup();
         _d.logSetupFailed();
+
+        /* solium-disable-next-line security/no-send */
+        _d.depositOwner().call.value(_d.keepSetupFee)("");
+        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
 
         fundingTeardown(_d);
     }
--- a/solidity/contracts/deposit/DepositFunding.sol
+++ b/solidity/contracts/deposit/DepositFunding.sol
@@ -99,12 +99,11 @@ library DepositFunding {
 
         _d.setFailedSetup();
         _d.logSetupFailed();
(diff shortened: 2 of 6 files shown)


solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -396,6 +399,26 @@ library DepositUtils {
...
+    /// @notice     Adds a given amount to an addresses withdrawal allowance
+    /// @dev        Withdrawals can only happen post contract-end-state.
+    function enableWithdrawal(DepositUtils.Deposit storage _d, address _withdrawer, uint256 _amount) internal {
+        withdrawalAllowances[_withdrawer].add(_amount);
+    }
+
+    /// @notice     Withdraw caller's allowance.
+    /// @dev        Withdrawals can only happen post contract-end-state.
+    function withdrawFunds(DepositUtils.Deposit storage _d) internal {
+        require(_d.inEndState(), "Contact not yet terminated");
```
Shadowfiend, line 412: Contract*

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -190,21 +196,18 @@ library DepositLiquidation {
...
         // we simply ignore such a tiny amount and leave some wei dust in escrow
         uint256 contractEthBalance = address(this).balance;
-        address payable initiator = _d.liquidationInitiator;
 
-        if (initiator == address(0)){
-            initiator = address(0xdead);
-        }
-        if (contractEthBalance > 1) {
+        if (contractEthBalance > _valueToDistribute) {
+            uint256 remainingUnallocated = contractEthBalance.sub(_valueToDistribute);
```
Shadowfiend, line 201: `_valueToDistribute` really shouldn't start with an underscore, as a side note.

solidity/.soliumrc.json
```diff
@@ -5,6 +5,7 @@
     ],
     "rules": {
         "security/no-block-members": "off",
-        "security/no-inline-assembly": "off"
+        "security/no-inline-assembly": "off",
+        "security/no-call-value": "off"
```
Shadowfiend, line 9: We don't need this.

solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -97,14 +90,13 @@ library DepositFunding {
...
-        /* solium-disable-next-line security/no-send */
-        _d.depositOwner().send(_d.keepSetupFee);
-        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
-
         _d.setFailedSetup();
         _d.logSetupFailed();
-
         fundingTeardown(_d);
+
+        /* solium-disable-next-line security/no-send */
```
Shadowfiend, line 97: We can lose all of these.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -396,6 +399,26 @@ library DepositUtils {
...
+        require(withdrawalAllowances[msg.sender] > 0, "Nothing to withdraw"");
+
+        // zero-out to prevent reentrancy
+        uint256 available = withdrawalAllowances[msg.sender];
+        withdrawalAllowances[msg.sender] = 0;
+    
+        msg.sender.call.value(available)("");
```
Shadowfiend, line 419: We *will* need a linter annotation here I guess.

solidity/.soliumrc.json
NicholasDotSol: you'd rather have the solium-disable comment?

Shadowfiend: What I really want is a way to annotate a function as “I know this is going to do a `.call.value` and I'm completely okay with that”, and only in the call path of that one function is anyone allowed to do `.call.value`. I would like this compiler-enforced.

In the absence of such a guard, this at least draws extra attention to where we're doing `call.value`. But I wish we could annotate the whole function.

Basically I'm trying to use a linter to do a compiler's job and it's terrible. But I think it's marginally less terrible than not doing so.

NicholasDotSol: just saw https://github.com/keep-network/tbtc/pull/569#discussion_r405896462

NicholasDotSol: Makes sense to use linter annotation over the rule if it's just one instance

NicholasDotSol:
close in favor of #570

PR #570, opened 2020-04-09
Use a pull model to withdraw deposit ETH balances

### LINKED ISSUE #551: Secure payments from bad recipient behavior
There are several cases where a third party is the recipient of an ETH payment from tBTC:
- Liquidation sends the appropriate amount of seized bonds to the liquidation auction winner, and potentially returns the leftover bond to the liquidation initiator, and possibly to the signers.
- Fraud liquidation during redemption immediately sends the redeemer the seized bonds from the signing group.
- Signing group fraud during funding sends the funder the seized bonds from the signing group.
- Signing group setup failure refunds the funder for the deposit opening fee and potentially returns leftover bond to the signers.
- Successful redemption returns leftover bond to the signers.

In all these cases, it's important that a misconfigured or malicious contract on the receiving side cannot prevent the system from completing the action it is executing.

There are two approaches here: additional bookkeeping and allowing relevant parties to *withdraw* their due amounts, or securing the outbound transfers so that they cannot break the remainder of the transaction.

NicholasDotSol:
Push models introduce security and stability concerns, a pull model allows for better security, as well as more flexibility and freedom to third-party contract fallbacks given `.call.value()`'s gas-forwarding.

- Accumulated Eth balance can only be withdrawn after the contract reaches an end-state.

Closes #551.

### DIFF
--- a/solidity/contracts/deposit/DepositFunding.sol
+++ b/solidity/contracts/deposit/DepositFunding.sol
@@ -163,7 +163,7 @@ library DepositFunding {
         require(_isFraud, "Signature is not fraudulent");
         _d.logFraudDuringSetup();
 
-        // allow deposit owner to withdraw sized bonds after contract termination;
+        // Allow deposit owner to withdraw seized bonds after contract termination.
         uint256 _seized = _d.seizeSignerBonds();
         _d.enableWithdrawal(_d.depositOwner(), _seized);
 
--- a/solidity/contracts/deposit/DepositFunding.sol
+++ b/solidity/contracts/deposit/DepositFunding.sol
@@ -78,13 +78,6 @@ library DepositFunding {
         return true;
     }
 
-    /// @notice     Seizes signer bonds and distributes them to the funder.
-    /// @dev        This is only called as part of funding fraud flow.
-    function distributeSignerBondsToFunder(DepositUtils.Deposit storage _d) internal {
-        uint256 _seized = _d.seizeSignerBonds();
-        _d.depositOwner().transfer(_seized);  // Transfer whole amount
-    }
-
     /// @notice     Anyone may notify the contract that signing group setup has timed out.
     /// @param  _d  Deposit storage pointer.
     function notifySignerSetupFailure(DepositUtils.Deposit storage _d) public {
@@ -98,7 +91,7 @@ library DepositFunding {
         uint256 _seized = _d.seizeSignerBonds();
 
         /* solium-disable-next-line security/no-send */
-        _d.depositOwner().send(_d.keepSetupFee);
+        _d.enableWithdrawal(_d.depositOwner(), _d.keepSetupFee);
         _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
 
         _d.setFailedSetup();
@@ -169,7 +162,11 @@ library DepositFunding {
         bool _isFraud = _d.submitSignatureFraud(_v, _r, _s, _signedDigest, _preimage);
         require(_isFraud, "Signature is not fraudulent");
         _d.logFraudDuringSetup();
-        distributeSignerBondsToFunder(_d);
+
+        // allow deposit owner to withdraw sized bonds after contract termination;
+        uint256 _seized = _d.seizeSignerBonds();
+        _d.enableWithdrawal(_d.depositOwner(), _seized);
+
         fundingFraudTeardown(_d);
         _d.setFailedSetup();
         _d.logSetupFailed();
--- a/solidity/contracts/deposit/DepositLiquidation.sol
+++ b/solidity/contracts/deposit/DepositLiquidation.sol
@@ -179,8 +179,8 @@ library DepositLiquidation {
         }
 
         // Distribute funds to auction buyer
-        uint256 _valueToDistribute = _d.auctionValue();
-        _d.enableWithdrawal(msg.sender, _valueToDistribute);
+        uint256 valueToDistribute = _d.auctionValue();
+        _d.enableWithdrawal(msg.sender, valueToDistribute);
 
         // Send any TBTC left to the Fee Rebate Token holder
         _d.distributeFeeRebate();
@@ -195,8 +195,8 @@ library DepositLiquidation {
         if (initiator == address(0)){
             initiator = address(0xdead);
         }
-        if (contractEthBalance > _valueToDistribute) {
-            uint256 remainingUnallocated = contractEthBalance.sub(_valueToDistribute);
+        if (contractEthBalance > valueToDistribute) {
+            uint256 remainingUnallocated = contractEthBalance.sub(valueToDistribute);
             if (_wasFraud) {
                 _d.enableWithdrawal(initiator, remainingUnallocated);
             } else {
--- a/solidity/contracts/deposit/DepositLiquidation.sol
+++ b/solidity/contracts/deposit/DepositLiquidation.sol
@@ -195,7 +195,7 @@ library DepositLiquidation {
         if (initiator == address(0)){
             initiator = address(0xdead);
         }
-        if (contractEthBalance > valueToDistribute) {
+        if (contractEthBalance > valueToDistribute + 1) {
             uint256 remainingUnallocated = contractEthBalance.sub(valueToDistribute);
             if (_wasFraud) {
                 _d.enableWithdrawal(initiator, remainingUnallocated);
--- a/solidity/contracts/deposit/DepositLiquidation.sol
+++ b/solidity/contracts/deposit/DepositLiquidation.sol
@@ -74,7 +74,7 @@ library DepositLiquidation {
         if (_d.auctionTBTCAmount() == 0) {
             // we came from the redemption flow
             _d.setLiquidated();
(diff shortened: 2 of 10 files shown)


solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -169,7 +162,11 @@ library DepositFunding {
         bool _isFraud = _d.submitSignatureFraud(_v, _r, _s, _signedDigest, _preimage);
         require(_isFraud, "Signature is not fraudulent");
         _d.logFraudDuringSetup();
-        distributeSignerBondsToFunder(_d);
+
+        // allow deposit owner to withdraw sized bonds after contract termination;
```
Shadowfiend, line 166: Seized? Guessing this is meant to be capitalized + end with a period as well <_<

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -195,16 +195,15 @@ library DepositLiquidation {
         if (initiator == address(0)){
             initiator = address(0xdead);
         }
-        if (contractEthBalance > 1) {
+        if (contractEthBalance > valueToDistribute) {
```
Shadowfiend, line 198: + 1?

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -66,6 +66,9 @@ library DepositUtils {
         uint256 fundedAt; // timestamp when funding proof was received
         bytes utxoOutpoint;  // the 36-byte outpoint of the custodied UTXO
 
+        /// @notice Map of ETH balances an address can withdraw after contract reaches ends-state.
```
Shadowfiend, line 69: Why a `@notice` here when we don't do it anywhere else?

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -396,6 +399,34 @@ library DepositUtils {
         return _postCallBalance.sub(_preCallBalance);
     }
 
+    /// @notice     Adds a given amount to an addresses withdrawal allowance.
```
Shadowfiend, line 402: I would rephrase “Adds a given amount to the withdrawal allowance for the address.” to avoid the awkward address possessive here…

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -396,6 +399,34 @@ library DepositUtils {
...
+
+        // zero-out to prevent reentrancy
+        _d.withdrawalAllowances[msg.sender] = 0;
+
+        /* solium-disable-next-line security/no-call-value */
+        msg.sender.call.value(available)("");
+    }
+
+    /// @notice     Get the caller's withdraw allowance.
+    /// @return     The caller's withdrawable allowance in wei.
```
Shadowfiend, line 425: We're back-and-forthing between “withdraw allowance” and “withdrawable allowance”. I'm fine with “withdraw allowance”, but let's use it everywhere.

solidity/test/DepositFundingTest.js
```diff
@@ -218,16 +218,15 @@ describe("DepositFunding", async function() {
...
       const signingGroupRequestedAt = await testDeposit.getSigningGroupRequestedAt.call()
-      const finalFunderBalance = await web3.eth.getBalance(owner)
 
-      expect(
-        new BN(finalFunderBalance).sub(new BN(initialFunderBalance)),
-      ).to.eq.BN(openKeepFee)
+      const withdrawable = await testDeposit.getWithdrawAllowance.call({
+        from: owner,
+      })
+      expect(withdrawable).to.eq.BN(new BN(openKeepFee))
```
Shadowfiend, line 229: This is kind of nice. Can we also assert that the deposit balance is equal to this amount?

solidity/test/DepositLiquidationTest.js
```diff
@@ -163,23 +163,25 @@ describe("DepositLiquidation", async function() {
...
+        from: buyer,
+      })
 
-      expect(
-        new BN(finalBalance),
-        "buyer balance should increase",
-      ).to.be.gte.BN(initialBalance)
+      expect(withdrawable, "buyer should have a withdrawable balance").to.eq.BN(
+        auctionValue,
+      )
```
Shadowfiend, line 184: Ditto here on asserting the deposit has this amount in escrow.

solidity/test/DepositLiquidationTest.js
```diff
@@ -203,24 +202,22 @@ describe("DepositLiquidation", async function() {
...
 
       const totalReward = (value * (100 - basePercentage)) / 100
       const split = totalReward / 2
 
-      expect(new BN(split)).to.eq.BN(initiatorBalanceDiff)
+      const withdrawable = await testDeposit.getWithdrawAllowance.call({
+        from: liquidationInitiator,
+      })
+
+      expect(new BN(split)).to.eq.BN(withdrawable)
```
Shadowfiend, line 220: \+ again for the deposit balance check.

solidity/test/DepositLiquidationTest.js
```diff
@@ -246,24 +240,22 @@ describe("DepositLiquidation", async function() {
...
-      const initiatorBalanceDiff = new BN(finalInitiatorBalance).sub(
-        new BN(initialInitiatorBalance),
-      )
       const signerBalanceDiff = new BN(finalSignerBalance).sub(
         new BN(initialSignerBalance),
       )
 
+      const withdrawable = await testDeposit.getWithdrawAllowance.call({
+        from: liquidationInitiator,
+      })
```
Shadowfiend, line 253: Uno mas on the deposit balance check.

solidity/test/DepositUtilsTest.js
```diff
@@ -641,6 +642,71 @@ describe("DepositUtils", async function() {
...
+
+    it("correctly adds value to withdrawalAllowances", async () => {
+      const value = new BN(10000)
+      const beneficiary = accounts[1]
+
+      await testDeposit.enableWithdrawal(accounts[1], value)
+
+      const withdrawable = await testDeposit.getWithdrawAllowance.call({
+        from: beneficiary,
+      })
```
Shadowfiend, line 662: We should probably assert this is 0 before the `enableWithdrawal` call.

solidity/test/DepositUtilsTest.js
```diff
@@ -641,6 +642,71 @@ describe("DepositUtils", async function() {
...
+  describe("withdrawFunds()", async () => {
+    beforeEach(async () => {
+      await createSnapshot()
+    })
+
+    afterEach(async () => {
+      await restoreSnapshot()
+    })
+
+    it("Correctly withdraws withdrawable balance at end-state", async () => {
```
Shadowfiend, line 677: We should make this test check every state that is an end-state, and make the next test check every state that isn't.

Also, as far as I know we don't capitalize our “it” strings. You should consider these the end of a sentence that follows from previous `describe`s (e.g. `DepositUtils withdrawFunds() correctly withdraws withdrawable balance at end-state`).

solidity/test/DepositUtilsTest.js
```diff
@@ -641,6 +642,71 @@ describe("DepositUtils", async function() {
...
+      await testDeposit.enableWithdrawal(beneficiary, value)
+      await testDeposit.withdrawFunds({from: beneficiary})
+
+      const withdrawable = await testDeposit.getWithdrawAllowance.call({
+        from: beneficiary,
+      })
+      const finalBalance = await web3.eth.getBalance(beneficiary)
+
+      expect(withdrawable).to.eq.BN(0)
+      expect(finalBalance).to.gt.BN(initialBalance)
```
Shadowfiend, line 693: Just because the final balance is > initial balance, doesn't mean we succeeded, right? We need to make sure we withdraw initial balance + the value enabled for withdrawal.

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: we do it in `approvedDigests` mapping right below. We can in-line both instead

Shadowfiend: I'm down either way---if this results in proper API doc JSON that's fine, though we should use `@dev` rather than `@notice` given I don't think we expect users to directly interact with the mapping.

solidity/contracts/deposit/DepositLiquidation.sol
NicholasDotSol: derp, brain glitched on this, thought about it too... nice catch

solidity/test/DepositFundingTest.js
NicholasDotSol: Yeah that would be good

solidity/test/DepositLiquidationTest.js
NicholasDotSol: In this case, we need `contractBalance = purchesedBond + remainder / 2` since we're not in fraud.

solidity/test/DepositUtilsTest.js
NicholasDotSol: That's right, I got lazy accounting for the gas cost since we're calling from `beneficiary`. Meant to fix this

Issue #553, opened 2020-04-04
Investigate cleaner way of identifying signer fraud liquidation redemption origin


Shadowfiend:
ConsenSys Diligence noted that this:

https://github.com/keep-network/tbtc/blob/5f7f05729c7670174fdd0e5d819b1926e7d06045/solidity/contracts/deposit/DepositLiquidation.sol#L74-L80

Is not an ideal way of deriving the prior state of the deposit. Let's investigate if there's a cleaner approach rooted in the deposit's actual state.

Shadowfiend:
Can we just check for `inRedemption` here?

Issue #319, opened 2019-10-09
Test liquidation flows


liamzebedee:
Liquidation flows located in `startSignerAbortLiquidation` (timeouts) and `startSignerFraudLiquidation` (stolen bonds/tbtc) are not thoroughly tested yet. 

Right now we have two types of liquidation - (1) auction and (2) on-chain using Uniswap. The liquidation flows should test that ETH/TBTC is properly refunded to the relevant actors in the system. 

We must also consider whether the liquidation comes from non-redemption (usual operation) or redemption (redeeming tbtc for btc). It may be wise to refactor some code in DepositLiquidation to make this simpler.
 
I previously implemented [some code](https://github.com/keep-network/tbtc/blob/test-liquidation-flows/implementation/test/DepositLiquidationTest.js#L686) for testing onchain liquidation, which might be helpful.

Shadowfiend:
These are generally taken care of at this point, both with state tests and with targeted tests @NicholasDotSol wrote.

PR #573, opened 2020-04-10
Liquidation-entry refactor

NicholasDotSol:
- `startSignerFraudLiquidatoin` and `StartSignerAbortLiquidation` have been consolidated to a single `startLiquidation` method. a boolean is passed to  differentiate between abort and fraud.

- if coming from redemption, always `setLiquidated` and make the signer bonds redeemer-withdrawable. Check the deposit state directly rather than relying on `redeemerAddress` being set to determine if we're coming from the liquidation flow.

closes: #553 

draft untill: 
- ~~ensure redeemable state `startLiquidation` path doesn't introduce any holes~~ -> moved to #319 .

### DIFF
--- a/solidity/contracts/deposit/DepositLiquidation.sol
+++ b/solidity/contracts/deposit/DepositLiquidation.sol
@@ -61,42 +61,37 @@ library DepositLiquidation {
         return (_bondValue.mul(100).div(_lotValue));
     }
 
-    /// @notice         Starts signer liquidation due to fraud.
-    /// @dev            Liquidation is done by auction.
-    /// @param  _d      Deposit storage pointer.
-    function startSignerFraudLiquidation(DepositUtils.Deposit storage _d) internal {
-        _d.logStartedLiquidation(true);
+    /// @notice           Starts signer liquidation due to abort, undercollateralization, or fraud
+    /// @dev              Liquidation is done by auction.
+    /// @param _wasFraud  Boolean checking liquidation cause. True if fraud, false otherwise.
+    /// @param _d         Deposit storage pointer.
+    function startLiquidation(DepositUtils.Deposit storage _d, bool _wasFraud) internal {
+        _d.logStartedLiquidation(_wasFraud);
 
-        // Reclaim used state for gas savings
-        _d.redemptionTeardown();
         uint256 _seized = _d.seizeSignerBonds();
 
-        if (_d.auctionTBTCAmount() == 0) {
-            // we came from the redemption flow
+        // if we come from the redemption flow we shouldn't go to auction.
+        // Instead give the signer bonds to redeemer
+        if (_d.inRedemption()) {
             _d.setLiquidated();
             _d.enableWithdrawal(_d.redeemerAddress, _seized);
             _d.logLiquidated();
+            // Reclaim used state for gas savings
+            _d.redemptionTeardown();
             return;
         }
 
-        _d.liquidationInitiator = msg.sender;
-        _d.liquidationInitiated = block.timestamp;  // Store the timestamp for auction
-
-        _d.setFraudLiquidationInProgress();
-    }
-
-    /// @notice         Starts signer liquidation due to abort or undercollateralization.
-    /// @dev            Liquidation is done by auction.
-    /// @param  _d      Deposit storage pointer.
-    function startSignerAbortLiquidation(DepositUtils.Deposit storage _d) internal {
-        _d.logStartedLiquidation(false);
         // Reclaim used state for gas savings
         _d.redemptionTeardown();
-        _d.seizeSignerBonds();
-
-        _d.liquidationInitiated = block.timestamp;  // Store the timestamp for auction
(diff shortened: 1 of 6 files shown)


solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -61,42 +61,37 @@ library DepositLiquidation {
         return (_bondValue.mul(100).div(_lotValue));
     }
 
-    /// @notice         Starts signer liquidation due to fraud.
-    /// @dev            Liquidation is done by auction.
-    /// @param  _d      Deposit storage pointer.
-    function startSignerFraudLiquidation(DepositUtils.Deposit storage _d) internal {
-        _d.logStartedLiquidation(true);
+    /// @notice           Starts signer liquidation due to abort, undercollateralization, or fraud
```
Shadowfiend, line 64: How about something like “Starts signer liquidation by seizing signer bonds. If the deposit is currently being redeemed, the redeemer receives the full bond value; otherwise, a falling price auction begins to buy 1 TBTC in exchange for a portion of the seized bonds; see purchaseSignerBondsAtAuction.”

Since this is internal, I think the whole thing can be `@dev`.

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -61,42 +61,37 @@ library DepositLiquidation {
...
+    /// @dev              Liquidation is done by auction.
+    /// @param _wasFraud  Boolean checking liquidation cause. True if fraud, false otherwise.
```
Shadowfiend, line 66: I think this could be more clearly stated as “True if liquidation is being started due to fraud, false if for any other reason.”

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -61,42 +61,37 @@ library DepositLiquidation {
...
-        if (_d.auctionTBTCAmount() == 0) {
-            // we came from the redemption flow
+        // if we come from the redemption flow we shouldn't go to auction.
+        // Instead give the signer bonds to redeemer
+        if (_d.inRedemption()) {
             _d.setLiquidated();
             _d.enableWithdrawal(_d.redeemerAddress, _seized);
             _d.logLiquidated();
+            // Reclaim used state for gas savings
+            _d.redemptionTeardown();
```
Shadowfiend, line 80: We can still do this once alongside the bond seizure, right? `redemptionTeardown` doesn't change the deposit state, so the `_d.inRedemption()` check should still work.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -367,11 +356,11 @@ library DepositUtils {
     /// @notice     Deletes state after termination of redemption process.
     /// @dev        We keep around the redeemer address so we can pay them out.
     function redemptionTeardown(Deposit storage _d) public {
-        // don't 0 redeemerAddress because we use it to calculate auctionTBTCAmount
         _d.redeemerOutputScript = "";
         _d.initialRedemptionFee = 0;
         _d.withdrawalRequestTime = 0;
         _d.lastRequestedDigest = bytes32(0);
+        _d.redeemerAddress = address(0);
```
Shadowfiend, line 363: This affects my sinister plan above. I kind of like keeping the redeemer address as part of the final state of a deposit though… It means a closed deposit knows its original lot size, its final owner, and its redeemer. Feels clean somehow? Though I realize this may make the final tx cheaper 🤔

Shadowfiend: It also feels like cleaning this up could inadvertently break some stuff that isn't worth spending time on fixing…

solidity/contracts/deposit/DepositLiquidation.sol
NicholasDotSol: `redemptionTeardown()` zeroes out `redeemerAddress`
This is new, I just added it to claim more state back.

NicholasDotSol: I just saw comment below...

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: I'd rather save the gas than have a redeemer value in storage. 

> It also feels like cleaning this up could inadvertently break some stuff that isn't worth spending time on fixing…

when we reclaim this state we are always wrapping up. as long as the state is reclaimed after its last use there should be no problem

Shadowfiend: We're also assuming the gas saved will be saved, since there's a limit to how much gets refunded via state release. The transactions that do this aren't super expensive.

NicholasDotSol: that's a fair point, but we're definitely not reaching the cap for functions like `provideRedemptionProof`. We'll surely see the refund there

PR #576, opened 2020-04-13
Inter-medi[anizer]-ary: Add intermediary authorized medianizer access contract for Ropsten

Shadowfiend:
So that new system deployments don't require new contract authorizations
on the Maker side, an intermediate medianizer has already been deployed
to Ropsten for Maker to authorize. It is designed to have an updatable
reference to the Maker medianizer, and pass through reads to that
instance.

The already-deployed address is added to the externals file, while the
code is checked in to the price-feed directory, but not migrated as part
of the system migrations.

-------

Still needed here:
- [x] ~~Merging #545~~ Rebase onto #545.
- ~~Rebasing this against the resulting `master` and aligning externals hookup.~~


solidity/contracts/price-feed/IntermediaryMedianizer.sol
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(IMedianizer realMedianizer) public {
+        _realMedianizer = realMedianizer;
+    }
+
+    function getMedianizer() external view returns (IMedianizer) {
+        return _realMedianizer;
+    }
+
+    function read() external view returns (uint256) {
```
NicholasDotSol, line 25: we use the `peek` functions too, in order to check if the medialized is active

Shadowfiend: Well crap. That's going to require a new deployment. Wish we could just try/catch, but I guess that's for the land of Solidity 0.6.

Shadowfiend: Amended this as part of the rebase.

solidity/contracts/external/IMedianizer.sol
```diff
@@ -5,12 +5,12 @@ pragma solidity 0.5.17;
 interface IMedianizer {
     /// @notice Get the current price.
     /// @dev May revert if caller not whitelisted.
-    /// @return Price (USD) with 18 decimal places.
+    /// @return Designated price with 18 decimal places.
     function read() external view returns (uint256);
 
     /// @notice Get the current price and check if the price feed is active
     /// @dev May revert if caller not whitelisted.
     /// @return Price (USD) with 18 decimal places.
```
NicholasDotSol, line 13: `Price (USD) -> Designated price` like above?
` with 18 decimal places` is tris true for all maker medianizers?

Shadowfiend: It appears to be true for all of them, yes.

Issue #554, opened 2020-04-04
Align comments, TODOs, and spec with final system details


Shadowfiend:
There are currently a variety of TODOs, comments, and spec details that have drifted from the contracts as we've made final tweaks, updates, and addressed audit notes. Make sure these propagate through non-code artifacts.

Also see #293 .

mhluongo:
Closed by #645

PR #581, opened 2020-04-18
Parameter Finalization: Clear TODOs, set final parameter values

### LINKED ISSUE #555: Finalize system constants in code
The code currently features a few system constants that are not at their final value. Set these to their mainnet value and, if needed, adjust them in stub contracts for testing. See also #396 and #116.

Shadowfiend:
Other things:
- `TBTCDevelopmentConstants`, with 1 confirmation and the old min redemption fee. Deployed only for keep_dev and local, otherwise the standard constants are used. (That means 6 confirmations for Ropsten once this lands, heads up!)
- Put files where they belong (mostly constants now live in `system/`, and test files got moved to align with their non-test counterparts).

------

Missing:
- [x] A development shim with 1 confirmation instead of 6? At least want this for the internal testnet.

Closes #555. See #554.


Shadowfiend:
@liamzebedee @NicholasDotSol another small one here!

solidity/migrations/2_deploy_contracts.js
```diff
@@ -57,6 +58,11 @@ const all = [
 
 module.exports = (deployer, network, accounts) => {
   deployer.then(async () => {
+    if (network == "keep_dev" || network == "local") {
```
liamzebedee, line 61: Are you missing the "development" network? We've got it [here](https://github.com/keep-network/tbtc/blob/master/solidity/truffle-config.js#L52) in truffle-config.

solidity/migrations/2_deploy_contracts.js
```diff
@@ -57,6 +58,11 @@ const all = [
...
+      // For keep_dev and local, replace constants with testnet constants.
+      all[all.indexOf(TBTCConstants)] = TBTCDevelopmentConstants
```
liamzebedee, line 63: A little confused by what's happening here. 🤔

 - `deployer.link(library, destinations)` links a library to the destination contracts.
 - we deploy the `TBTCConstants` contract [below](https://github.com/keep-network/tbtc/blob/d4af3b25cc146971f1982a75cbd4388d76201e20/solidity/migrations/2_deploy_contracts.js#L80), and link it into the dependent contracts (eg. TBTCSystem). The TBTCConstants that is linked is the original, TBTCConstants?
 - You've (re)set TBTCConstants in the `all` array, which is the array of contracts with unresolved libraries/symbols.

solidity/migrations/2_deploy_contracts.js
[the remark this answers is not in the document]
Shadowfiend: Hmmm... I think this is a difference between here and one of our other projects. I’ll either rename that one local for consistency or change this to refer to development.

solidity/migrations/2_deploy_contracts.js
Shadowfiend: Hah. I missed the deployment line in the update 🤦‍♂️  will fix.

solidity/migrations/2_deploy_contracts.js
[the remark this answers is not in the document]
Shadowfiend: We do indeed use `local` in keep-ecdsa and keep-core both, but rather than disentangle this against truffle's preference for `development` as the default network I'm going to just tweak this conditional. More later.

solidity/migrations/2_deploy_contracts.js
```diff
@@ -57,6 +58,15 @@ const all = [
 
 module.exports = (deployer, network, accounts) => {
   deployer.then(async () => {
+    let constantsContract = TBTCConstants
+    if (network == "keep_dev" || network == "development") {
+      // For keep_dev and development, replace constants with testnet constants.
+      // Masquerade as TBTCConstants like a sinister fellow.
```
liamzebedee, line 64: 👻

liamzebedee: Still don't quite understand why we're adding it to the `all` array @Shadowfiend. AFAIK the *TBTCConstants library doesn't need anything linked into it.

Shadowfiend: I think I added this back when I misunderstood the reason for the issue that led me to hardcore the contract name here. I’ll see about pulling it back out.

PR #583, opened 2020-04-18
Escape Hatch: Add a way for owners of a funding-failed deposit to ask nicely for their UTXO(s) back

### LINKED ISSUE #550: "Whoops I misfunded, please return" function?
Overfunding is something we can't really prevent, nor can we incontrovertibly prevent a funder from constructing a funding transaction whose proof can't be processed. In the former case, the deposit represents an incorrect lot size, losing value for the funder, but the funder can immediately redeem the deposit. In the latter case, the funder's UTXO is forfeit in the system design.

The proposal here is to introduce a “please return my UTXO” call on the Deposit contract that is unenforced. It would take a receiver script and immediately close the deposit and return signer bonds, but honest signers would ideally sign and broadcast a signed transaction to return the bad funding UTXO to the given script. The system would provide *no* guarantee of such, however, and for all intents and purposes this would be treated as a funding failure.

Shadowfiend:
To allow for mistakes in funding that can't be managed on-chain, this PR introduces a new deposit function, `requestFunderAbort`, that can only be called by the TDT holder after deposit setup has failed. The function merely emits a `FunderAbortRequested` event that carries a funder-specified output script. Signers *should*, though they are not obligated by protocol or cryptoeconomically incentivized, coordinate to return any UTXO to the signer-controlled wallet back to the requested output script in the abort request.

------

Remaining:
- [x] Tests.

Closes #550.


solidity/contracts/deposit/Deposit.sol
```diff
@@ -282,6 +282,25 @@ contract Deposit is DepositFactoryAuthority {
         return true;
     }
 
+    /// @notice Requests a funder abort for a failed-funding deposit; that is,
+    ///         requests the return of a sent UTXO to _abortOutputScript. It
+    ///         imposes no requirements on the signing group. Signers may or may
+    ///         not send their UTXO to the requested output script, at their
+    ///         discretion and with no penalty for failing to do so. This can be
```
mhluongo, line 289: I think it's worth explaining that while "there's no protocol incentive" a signer still "should" return funds, or similar? Just being clear that while the protocol can't enforce it, that doesn't mean depositors need to expect signers will steal their funds. Most signers at launch will be known entities after all

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -44,6 +44,9 @@ library DepositUtils {
         uint256 courtesyCallInitiated; // When the courtesy call is issued
         address payable liquidationInitiator;
 
+        // SET SOLELY IF FUNDER ABORT IS REQUESTED
+        bytes funderAbortOutputScript; // An output script in case of funder abort
```
mhluongo, line 48: Should we use storage here or rely on events?

Shadowfiend: Asked myself same… Feels weird to only rely on events given how we run the rest of things, but there's also no need to confirm anything about the script later, so we could do without storage…

Shadowfiend: Switching to events-only here. Also making this actually emit the event instead of just pretending 😬

solidity/contracts/deposit/Deposit.sol
Shadowfiend: Did a pass at this, have a look.

Shadowfiend:
@NicholasDotSol one more for youuuu.

solidity/test/DepositFundingTest.js
```diff
@@ -394,6 +393,58 @@ describe("DepositFunding", async function() {
...
+    before(async () => {
+      timer = await tbtcConstants.getFundingTimeout.call()
+      owner = accounts[1]
+      await tbtcDepositToken.forceMint(
+        owner,
+        web3.utils.toBN(testDeposit.address),
+      )
+    })
+
+    beforeEach(async () => {
```
NicholasDotSol, line 409: nit, but we can move all of this to the `before` block, and then: 
```
    beforeEach(async () => {
      await createSnapshot()
    })

    afterEach(async () => {
      await restoreSnapshot()
    })
```

Shadowfiend: Bah, I actually wanted to do this, but I think we're close enough here that I'm going to punt it to an imaginary future improvement.

PR #584, opened 2020-04-18
Test refinements

Shadowfiend:
Went a little ape here, but the commits say it all. My main goal here was to wrap up remaining compiler warnings and use receipt block info instead of `getBlock("latest")` wherever it made sense.


Shadowfiend:
@liamzebedee @NicholasDotSol same, ready for review when either of you gets a chance.

solidity/test/TBTCSystemTest.js
```diff
@@ -323,30 +323,31 @@ describe("TBTCSystem", async function() {
...
+        const {
+          receipt: {blockNumber: updateStartBlock},
+        } = truffleReceipt
+        const block = await web3.eth.getBlock(updateStartBlock)
+
+        expectEvent(truffleReceipt, "LotSizesUpdateStarted", {
+          _timestamp: new BN(block.timestamp),
+        })
+        expect(truffleReceipt.logs[0].args[0][0]).to.eq.BN(testSizes[0])
+        expect(truffleReceipt.logs[0].args[0][1]).to.eq.BN(testSizes[1])
```
liamzebedee, line 336: 👏

.circleci/config.yml
```diff
@@ -61,12 +61,6 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
-      - run: sudo npm install -g ganache-cli@6.4.2
```
liamzebedee, line 64: @NicholasDotSol what do we use now?

Shadowfiend: I believe something in the chain (possibly openzeppelin) just runs this in the same process.

PR #587, opened 2020-04-20
liquidation tests direction gauge

NicholasDotSol:
Test liquidation paths more organically. Also test different scenarios based on number of deposits
and price feed fluctuation. The intent is to approximate deposit behavior, simulate expected behavior, and validate outcomes


NicholasDotSol:
Should be a draft

PR #590, opened 2020-04-21
Expose collateralization functions and Keep address

NicholasDotSol:
These functions return per-deposit collateralization details. Exposing these can be useful for monitoring and managing liquidation easily.
The Keep address can be useful in order to easily obtain useful info about the keep.

### DIFF
--- a/solidity/contracts/deposit/Deposit.sol
+++ b/solidity/contracts/deposit/Deposit.sol
@@ -39,6 +39,13 @@ contract Deposit is DepositFactoryAuthority {
         require(msg.sender == self.keepAddress, "Deposit contract can only receive ETH from underlying keep");
     }
 
+    /// @notice     Get the keep contract address associated with the Deposit.
+    /// @dev        The keep contract address is saved on Deposit initialization.
+    /// @return     Address of the Keep contract.
+    function getKeepAddress() public view returns (address) {
+        return self.keepAddress;
+    }
+
     /// @notice     Get the integer representing the current state.
     /// @dev        We implement this because contracts don't handle foreign enums well.
     ///             see DepositStates for more info on states.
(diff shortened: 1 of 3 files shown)


solidity/contracts/test/deposit/TestDeposit.sol
```diff
@@ -123,10 +123,6 @@ contract TestDeposit is Deposit {
         self.keepAddress = _keepAddress;
     }
 
-    function getKeepAddress() public view returns (address) {
-        return self.keepAddress;
-    }
-
```
Shadowfiend, line 129: You have getters for undercollateralized percentages below as well.

NicholasDotSol: took care of them

solidity/contracts/deposit/Deposit.sol
```diff
@@ -384,6 +391,39 @@ contract Deposit is DepositFactoryAuthority {
     // LIQUIDATION
     //
 
+    /// @notice Get the current collateralization level for this Deposit.
+    /// @dev    This value represents the percentage of the backing BTC value the signers
+    ///         currently must hold as bond.
+    /// @return The severe collateralization level for this deposit.
```
liamzebedee, line 397: @NicholasDotSol potential copypasta re: "severe"?

liamzebedee: Also is there a reason why `getCollateralizationPercentage` is uint256? Seems like we could return uint16 here too, just curious.

NicholasDotSol: > @NicholasDotSol potential copypasta re: "severe"?

yeah good catch, severe -> current

> Also is there a reason why getCollateralizationPercentage is uint256? Seems like we could return uint16 here too, just curious.

There is no reason to move this to a smaller uint size. It makes sense for other collateralization values that take up storage since we can take advantage of bit-packing. But Since this value is calculated in memory, converting from and to uint types will just take additional gas

liamzebedee: > yeah good catch, severe -> current

also note that this is repeated below too. 

> There is no reason to move this to a smaller uint size. It makes sense for other collateralization values that take up storage since we can take advantage of bit-packing. But Since this value is calculated in memory, converting from and to uint types will just take additional gas

Alright, fair enough. Didn't know that there were gas costs for casting to lower types, thought everything was the same word-size in memory.

PR #575, opened 2020-04-10
4x400: Early pass at a relay maintainer initcontainer

Shadowfiend:
> [The **4 × 400 metres relay** or long relay is an athletics track event in which teams consist of four runners who each complete 400 metres or one lap. It is traditionally the final event of a track meet. At top class events, the first 500 metres is run in lanes. Start lines are thus staggered over a greater distance than in an individual 400 metres race; the runners then typically move to the inside of the track. The slightly longer 4 × 440 yards relay was a formerly run British and American event, until metrication was completed in the 1970s.](https://en.wikipedia.org/wiki/4_×_400_metres_relay)

This is incomplete and requires a bit more massaging, both in terms of
where and how we get the operator key and in terms of populating the
remainder of the environment in the initcontainer vs via kube config.

Also to be included here: migration of the relay contracts for keep-test
and quite possibly for keep-dev.

PR #591, opened 2020-04-22
keep-prd: Bcoin + Electrum

sthompson22:
Fresh on the heels of https://github.com/keep-network/tbtc/pull/575 we're setting up some mainnet Bcoin deployments.

### Bcoin

This is more or less the same as what we did ^ with some beefcaked configs:

1. Custom StorageClass with SSD's
2. 2 replica StatefulSet
3. Enabled tx indexing on all 3 environments.  This is required by Electrum.  See https://electrumx.readthedocs.io/en/latest/HOWTO.html

### Electrum

Here we provide a production deployment for Electrum.  This looks much like our lesser environments except I've stripped all of the `tbtc` referencing in the resource names.  Notably the Service config is missing.  I'll do this in the next step when we're ready to expose to the internet.

Something else to note here.  keep-dev and keep-test configs are absent from this PR.  This is due to us maintaining those configs with the tbtc-maintainers repo.  As we phase out maintainers I'll shift the Electrum configs, with naming refactors to this repo.

### DIFF
--- a/infrastructure/kube/keep-prd/bcoin-service.yaml
+++ b/infrastructure/kube/keep-prd/bcoin-service.yaml
@@ -1,0 +1,20 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: bcoin
+  namespace: default
+  labels:
+    app: bitcoin-node
+    type: bcoin
+spec:
+  ports:
+  - port: 8332
+    targetPort: 8332
+    name: rpc
+  - port: 8333
+    targetPort: 8333
+    name: mainnet
+  selector:
+    app: bitcoin-node
+    type: bcoin
--- a/infrastructure/kube/keep-prd/bcoin-statefulset.yaml
+++ b/infrastructure/kube/keep-prd/bcoin-statefulset.yaml
@@ -1,0 +1,64 @@
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: bcoin
+  namespace: default
+  labels:
+    app: bitcoin-node
+    type: bcoin
+spec:
+  replicas: 2
+  serviceName: bcoin
+  volumeClaimTemplates:
+  - metadata:
+      name: bcoin-data
+    spec:
+      storageClassName: bcoin
+      accessModes: [ReadWriteOnce]
+      resources:
+        requests:
+          storage: 300Gi
+  selector:
+    matchLabels:
+      app: bitcoin-node
+      type: bcoin
+  template:
+    metadata:
+      labels:
+        app: bitcoin-node
+        type: bcoin
+    spec:
+      containers:
+      - name: bcoin
+        image:  gcr.io/keep-prd-210b/bcoin
+        ports:
+          - containerPort: 8332
+          - containerPort: 8333
+        env:
+          - name: BCOIN_NETWORK
+            value: main
+          - name: BCOIN_HTTP_HOST
+            value: '0.0.0.0'
+          - name: BCOIN_HTTP_PORT
+            value: '8332'
+          - name: BCOIN_PORT
+            value: '8333'
+          - name: BCOIN_LOG_LEVEL
+            value: info
+          - name: BCOIN_PREFIX
+            value: /mnt/.bcoin
+          - name: BCOIN_API_KEY
+            valueFrom:
+              secretKeyRef:
+                name: bcoin
+                key: http-api-key
+        volumeMounts:
+          - name: bcoin-data
+            mountPath: /mnt/.bcoin
+        command:
+          ["bcoin"]
+      volumes:
+      - name: bcoin-data
+        persistentVolumeClaim:
+          claimName: bcoin-data
--- a/infrastructure/kube/keep-prd/bcoin-statefulset.yaml
+++ b/infrastructure/kube/keep-prd/bcoin-statefulset.yaml
@@ -53,6 +53,8 @@ spec:
               secretKeyRef:
                 name: bcoin
                 key: http-api-key
+          - name: BCOIN_INDEX_TX
+            value: 'true'
         volumeMounts:
           - name: bcoin-data
             mountPath: /mnt/.bcoin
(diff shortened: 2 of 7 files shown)


sthompson22:
These are up and syncing

infrastructure/kube/keep-prd/bcoin-service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  ports:
+  - port: 8332
+    targetPort: 8332
+    name: rpc
+  - port: 8333
+    targetPort: 8333
+    name: mainnet
+  selector:
+    app: bitcoin-node
+    type: bcoin
```
Shadowfiend, line 20: The repetitiveness between name, type, labels, and selectors in these configs is slowly driving me insane. These all feel like opportunities for screwups down the line---updating A but not B, resulting in weird behavior that's hard to diagnose. Starting to lean towards some sort of hybrid where we use a generator to create the configs, and then commit the generated configs for good diffing in PRs...

Anyway, way outside the scope of this PR.

sthompson22: I took setting up prd as an opportunity to revisit labels.  The idea is to match lesser environment deployments to these labels, so they're all the same.

sthompson22: If you want to have a separate discussion about labeling patterns I'm happy to do that.

infrastructure/kube/keep-prd/bcoin-statefulset.yaml
```diff
@@ -0,0 +1,66 @@
...
+  serviceName: bcoin
+  volumeClaimTemplates:
+  - metadata:
+      name: bcoin-data
+    spec:
+      storageClassName: bcoin
+      accessModes: [ReadWriteOnce]
+      resources:
+        requests:
+          storage: 450Gi
```
Shadowfiend, line 21: Have we captured somewhere that we should be monitoring this to make sure we can react before we run out of disk?

infrastructure/kube/keep-prd/bcoin-service.yaml
Shadowfiend: I actually like that labels, names, and types are all related, I think it was a good update. What I'm saying instead is that this consistent style may be an opportunity to work to define something once and get a matching YAML, vs having to repeat it 3 times in 20 lines. I realize that can be a slippery slope though; definitely worth a separate discussion.

infrastructure/kube/keep-prd/bcoin-statefulset.yaml
sthompson22: We'll have an alert policy for disk space in GCP Monitoring, otherwise no.

infrastructure/kube/keep-prd/bcoin-statefulset.yaml
```diff
@@ -0,0 +1,66 @@
...
+            value: '8333'
+          - name: BCOIN_LOG_LEVEL
+            value: info
+          - name: BCOIN_PREFIX
+            value: /mnt/.bcoin
+          - name: BCOIN_API_KEY
+            valueFrom:
+              secretKeyRef:
+                name: bcoin
+                key: http-api-key
```
Shadowfiend, line 55: For my own understanding, this tells the bcoin node the API key others will use to interact with it?

infrastructure/kube/keep-prd/bcoin-service.yaml
sthompson22: Ah I see, with you there.  This config layer needs some attention for sure, and I'm not sure Helm is the answer.

I think part of me has intentionally kept this layer as dumb as possible until there's an opportunity to research a more elegant solution.

infrastructure/kube/keep-prd/bcoin-storageclass.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: bcoin
+  namespace: default
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-ssd
+  replication-type: none
+reclaimPolicy: Retain
+allowVolumeExpansion: true
+mountOptions:
+  - debug
```
Shadowfiend, line 13: Again for my own edification, what does debug do as a mount option?

infrastructure/kube/keep-prd/bcoin-statefulset.yaml
sthompson22: Aye.

infrastructure/kube/keep-prd/electrumx-statefulset.yaml
```diff
@@ -0,0 +1,76 @@
...
+  serviceName: electrumx
+  volumeClaimTemplates:
+  - metadata:
+      name: electrumx-data
+    spec:
+      storageClassName: electrumx
+      accessModes: [ReadWriteOnce]
+      resources:
+        requests:
+          storage: 450Gi
```
Shadowfiend, line 21: This is obviously the first time I've taken a deep look at our Electrum deployment heh---does it replicate so much data that it needs the same amount of storage as our main node?

infrastructure/kube/keep-prd/bcoin-statefulset.yaml
Shadowfiend: I think that's all we need, just making sure it's on the radar.

infrastructure/kube/keep-prd/bcoin-storageclass.yaml
sthompson22: This should print debugging info on mount to help troubleshoot if things go sideways.

infrastructure/kube/keep-prd/electrumx-statefulset.yaml
sthompson22: It doesn't, but I wasn't entirely sure of the ratio and I felt like if something we sideways we may not have enough time.

I figured if it was at least as large as bcoin storage we wouldn't have any issues.  A bit of a sledgehammer.

Issue #509, opened 2020-03-03
Funding/redemption script revert error string may be malformed


NicholasDotSol:
On a high-level revert, `returnData` will already include the typical “error selector”. As the scripts propagate this error message, it will add another error selector, which may make it difficult to read the error message.

- [ ] Ensure nested error selectors are handled in external libraries, rather than adding an assembly-level revert to the affected contracts.

liamzebedee:
Some more reference on how the `returnData` for revert is structured -

> The provided string is abi-encoded as if it were a call to a function Error(string). In the above example, revert("Not enough Ether provided."); returns the following hexadecimal as error return data:

> 0x08c379a0                                                         // Function selector for Error(string)
0x0000000000000000000000000000000000000000000000000000000000000020 // Data offset
0x000000000000000000000000000000000000000000000000000000000000001a // String length
0x4e6f7420656e6f7567682045746865722070726f76696465642e000000000000 // String data

https://solidity.readthedocs.io/en/latest/control-structures.html#error-handling-assert-require-revert-and-exceptions

liamzebedee:
@Shadowfiend @NicholasDotSol I don't quite understand the auditor's concern re: adding an assembly-level revert to the affected contracts. Why don't we just slice off the Error function selector (see above) in Solidity?

Shadowfiend:
I prefer this approach I think but never got a chance to poke at it.

Shadowfiend:
I think the audit note was saying that rather than trying to revert from the contracts the scripts were calling in such a way that there would be no nested selector, we should handle the nested selector in the external library---in this case, the scripts---which is what we've chosen to do.

PR #593, opened 2020-04-22
fix: throw nested revert messages

liamzebedee:
Fixes https://github.com/keep-network/tbtc/issues/509

### DIFF
--- a/solidity/contracts/scripts/FundingScript.sol
+++ b/solidity/contracts/scripts/FundingScript.sol
@@ -50,11 +50,20 @@ contract FundingScript {
         // Call the VendingMachine.
         // We could explictly encode the call to vending machine, but this would
         // involve manually parsing _extraData and allocating variables.
-        (bool success, bytes memory returnData) = address(vendingMachine).call(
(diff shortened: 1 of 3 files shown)


solidity/contracts/scripts/FundingScript.sol
```diff
@@ -50,11 +50,20 @@ contract FundingScript {
...
+        // from the contract call.
+        (bool success, bytes memory returnData) = address(vendingMachine).call(_extraData);
+
+        string memory revertMessage;
+        assembly {
+            // A revert message is ABI-encoded as a call to Error(string).
+            // Slicing the Error() signature (4 bytes) and Data offset (4 bytes)
+            // leaves us with a pre-encoded string.
+            // We also slice off the ABI-coded length of returnData (32).
+            revertMessage := add(returnData, 0x44)
```
Shadowfiend, line 63: Could help to articulate the return data length in something more similar to `byte` (e.g. either nibble,  or half byte), or clarify that it's 32 bits. Got confused on the math for a moment heh.

PR #595, opened 2020-04-22
More liquidation tests

NicholasDotSol:
Test liquidation paths more organically. Also test different scenarios based on number of deposits
and price feed fluctuation. The intent is to approximate/simulate expected behavior, and validate outcomes


solidity/contracts/test/deposit/TestDeposit.sol
```diff
@@ -243,6 +243,9 @@ contract TestDeposit is Deposit {
         return self.withdrawalRequestTime;
     }
 
+    function getCollateralizationPercentage() public view returns (uint256) {
+        return self.getCollateralizationPercentage();
+    }
```
Shadowfiend, line 248: This is public now yah?

NicholasDotSol: yeah, and It will be here too once we merge master

solidity/contracts/deposit/DepositLiquidation.sol
```diff
@@ -132,7 +132,7 @@ library DepositLiquidation {
     /// @param  _d              Deposit storage pointer.
     /// @param _txOutputVector  All transaction outputs prepended by the number of outputs encoded as a VarInt, max 0xFC(252) outputs.
     /// @return                 False if output paying redeemer was found, true otherwise.
-    function validateRedeemerNotPaid(
+    function tPaid(
```
Shadowfiend, line 135: Flagging this, but #636 removes it.

solidity/test/liquidation-tests/liquidation-test-utils/createInstance.js
```diff
@@ -0,0 +1,229 @@
...
+    default:
+      throw new Error('required state destination does not have a supported path')
+  }
+}
+
+module.exports.createInstance = createInstance
+module.exports.toActive = toActive
+module.exports.toAwaitingFundingProof = toAwaitingFundingProof
+module.exports.toAwaitingWithdrawalSignature = toAwaitingWithdrawalSignature
+module.exports.toAwaitingWithdrawalProof = toAwaitingWithdrawalProof
```
Shadowfiend, line 228: To what extent can we replace/have we replaced this with the state tests? It feels like it's doing very similar exploration of the state space.

PR #602, opened 2020-05-04
Remove Deposit fallback restriction

NicholasDotSol:
Signer bonds enter the system through the bonding contract, not the backing `ecdsaKeep` address.

Note that removing this also allows anyone to send ether to the Deposit.
Ether sent to this contract by accident will not be retrievable.


Shadowfiend:
Install pre-commit please 🙏 If you need to put it in a setup script (say, [scripts/macos-setup.sh](https://github.com/keep-network/keep-ecdsa/blob/master/scripts/macos-setup.sh) you can run in all your clones, I’ll happily review the PR ;)

solidity/contracts/deposit/Deposit.sol
```diff
@@ -36,7 +36,6 @@ contract Deposit is DepositFactoryAuthority {
     }
 
     function () external payable {
-        require(msg.sender == self.keepAddress, "Deposit contract can only receive ETH from underlying keep");
```
liamzebedee, line 39: I suggest we check for the `msg.data` length, to avoid allowing any unknown function calls to pass through to the fallback function.

`require(msg.data.length == 0, "Deposit contract was called with unknown function selector.");`

liamzebedee: > require(msg.data.length == 0, "Deposit contract was called with unknown function selector.");

Oops, this is only the message data. Looks like you should test if `msg.sig == 0x0` or something of that sort. ([ref](https://ethereum.stackexchange.com/questions/61826/how-to-extract-function-signature-from-msg-data)).

NicholasDotSol: what do you mean by unknown function calls?

NicholasDotSol:
> Install pre-commit please  If you need to put it in a setup script (say, [scripts/macos-setup.sh](https://github.com/keep-network/keep-ecdsa/blob/master/scripts/macos-setup.sh) you can run in all your clones, I’ll happily review the PR ;)

Just set this up on my VM. I ran the lint locally but it does not show any issues found, any thoughts @Shadowfiend

Shadowfiend:
What did you run? `npm run lint` is the ticket, as it will run solium as well as prettier. `npm run lint:fix` is the fixing move. The precommit will only run on changed files unless you tell it otherwise, so that piece will only work if you rebase -i or similar, for example.

solidity/contracts/deposit/Deposit.sol
Shadowfiend: See [this error](https://discordapp.com/channels/590951101600235531/707228822445097051/707249615199535185) from Discord and [my explanation](https://discordapp.com/channels/590951101600235531/707228822445097051/707309632929923175). The fallback runs on any unknown method call, not just on a value send. If we `require` that the signature be `0x0`, we can provide a proper revert in cases where the fallback is called because the called method didn't exist, vs simply a value send.

NicholasDotSol: > Oops, this is only the message data. Looks like you should test if msg.sig == 0x0 or something of that sort. (ref).

msg.data.length is good :)

liamzebedee: @NicholasDotSol, tested it, it's not. Try calling an unknown method with no args and value attached. It will not revert.

liamzebedee: Oh, you're right. Must've stumbled on my end.

solidity/test/DepositUtilsTest.js
```diff
@@ -96,6 +96,34 @@ describe("DepositUtils", async function() {
...
+      // receive value from direct send
+      await testDeposit.send(10)
+    })
+
+    it("Reverts if Deposit is called with unknown function selector", async () => {
+      await expectRevert(
+        web3.eth.sendTransaction({
+          from: accounts[0],
+          to: testDeposit.address,
+          data: "0xdecaf",
```
liamzebedee, line 120: It might be better to test a 4 byte selector here? The encoder probably pads it right with zeroes, though it's good to be explicit where we can.

Shadowfiend: Loldecaf.

liamzebedee: Yeah that's a good one @NicholasDotSol, haven't seen it before :P

liamzebedee:
Just left is updating with master @NicholasDotSol and we're g2g.

Shadowfiend:
Merging this next.

PR #604, opened 2020-05-05
fix: Deposit no longer reverts if ETH transfer is not from Keep

liamzebedee:
We changed the value transfer semantics of the Keep bond. Now the seized bonds originate from the KeepBonding contract. Rather than adding an additional variable to enforce strict transfers from only the KeepBonding contract, we are going to remove the restriction for now.


pdyraga:
Isn't it addressed in https://github.com/keep-network/tbtc/pull/602?

liamzebedee:
Good spot @pdyraga, wasn't aware @NicholasDotSol picked it up.

PR #608, opened 2020-05-06
relay-maintainer: Include TestnetRelay

sthompson22:
### Intro

Summa Relays have two different flavors of Relay.  One for mainnet and one for non-mainnet environments.  Up until this point we've been deploying the mainnet Relay contract to our internal networks for testing.  This has resulted in a relay-maintainer submitting failing difficulty submissions to to the Relay contract.

#### What We Did

- Updated the starting blocks for Bitcoin mainnet and testnet to reduce catch-up time.
- When `network = keep_dev || ropsten` deploy the `TestnetRelay` contract.
- Bump Summa relay-sol package:  The version we were using didn't have `TestnetRelay` included.
- Import `TestnetRelay` contract in `TBTCSystem` so that it gets compiled by Truffle (Migration fails otherwise)
- Set the relay contract address via env var instead of npm package:  Depending on the network we need either the Relay contract address or the `TestnetRelay` contract address.  When we did the original implementation of this InitContainer we were unaware that `TestnetRelay` was a thing.  Rather than get fancy in the InitContainer to conditionally select which package artifact we use, we're just going to set the contract address directly in the Kube config via env var.


Shadowfiend:
Mmm you no have precommit set up.

sthompson22:
> Mmm you no have precommit set up.

fixed

PR #610, opened 2020-05-07
Keep factory load balancer strategy

lukasz-zimnoch:
<s>Depends on https://github.com/keep-network/keep-ecdsa/pull/441</s>

tBTC wants to work with two ECDSA keep factories: 
- based on KEEP stake and ETH bond,
- based on ETH stake and ETH bond.

The first factory is implemented and well-tested, the second one is expected to be ready before the stakedrop. To do not delay the launch of tBTC we make the fully-backed (ETH stake) factory pluggable. We do the same for the strategy that will load-balance between two factories based on their weights.


solidity/contracts/system/TBTCSystem.sol
```diff
@@ -440,7 +441,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         view
         returns (uint256)
     {
-        IBondedECDSAKeepFactory _keepFactory = keepFactoryStrategy.chooseFactory(address(this));
+        IBondedECDSAKeepFactory _keepFactory = keepFactoryStrategy.selectFactory(address(this));
```
eth-r, line 444: I'm not sure we're able to do it this way. The factory that is going to be selected when actually doing a deposit may change, so it would be safest to get the `openKeepFeeEstimate()` through the selection strategy, querying both and taking the maximum if necessary.

Another way to do it would be to have `selectFactory()` return a constant value until it's refreshed with `selectFactoryAndRefresh()` (which returns the value of `selectFactory()` and also performs the selection of the next factory with a new seed). In that case this would work, but we'd need to `selectFactoryAndRefresh()` in `requestNewKeep()`.

.circleci/config.yml
```diff
@@ -177,7 +177,7 @@ jobs:
             echo $GCLOUD_SERVICE_KEY > ~/gcloud-service-key.json
             gcloud auth activate-service-account --key-file ~/gcloud-service-key.json
             export UNISWAP_CONTRACT_DATA="uniswap-contract-data.txt"
-            export BONDED_ECDSA_KEEP_VENDOR_CONTRACT_DATA="BondedECDSAKeepVendor.json"
+            export BONDED_ECDSA_KEEP_FACTORY_CONTRACT_DATA="BondedECDSAFactoryVendor.json"
```
pdyraga, line 180: We need to have a reference to `BondedECDSAKeepFactory.sol`, not the vendor. Even though with the code here we can't update the vendor, we can still update the factory and the whole point is that Keep-org should not be able to do that for tBTC on mainnet.

solidity/contracts/system/IKeepFactorySelector.sol
```diff
@@ -0,0 +1,19 @@
...
+/// @notice Selects an appropriate keep factory for the given application.
+interface IKeepFactorySelector {
+
+    /// @notice Selects keep factory for the given application.
+    /// @param _application Application address.
+    /// @param _regularFactory Regular keep factory.
+    /// @param _fullyBackedFactory Fully backed keep factory.
+    /// @return Selected keep factory.
+    function selectFactory(
+        address _application,
```
pdyraga, line 15: Do we need this parameter? I think the whole point is that the selection strategy is specific to tBTC, especially that this interface is specified in tBTC system contracts. I think it's fair to assume the strategy will be uniquely deployed for tBTC.

solidity/contracts/system/IKeepFactorySelector.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity 0.5.17;
+
+import {IBondedECDSAKeepFactory} from "@keep-network/keep-ecdsa/contracts/api/IBondedECDSAKeepFactory.sol";
+
+/// @title Keep factory selector.
+/// @notice Selects an appropriate keep factory for the given application.
```
pdyraga, line 6: `for the given application.` We have this interface defined in tBTC and the strategy will be for tBTC-only. We do not need to be generic. Application is always tBTC in this context.

solidity/contracts/system/KeepFactoryStrategy.sol
```diff
@@ -0,0 +1,41 @@
...
+    /// @param _application Application address.
+    /// @return Selected keep factory.
+    function selectFactory(
+        Storage storage _self,
+        address _application
+    ) public view returns (IBondedECDSAKeepFactory) {
+        require(
+            address(_self.regularFactory) != address(0),
+            "Regular factory address must be set"
+        );
```
pdyraga, line 26: I think it's fair to assume `regularFactory` is always set.

solidity/contracts/system/KeepFactoryStrategy.sol
```diff
@@ -0,0 +1,41 @@
+pragma solidity 0.5.17;
+
+import {IBondedECDSAKeepFactory} from "@keep-network/keep-ecdsa/contracts/api/IBondedECDSAKeepFactory.sol";
+import "./IKeepFactorySelector.sol";
+
+library KeepFactoryStrategy {
```
pdyraga, line 6: Maybe `KeepFactorySelection` or `KeepFactorySelectionStrategy`?

solidity/contracts/system/TBTCSystem.sol
pdyraga: I like the `selectFactoryAndRefresh` idea. It's cheaper and querying both will result in a question of what should we do with the surplus.

.circleci/config.yml
lukasz-zimnoch: Nice catch! Seems I made an edit error :man_facepalming:

solidity/contracts/system/IKeepFactorySelector.sol
lukasz-zimnoch: Yeah, you're right.

solidity/test/KeepFactorySelectionTest.js
```diff
@@ -0,0 +1,243 @@
...
+    // No selection strategy set.
+    it("returns KEEP stake factory if strategy is not set", async () => {
+      await keepFactorySelection.setFullyBackedKeepFactory(
+        ethStakeFactory.address,
+      )
+      const selected = await keepFactorySelection.selectFactoryAndRefresh.call()
+      await keepFactorySelection.selectFactoryAndRefresh()
+      expect(selected, "unexpected factory selected").to.equal(
+        keepStakeFactory.address,
+      )
```
eth-r, line 123: We could add another test here to check that the selected factory stays the same, as this one is from initialization.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -457,7 +461,35 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+    /// set, KEEP-stake factory is the default choice. When both factories are
+    /// set, as well as keep factory selection strategy, TBTCSystem load
+    /// balances between two factories based on the selection strategy choices.
+    /// @dev Can be called only one time!
+    /// @param _factorySelector Address of the keep factory selection strategy.
+    function setKeepFactorySelector(
+        address _factorySelector
+    ) external onlyOwner {
+        keepFactorySelection.setKeepFactorySelector(_factorySelector);
+    }
```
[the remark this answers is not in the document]
mhluongo: Yes please re the delay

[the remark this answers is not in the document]
pdyraga: OK, I'll make it one function following all governance requirements.

pdyraga:
> Also was the original idea to use a beacon output for selection rather than the keccak approach here?

I am not aware of this requirement. If we use a beacon, we'll increase the price of opening a new deposit twice. I am not sure if we need randomness here. Signers are randomly selected within a pool and here it's more about load balancing between pools than a random selection. I think it's more about round-robin with some weights in this case.

Shadowfiend:
Works for me. I think I'm thinking way back into some of the very early discussions around this in January, perfectly fine if we've moved away from this.

Shadowfiend:
Whoops, forgot to write next steps here. Merging this now, I'm going to merge master into #609 and then implement the time-delayed governance as part of the unified strategy and testing.

PR #616, opened 2020-05-09
Use the full-featured relay contract

prestwich:
Current deploy scripts use the feature-light `Relay` when relaying Bitcoin mainnet. This introduces a bug where the migration passes a  constructor argument that is used by `OnDemandSPV` but not `Relay`. Instead they should use the full-featured `OnDemandSPV` contract.

### DIFF
--- a/solidity/migrations/2_deploy_contracts.js
+++ b/solidity/migrations/2_deploy_contracts.js
@@ -23,7 +23,7 @@ const prices = require("./prices")
 const SatWeiPriceFeed = artifacts.require("SatWeiPriceFeed")
 
 // Bitcoin difficulty relays.
-const Relay = artifacts.require("@summa-tx/relay-sol/contracts/OnDemandSPV")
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Current intent is to deploy `Relay`, since we're not really using the additional features `OnDemandSPV` provides. I forgot to forward-port a fix to the constructor args earlier this week from a Ropsten deploy, as we've been having trouble with our private testnet deploys for other reasons.

Happy to deploy `OnDemandSPV` though; should we be passing the first ID as `2**32`? That seemed like the intent in the relays repo for this contract, but it didn't seem like a crucial aspect.

prestwich:
Definitely use `OnDemandSPV` for forwards compatibility. The _next_ version of tBTC will want it, as it's a big step up over the current tBTC SPV ergonomics.

`firstID` can be anything, so long as it's well-documented what choice you made. It's primarily to simplify request disambiguation when a single maintainer is serving relays on multiple networks/

Shadowfiend:
> ❌ ci/circleci: lint

All right, well, did that to myself. Turboing for now.

PR #617, opened 2020-05-09
Remove FRT holder check from distributeFeeRebate

NicholasDotSol:
Removing this check should be okay since distributing the fee rebate happens as the last value
transfer before an end-state.
This also means that calling distributeFeeRebate needs to be done more carefully.

The reason for this change was to allow for the funding script to correctly return the fee rebate even when `redeemerAddress == rebateTokenHolder.`


liamzebedee:
🙌

PR #618, opened 2020-05-11
Phased Single Shot: Make factory update a single-shot time-delayed change

Shadowfiend:
Exactly what it says, moving #610 to a two-phase time-delayed update. Also ensure it only happens once, while allowing a replacement of the upcoming change. That is, the change can be begun as many times as desired, but can only be finalized once. Once it's finalized, it cannot be begun again.

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -200,11 +200,11 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     {
         require(
             _signerFeeDivisor > 9,
-            "Signer fee divisor must be greater than 9, for a signer fee that is <= 10%."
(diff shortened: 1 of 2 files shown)


solidity/contracts/system/TBTCSystem.sol
```diff
@@ -365,6 +394,35 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+        keepFactorySelection.setFullyBackedKeepFactory(newEthBackedFactory);
+
+        emit KeepFactorySingleShotUpdated(
+            newFactorySelector,
+            newEthBackedFactory
+        );
+
+        keepFactorySingleShotUpdateInitiated = 0;
+        newFactorySelector = address(0);
+        // Keep newEthBackedFactory set as a marker that the update has already
```
NicholasDotSol, line 422: bools? ain't nobody need bools.

PR #619, opened 2020-05-11
relay-maintainer: keep-dev/test adjustments, keep-prd setup

sthompson22:
### Intro

This is really a collection of tweaks to keep-dev/test as we worked through getting the relay-maintainer working, along with the expected setup for keep-prd.

The keep-prd configuration looks different (sadly) because we're not using InitContainers for now in keep-prd.

#### provisioner script

- Made `infuraKey` configurable via env var `INFURA_PROJECT_ID`: For environments that don't use Infura (keep-dev) this isn't required.   If you're using an Infura host without this value set, requests fail with an http `400`.

#### keep-dev

- Added `INFURA_PROJECT_ID` per provisioner script notes.  Set to empty string since we're using a local ETH chain in keep-dev.
- Added the latest relay contract address

#### keep-test

- Use the hand surgery version of the relay-maintainer InitContainer to allow for configuration of `infuraKey`.  This will go away once the changes to the InitContainer made here propagate through the keep-test build/deploy cycle.
- Use the correct eth-network ConfigMap (internal -> ropsten)
- Set latest relay contract address

#### keep-prd

- Configure Deployment with setting the `SUMMA_*` env vars directly in the Deployment config file.
- Everything here is baked into Secrets for now.


sthompson22:
Drafting until the keep-prd config is proven.

sthompson22:
Something else to note here now that we're using Alchemy...I'm not sure the relay maintainer works with Alchemy as-is.  We can continue to use Infura for this app, feels weird though.

infrastructure/kube/keep-test/relay-maintainer-deployment.yaml
```diff
@@ -23,47 +23,50 @@ spec:
...
-                  name: eth-network-internal
+                  name: eth-network-ropsten
                   key: network-id
 
             - name: ETH_NETWORK_NAME
-              value: internal
+              value: ropsten
+
+            - name: INFURA_PROJECT_ID
+              value: 414a548bc7434bbfb7a135b694b15aa4
```
Shadowfiend, line 46: Do we want to pull this to a secret?

sthompson22: Part of me was hoping this would be temporary enough that it wouldn't matter...but you're right, should be done anyway.  I'll move it.

PR #620, opened 2020-05-12
AwaitingFundingProof to Active state transition

NicholasDotSol:
moving from signer setup to active breaks previous state tests


NicholasDotSol:
closing, was just pushed as reference

PR #622, opened 2020-05-12
Active to awaitingWithdrawalSignature transition

NicholasDotSol:
Change the transaction inputs to use a real Ropsten transaction.
This allows for a seamless transition to redemption, since no forced state transitions are needed.


NicholasDotSol:
closed in favor of #629

PR #623, opened 2020-05-13
Locked Price Feed: Lock down Bitcoin price access

Shadowfiend:
With the intent of using an access-controlled Maker feed, the price has
to be guarded in our contracts in turn.

Access control happens on `SatWeiPriceFeed` (which only allows `TBTCSystem`
to call `getPrice`) and on `TBTCSystem` (which only allows deposits to call
`fetchBitcoinPrice`).

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -483,10 +483,16 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     }
 
     /// @notice Get the price of one satoshi in wei.
-    /// @dev Reverts if the price of one satoshi is 0 wei, or
-    ///      if the price of one satoshi is 1 ether.
+    /// @dev Reverts if the price of one satoshi is 0 wei, or if the price of
+    ///      one satoshi is 1 ether. Can only be called by a deposit with minted
+    ///      TDT.
     /// @return The price of one satoshi in wei.
     function fetchBitcoinPrice() external view returns (uint256) {
+        require(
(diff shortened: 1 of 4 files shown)


solidity/contracts/system/TBTCSystem.sol
```diff
@@ -483,10 +483,16 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     }
 
     /// @notice Get the price of one satoshi in wei.
-    /// @dev Reverts if the price of one satoshi is 0 wei, or
-    ///      if the price of one satoshi is 1 ether.
+    /// @dev Reverts if the price of one satoshi is 0 wei, or if the price of
+    ///      one satoshi is 1 ether. Can only be called by a deposit with minted
+    ///      TDT.
```
NicholasDotSol, line 488: `Can only be called by a deposit with minted TDT` -> `Can only be called by a Deposit contract`?

Shadowfiend: Bit torn on this one, but I think this is more accurate than the actual revert message… The revert message is more consistent and understandable, though. I'm comfortable with using the specific but slightly more complicated explanation in `@dev`.

NicholasDotSol: i'm fine with it. just thought the comment might allow for someone to infer that a deposit can possibly not have a minted TDT.

Shadowfiend: Yeah, fair. We can always field that question if it comes up.

PR #630, opened 2020-05-15
Final initial governance values

Shadowfiend:
Two tweaks:
- Governance time delay goes from 6 hours to 48 hours.
- Signing fees are allowed as low as 2 basis points/0.02%.

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -186,8 +186,8 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
             "Signer fee divisor must be greater than 9, for a signer fee that is <= 10%"
         );
         require(
-            _signerFeeDivisor < 2000,
-            "Signer fee divisor must be less than 2000, for a signer fee that is > 0.05%"
+            _signerFeeDivisor < 5000,
+            "Signer fee divisor must be less than 5000, for a signer fee that is > 0.02%"
         );
 
         newSignerFeeDivisor = _signerFeeDivisor;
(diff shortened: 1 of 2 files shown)


afmsavage:
Would you be open to more discussion around governance delays?  Is that time delay (48h) for how long between submission and the final tally?  From a trader/investor point of view, it is really hard to keep up with everything as far as different projects and what's happening day to day with them.  If a voting period is only 48h long, there will be a lot of people that miss this timeframe and do not vote.  This opens the voting up to manipulation.

I also know that a timeframe for governance benefits from a shorter timeframe in order to be more agile.  If there is an issue, you do not want 30days to go by before its possible to make a change that is affecting the system.

mhluongo:
> If a voting period is only 48h long, there will be a lot of people that miss this timeframe and do not vote. This opens the voting up to manipulation.

Actually, we've sort of done the opposite that I rthink you imagine- governance increases attack surface, so we have as little as possible.

There's no voting, it's a multisig privilegedaccount- and the changes only impact new deposits. I agree if this needed coordination it would need to be way longer, but "governance" here can only
* change fees for new deposits
* change collateral ration for new deposits
* change lot sizes for new deposits
* propose a price feed fallback, which has a 90-day delay
* and pull an emergency "stop new deosits" lever in case of a 0-day, which can only be called one.

No contract upgrades or changing any existing deposits, nothing really to vote on- but we can put an additional delay on at the wallet level after the fact by switching it for a contract.

Check out [this post](https://tbtc.network/developers/tbtc-security-model/) and see if you're still uncomfortable.

afmsavage:
No more concern @mhluongo !  I did think of governance as a catch all term that could alert more important levers such as governance on Maker.  48h is more than sufficient now that I looked through that link.

mhluongo:
Shwew! We've tried to be sensitive to this stuff in particular so glad you agree.

PR #634, opened 2020-05-19
Ensure bond is sufficient to potentially refund openKeep fee 

NicholasDotSol:
Revert on deposit creation if the keep does not have sufficient bond to refund open-keep fee in case of signer setup failure.


solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -438,9 +438,12 @@ library DepositUtils {
...
         require(address(this).balance >= _ethValue, "Not enough funds to send");
-        IBondedECDSAKeep _keep = IBondedECDSAKeep(_d.keepAddress);
-        _keep.returnPartialSignerBonds.value(_ethValue)();
-        return true;
+        if(_ethValue > 0){
+            IBondedECDSAKeep _keep = IBondedECDSAKeep(_d.keepAddress);
+            _keep.returnPartialSignerBonds.value(_ethValue)();
+            return true;
+        }
+        return false;
```
Shadowfiend, line 446: This is going to fall under the return thing we're doing in the other PR, but slightly different: this function only used to ever return true… But also no one pays attention to the return value. I think we can drop the return value.

solidity/test/DepositFundingTest.js
```diff
@@ -125,6 +130,27 @@ describe("DepositFunding", async function() {
       )
     })
 
+    it("reverts if bond is insufficient to cover a deposit creation fee refund", async () => {
+      const depositFee = await tbtcSystemStub.getNewDepositFeeEstimate()
+
+      await ecdsaKeepStub.send(depositFee - 1)
+      await ecdsaKeepStub.setBondAmount(depositFee - 1)
```
Shadowfiend, line 137: Head-whirly that we don't have to mess with the ETHBTC price here, but more illustrative of how complicated all the bits are.

NicholasDotSol: yeah it's definitely gotten incrementally more complicate

solidity/contracts/deposit/DepositFunding.sol
```diff
@@ -93,9 +96,11 @@ library DepositFunding {
         // refund the deposit owner the cost to create a new Deposit at the time the Deposit was opened.
         uint256 _seized = _d.seizeSignerBonds();
 
-        /* solium-disable-next-line security/no-send */
-        _d.enableWithdrawal(_d.depositOwner(), _d.keepSetupFee);
-        _d.pushFundsToKeepGroup(_seized.sub(_d.keepSetupFee));
+        if(_seized >= _d.keepSetupFee){
+            /* solium-disable-next-line security/no-send */
```
Shadowfiend, line 100: Well this obviously no longer applies :p

PR #636, opened 2020-05-20
Bugfixes related to redeemer output scripts

prestwich:
- require that an output script can be parsed as a sh, pkh, wpkh, or wsh
- fix script equality check in `redemptionTransactionChecks`
- remove dead codepath `validateRedeemerNotPaid`
- remove unnecessary abi.encodePacked for bytes types

~~Does not include test vectors.~~ Does now.

Specific things to consider:
- To ensure L100-101 is implemented properly, verify that redeemer output script is stored with its length prefix (e.g. `0x1976a914...` not `0x76a914`
- The check on L352 may be overkill. Typically this check would be made to prevent reverts due to buffer overrun. I've included it out of an abundance of caution as it is gas-cheap.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -71,7 +71,16 @@ contract MySystem {
 
 == Security
 
-tBTC is currently under audit.
+tBTC's first 6-week audit was completed by ConsenSys Diligence on March 27,
+2020, against https://github.com/keep-network/tbtc/commit/fbb2018c41456d19ec20eb28a17070ee2b10eb5d[fbb2018c41].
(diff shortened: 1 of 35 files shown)


prestwich:
Tests are passing locally. Linter crashes instead of producing error output tho. Output scripts are a good candidate for fuzz/prop testing, as they will always follow a constrained set of templates.

Shadowfiend:
Merge walked over a couple of variable declarations, 'bout to get that sorted.

PR #640, opened 2020-05-24
Restraining Bolt: Limit the TBTC supply for the first 4 months

mhluongo:
> [Restraining bolts](https://starwars.fandom.com/wiki/Restraining_bolt) were small, cylindrical devices that could be affixed to a droid in order to limit its functions and enforce its obedience.

-- [Wookipedia](https://starwars.fandom.com/wiki/Restraining_bolt)

Limit the TBTC supply to the following schedule, as well as adding a 1-day "grace period" to allow signers to ramp up.

| Month | Cap (BTC) |
| ----- | --------- |
| 1     | 100       |
| 2     | 250       |
| 3     | 500       |
| 4     | 1000      |
| 5+    | 21M       |

This approach limits the fungible supply cap, but doesn't prevent an unlimited number of deposits, so long as the TDTs are held. Restricted deposits based on TDTs would be more robust to circumvention, but the complexity costs more than the assurance, IMO.

I ended up storing the `TBTCToken` contract reference directly on `TBTCSystem`. Open to refactoring if anyone has an issue with that.

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -63,6 +63,8 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     uint256 pausedTimestamp;
     uint256 constant pausedDuration = 10 days;
 
+    TBTCToken public tbtcToken;
+
     ISatWeiPriceFeed public priceFeed;
     IRelay public relay;
 
@@ -142,13 +144,56 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
             _keepThreshold,
             _keepSize
         );
+        tbtcToken = _tbtcToken;
         setTbtcDepositToken(_tbtcDepositToken);
         initializedTimestamp = block.timestamp;
         allowNewDeposits = true;
     }
 
-    /// @notice gets whether new deposits are allowed.
-    function getAllowNewDeposits() external view returns (bool) { return allowNewDeposits; }
+    /// @notice returns whether new deposits should be allowed.
+    function getAllowNewDeposits() external view returns (bool) {
+        if (initializedTimestamp == 0 || !allowNewDeposits) {
+            return false;
+        }
+
+        uint256 age = block.timestamp - initializedTimestamp;
+
+        uint256 supply = tbtcToken.totalSupply();
+        uint256 bufferedSupply = supply.add(getMaxLotSize().mul(10 ** 10));
+
+        if(age < 1 days) {
+            return bufferedSupply < 2 * 10 ** 18;
+        }
+
+        if (age < 30 days) {
+            return bufferedSupply < 100 * 10 ** 18;
+        }
+
+        if (age < 60 days) {
+            return bufferedSupply < 250 * 10 ** 18;
+        }
+
+        if (age < 90 days) {
+            return bufferedSupply < 500 * 10 ** 18;
+        }
+
+        if (age < 120 days) {
+            return bufferedSupply < 1000 * 10 ** 18;
+        }
+
+        return allowNewDeposits;
+    }
+
+    // @notice get the largest lot size currently enabled for deposits.
+    function getMaxLotSize() public view returns (uint256) {
+        uint256 max = 0;
+        for (uint i = 0; i<lotSizesSatoshis.length; i++) {
+            if (lotSizesSatoshis[i] > max) {
+                max = lotSizesSatoshis[i];
+            }
+        }
+        return max;
+    }
 
     /// @notice One-time-use emergency function to disallow future deposit creation for 10 days.
     function emergencyPauseNewDeposits() external onlyOwner returns (bool) {
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -44,7 +44,6 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         uint256 _timestamp
     );
 
-
     event EthBtcPriceFeedAdded(address _priceFeed);
     event LotSizesUpdated(uint64[] _lotSizes);
     event AllowNewDepositsUpdated(bool _allowNewDeposits);
@@ -59,11 +58,11 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         address _ethBackedFactory
     );
 
-    uint256 initializedTimestamp = 0;
+    bool initialized = false;
     uint256 pausedTimestamp;
     uint256 constant pausedDuration = 10 days;
 
-    TBTCToken public tbtcToken;
+    VendingMachine public vendingMachine;
 
     ISatWeiPriceFeed public priceFeed;
     IRelay public relay;
@@ -125,7 +124,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         uint16 _keepThreshold,
         uint16 _keepSize
     ) external onlyOwner {
-        require(initializedTimestamp == 0, "already initialized");
+        require(!initialized, "already initialized");
 
         keepFactorySelection.initialize(_defaultKeepFactory);
 
@@ -144,49 +143,20 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
             _keepThreshold,
             _keepSize
         );
-        tbtcToken = _tbtcToken;
+        vendingMachine = _vendingMachine;
         setTbtcDepositToken(_tbtcDepositToken);
-        initializedTimestamp = block.timestamp;
+        initialized = true;
         allowNewDeposits = true;
     }
 
     /// @notice returns whether new deposits should be allowed.
     function getAllowNewDeposits() external view returns (bool) {
-        if (initializedTimestamp == 0 || !allowNewDeposits) {
-            return false;
-        }
+        if (!allowNewDeposits) { return false; }
 
-        uint256 supply = tbtcToken.totalSupply();
+        uint256 supply = vendingMachine.getMintedSupply();
         uint256 bufferedSupply = supply.add(getMaxLotSize().mul(10 ** 10));
 
-        return bufferedSupply <= getMaxSupply() && allowNewDeposits;
-    }
-
-    // @notice get the maximum TBTC token supply in BTC * 10 ** 18
-    function getMaxSupply() public view returns (uint256) {
-        uint256 age = block.timestamp - initializedTimestamp;
-
-        if(age < 1 days) {
-            return 2 * 10 ** 18;
-        }
-
-        if (age < 30 days) {
-            return 100 * 10 ** 18;
-        }
-
-        if (age < 60 days) {
-            return 250 * 10 ** 18;
-        }
-
-        if (age < 90 days) {
-            return 500 * 10 ** 18;
-        }
-
-        if (age < 120 days) {
-            return 1000 * 10 ** 18;
-        }
-
-        return 21000000 * 10 ** 18;
+        return bufferedSupply <= vendingMachine.getMaxSupply();
     }
 
     // @notice get the largest lot size currently enabled for deposits, in satoshis
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -156,35 +156,40 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
             return false;
         }
 
-        uint256 age = block.timestamp - initializedTimestamp;
-
         uint256 supply = tbtcToken.totalSupply();
         uint256 bufferedSupply = supply.add(getMaxLotSize().mul(10 ** 10));
 
+        return bufferedSupply <= getMaxSupply() && allowNewDeposits;
+    }
+
+    // @notice get the maximum TBTC token supply in BTC * 10 ** 18
+    function getMaxSupply() public view returns (uint256) {
+        uint256 age = block.timestamp - initializedTimestamp;
+
         if(age < 1 days) {
-            return bufferedSupply < 2 * 10 ** 18;
+            return 2 * 10 ** 18;
         }
 
         if (age < 30 days) {
-            return bufferedSupply < 100 * 10 ** 18;
+            return 100 * 10 ** 18;
         }
 
         if (age < 60 days) {
-            return bufferedSupply < 250 * 10 ** 18;
+            return 250 * 10 ** 18;
         }
 
         if (age < 90 days) {
-            return bufferedSupply < 500 * 10 ** 18;
+            return 500 * 10 ** 18;
         }
 
         if (age < 120 days) {
-            return bufferedSupply < 1000 * 10 ** 18;
+            return 1000 * 10 ** 18;
         }
 
-        return allowNewDeposits;
+        return 21000000 * 10 ** 18;
     }
 
-    // @notice get the largest lot size currently enabled for deposits.
+    // @notice get the largest lot size currently enabled for deposits, in satoshis
     function getMaxLotSize() public view returns (uint256) {
         uint256 max = 0;
         for (uint i = 0; i<lotSizesSatoshis.length; i++) {
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -59,7 +59,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         address _ethBackedFactory
     );
 
-    bool _initialized = false;
+    uint256 initializedTimestamp = 0;
     uint256 pausedTimestamp;
     uint256 constant pausedDuration = 10 days;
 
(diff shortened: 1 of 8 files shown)


mhluongo:
Ahh, git hooks - I'll get the linting fixed up.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -142,13 +144,61 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
-        _initialized = true;
+        initializedTimestamp = block.timestamp;
         allowNewDeposits = true;
     }
 
-    /// @notice gets whether new deposits are allowed.
-    function getAllowNewDeposits() external view returns (bool) { return allowNewDeposits; }
+    /// @notice returns whether new deposits should be allowed.
+    function getAllowNewDeposits() external view returns (bool) {
+        if (initializedTimestamp == 0 || !allowNewDeposits) {
```
Shadowfiend, line 155: If `initializedTimestamp == 0` then `allowNewDeposits == false`, as the only write to it is in `initialize`, which sets `allowNewDeposits` to `true`.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -142,13 +144,61 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+            return false;
+        }
+
+        uint256 supply = tbtcToken.totalSupply();
```
Shadowfiend, line 159: We can probably access `tbtcToken` directly on the vending machine, fwiw. No strong opinion on whether that's better or worse… It seems better encapsulated to consider the vending machine the source of truth on supply from `TBTCSystem`'s perspective, or even the ultimate arbiter of permitted supply. One scenario falls out from this line of reasoning…

Say there are 120 BTC under custody in the first month, and the 100th TBTC is minted. This hits the supply cap. `TBTCSystem` disallows further deposits until some TBTC is taken out of circulation. Should the vending machine still allow the minting of the remaining 20 BTC into TBTC?

If the answer is no, then perhaps what we need here is for the vending machine to track the permitted TBTC supply, refuse to mint further TBTC when the supply cap is reached, and expose e.g. `isTBTCMintingAllowed` or similar. Then `TBTCSystem` uses that function, which encapsulates the vending machine's assessment of the supply cap, as the additional factor here. This has the additional effect of removing the direct `TBTCToken` dependency from `TBTCSystem`.

solidity/test/TBTCSystemTest.js
```diff
@@ -9,6 +9,10 @@ const TBTCSystem = contract.fromArtifact("TBTCSystem")
 const SatWeiPriceFeed = contract.fromArtifact("SatWeiPriceFeed")
 const MockMedianizer = contract.fromArtifact("MockMedianizer")
 
+function btcToBTC18(n) {
```
Shadowfiend, line 12: This is really btcToTbtc, no?

solidity/contracts/system/TBTCSystem.sol
mhluongo: Right now :) I'd rather keep the semantic distinction unless you think the gas matters.

solidity/test/TBTCSystemTest.js
mhluongo: If that's the convention we're following in the codebase, yeah. I wish we had an explicit name for 10^-18 BTC unit though...

solidity/contracts/system/TBTCSystem.sol
mhluongo: Downsides of doing this in the vending machine:

* What we actually want is to stop new deposits to limit risk, not just stop minting. This is a close proxy without wider system changes
* The dApp would need to check the vending machine and refuse new deposits of the supply cap was hit

Upsides

* Cleaner implementation, perhaps, and nice to not add a token contract reference?

I think the best solution without additional bookkeeping might be "both"- stopping new deposits and additional minting. Unclear to me whether it's worth the effort though, and I want to make as few modifications as possible.

Shadowfiend: That’s what I meant: having the vending machine guard minting and expose the current stare through a public function, and having TBTCSystem use that public function with the reference it already has to guard deposits as well.

solidity/test/TBTCSystemTest.js
mhluongo: Addressed.

mhluongo:
This is ready for another look @Shadowfiend

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -142,13 +143,32 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+        vendingMachine = _vendingMachine;
         setTbtcDepositToken(_tbtcDepositToken);
-        _initialized = true;
+        initialized = true;
         allowNewDeposits = true;
     }
 
-    /// @notice gets whether new deposits are allowed.
-    function getAllowNewDeposits() external view returns (bool) { return allowNewDeposits; }
+    /// @notice returns whether new deposits should be allowed.
```
Shadowfiend, line 152: Sentences start with caps, we should probably give more detail on how this is decided, and we should probably set `@return` as well.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -142,13 +143,32 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+    function getAllowNewDeposits() external view returns (bool) {
+        if (!allowNewDeposits) { return false; }
+
+        uint256 supply = vendingMachine.getMintedSupply();
+        uint256 bufferedSupply = supply.add(getMaxLotSize().mul(10 ** 10));
+
+        return bufferedSupply <= vendingMachine.getMaxSupply();
+    }
+
+    // @notice get the largest lot size currently enabled for deposits, in satoshis
```
Shadowfiend, line 162: Sentences  have grammar 😬

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -142,13 +143,32 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
-    /// @notice gets whether new deposits are allowed.
-    function getAllowNewDeposits() external view returns (bool) { return allowNewDeposits; }
+    /// @notice returns whether new deposits should be allowed.
+    function getAllowNewDeposits() external view returns (bool) {
+        if (!allowNewDeposits) { return false; }
+
+        uint256 supply = vendingMachine.getMintedSupply();
+        uint256 bufferedSupply = supply.add(getMaxLotSize().mul(10 ** 10));
+
+        return bufferedSupply <= vendingMachine.getMaxSupply();
```
Shadowfiend, line 159: We’re comfortable allowing no new deposits if the largest available lot would push us over the cap is the goal here right?

solidity/contracts/system/VendingMachine.sol
```diff
@@ -74,11 +111,13 @@ contract VendingMachine is TBTCSystemAuthority{
 
         tbtcDepositToken.transferFrom(msg.sender, address(this), _tdtId);
 
-        // If the backing Deposit does not have a signer fee in escrow, mint it.
         Deposit deposit = Deposit(address(uint160(_tdtId)));
         uint256 signerFee = deposit.signerFee();
         uint256 depositValue = deposit.lotSizeTbtc();
 
+        require(getMintedSupply() + depositValue < getMaxSupply(), "Can't mint more than the max supply cap");
```
Shadowfiend, line 118: Tbh it feels cleaner and clearer to have a `canMint(uint256 amount)` helper that does this check, and hook up this require and the `getAllowNewDeposits` check to that helper. Probably just bikeshedding at this point though.

solidity/test/DepositFundingTest.js
```diff
@@ -155,6 +160,78 @@ describe("DepositFunding", async function() {
...
+      )
+
+      await increaseTime(30 * 24 * 60 * 60) // 30 days, into the 5th month
+      await createNewDeposit(fullBtc)
+      await mint("2099900000000000") // should new be at 21M BTC - 1000 sats
+      await expectRevert(
+        createNewDeposit(fullBtc),
+        "Opening new deposits is currently disabled.",
+      )
+    })
```
Shadowfiend, line 234: I don’t now that I love bundling this all into one big test , but it’ll do.

solidity/test/TBTCSystemTest.js
```diff
@@ -336,6 +336,14 @@ describe("TBTCSystem", async function() {
       expect(allowNewDeposits).to.equal(false)
     })
 
+    it("pauses new deposit creation after the 1-day grace period", async () => {
+      await increaseTime(24 * 60 * 60 + 1)
+      await tbtcSystem.emergencyPauseNewDeposits()
+
+      const allowNewDeposits = await tbtcSystem.getAllowNewDeposits()
+      expect(allowNewDeposits).to.equal(false)
```
Shadowfiend, line 344: I’m confused... what is this checking? Don’t fully understand the grace period either, I think —the PR description doesn’t really explain what it is meant to do practically, just something about “ramping up” for signers?

solidity/test/TBTCSystemTest.js
```diff
@@ -388,13 +396,6 @@ describe("TBTCSystem", async function() {
       await restoreSnapshot()
     })
 
-    it("pauses new deposit creation", async () => {
-      await tbtcSystem.emergencyPauseNewDeposits()
-
-      const allowNewDeposits = await tbtcSystem.getAllowNewDeposits()
-      expect(allowNewDeposits).to.equal(false)
-    })
```
Shadowfiend, line 396: 🥴

solidity/test/VendingMachineTest.js
```diff
@@ -130,6 +134,43 @@ describe("VendingMachine", async function() {
...
     })
+
+    it("updates the minted supply", async () => {
+      const mintedSupply = await vendingMachine.getMintedSupply()
+
+      await tbtcDepositToken.forceMint(owner, tdtId)
+      await tbtcDepositToken.approve(vendingMachine.address, tdtId, {
+        from: owner,
+      })
+      await vendingMachine.tdtToTbtc(tdtId, {from: owner})
```
Shadowfiend, line 145: Feels like a strangely grey box way to test this. Why not just force some TBTC minting and ensure the VM reports it?

solidity/test/VendingMachineTest.js
```diff
@@ -626,4 +667,83 @@ describe("VendingMachine", async function() {
...
+      let maxSupply = await vendingMachine.getMaxSupply()
+      expect(maxSupply).to.eq.BN(btcToTbtc(100))
+
+      await increaseTime(15 * 24 * 60 * 60) // 15 days
+      maxSupply = await vendingMachine.getMaxSupply()
+      expect(maxSupply).to.eq.BN(btcToTbtc(100))
+
+      await increaseTime(14 * 24 * 60 * 60 - 10 * 600) // 14 days minus 10 minutes
+      maxSupply = await vendingMachine.getMaxSupply()
+      expect(maxSupply).to.eq.BN(btcToTbtc(100))
```
Shadowfiend, line 706: You test that past the end time the value is not the expected one in the previous test. Why not do that here as well?

solidity/contracts/system/VendingMachine.sol
mhluongo: I'd rather extract the logic out of the `require` 🤷‍♂️

solidity/test/TBTCSystemTest.js
mhluongo: It began as a 0 supply cap, which meant `getAllowNewDeposits()` was false. It also had the side-effect of breaking a bunch of our tests.

I updated the tests, then realized a 1 BTC limit was actually more appropriate for day 1.

solidity/contracts/system/VendingMachine.sol
NicholasDotSol: `+` -> `.add()`

solidity/contracts/system/TBTCSystem.sol
mhluongo: No longer relevant

solidity/contracts/system/TBTCSystem.sol
mhluongo: Correct.

solidity/test/VendingMachineTest.js
mhluongo: I was trying to cover a realistic code path, but if there's a reason to prefer `TBTCToken.frozenMint()` I'm open to it

solidity/test/TBTCSystemTest.js
mhluongo: Adding.. more tests here. I want to show that adding the supply cap doesn't restrict the emergency pause.

solidity/test/VendingMachineTest.js
Shadowfiend: Mostly it's that all tests here are white-box and do force minting, so doing something completely different makes it seem like there's a underlying reason why this wouldn't work the way the other tests would. Not a huge deal.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -142,13 +143,32 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
         allowNewDeposits = true;
     }
 
-    /// @notice gets whether new deposits are allowed.
-    function getAllowNewDeposits() external view returns (bool) { return allowNewDeposits; }
+    /// @notice Returns whether new deposits should be allowed.
+    /// @return True if new deposits should be allowed, both by the emergency pause button
+    ///         and respected the max supply schedule.
+    function getAllowNewDeposits() external view returns (bool) {
+        if (!allowNewDeposits) { return false; }
```
Shadowfiend, line 156: Could this just be `return allowNewDeposits && ...`? Understand if you settled on this approach due to a clarity preference though.

solidity/contracts/system/VendingMachine.sol
```diff
@@ -20,9 +20,46 @@ contract VendingMachine is TBTCSystemAuthority{
...
+    uint256 createdAt;
+
     constructor(address _systemAddress)
         TBTCSystemAuthority(_systemAddress)
-    public {}
+    public {
+        createdAt = block.timestamp;
+    }
+
+    /// @notice return the outstanding minted TBTC supply
```
Shadowfiend, line 31: Still with the non-sentence sentences :p

solidity/contracts/system/VendingMachine.sol
```diff
@@ -20,9 +20,46 @@ contract VendingMachine is TBTCSystemAuthority{
...
+    function getMintedSupply() public view returns (uint256) {
+        return tbtcToken.totalSupply();
+    }
+
+    /// @notice get the maximum TBTC token supply in BTC * 10 ** 18 based on the
+    /// age of the contract deployment
```
Shadowfiend, line 37: Still some room here to elaborate in English IMO.

solidity/test/DepositFundingTest.js
```diff
@@ -155,6 +160,78 @@ describe("DepositFunding", async function() {
         "Opening new deposits is currently disabled.",
       )
     })
+
+    it("respects the supply cap schedule", async () => {
+      const bn = web3.utils.toBN
```
Shadowfiend, line 165: I'll say it bugs me just a touch that you've special-purposed a helper here, but this is such a pervasive issue that I understand why you didn't bring the hammer down :laughing:

mhluongo: It's the worst thing. I can't stand `BN`

solidity/contracts/system/VendingMachine.sol
mhluongo: Do you want a translation of the method into English? Fine with me, just unclear what level we're looking for

solidity/contracts/system/TBTCSystem.sol
mhluongo: This was actually because I assume it's cheaper gas-wise. I didn't confirm, but I wouldn't be surprised to see the Solidity compiler not properly unroll the short-circuiting

solidity/test/DepositFundingTest.js
```diff
@@ -155,6 +160,78 @@ describe("DepositFunding", async function() {
...
+        createNewDeposit(fullBtc),
+        "Opening new deposits is currently disabled.",
+      )
+
+      await increaseTime(30 * 24 * 60 * 60) // 30 days, into the 5th month
+      await createNewDeposit(fullBtc)
+      await mint("2099900000000000") // should new be at 21M BTC - 1000 sats
+      await expectRevert(
+        createNewDeposit(fullBtc),
+        "Opening new deposits is currently disabled.",
```
Shadowfiend, line 232: This does feel like a weird require message for a supply cap hit 🤔

Shadowfiend:
Btw, way to force-push and make it impossible for me to do a diff on the changes you've made since my previous reviews so I have to go through the whole diff again 😝

mhluongo:
> Btw, way to force-push and make it impossible for me to do a diff on the changes you've made

Not my finest moment :sweat_smile:

solidity/contracts/system/TBTCSystem.sol
Shadowfiend: ~30 gas less to do it in one statement. Not worth the time to change it tbh.

solidity/contracts/system/VendingMachine.sol
Shadowfiend: Eh, I think there's room to include the supply schedule itself in brief English, yeah. Beautiful revision IMO :)

PR #641, opened 2020-05-24
Who watches the watchmen? Disable the emergency pause after 12 months

mhluongo:
> **Quis custodiet ipsos custodes?** is a Latin phrase found in the work of the Roman poet Juvenal from his Satires (Satire VI, lines 347–348). It is literally translated as "Who will guard the guards themselves?", though it is also known by variant translations, such as "Who watches the watchers?" and "Who will watch the watchmen?".

-- [Wikipedia](https://en.wikipedia.org/wiki/Quis_custodiet_ipsos_custodes%3F)

365 days after contract initialization, stop responding to `emergencyPauseNewDeposits()`. Remaining governance functions are unaffected.

Based off #640.


Shadowfiend:
May want to target this branch at `restraining-bolt` and mark it as a draft until that lands.

mhluongo:
Ready for ya @Shadowfiend

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -173,6 +173,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     /// @notice One-time-use emergency function to disallow future deposit creation for 10 days.
     function emergencyPauseNewDeposits() external onlyOwner returns (bool) {
         require(pausedTimestamp == 0, "emergencyPauseNewDeposits can only be called once");
+        require(block.timestamp - initializedTimestamp < 365 days, "Can only be called within 365 days of initialization");
```
Shadowfiend, line 176: We should be consistent with the previous `require` and say e.g. `emergencyPauseNewDeposits can only be called…` (or flip the other one, either way).

mhluongo: I prefer including the method name, but we have a linter rule for long error messages.

Shadowfiend: Interesting. Looks like that’s an ethlint default. Feels like we should change it...

mhluongo: Addressed in https://github.com/keep-network/solium-config-keep/pull/1

mhluongo:
Good for another look on green @Shadowfiend

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -173,6 +173,8 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     /// @notice One-time-use emergency function to disallow future deposit creation for 10 days.
     function emergencyPauseNewDeposits() external onlyOwner {
         require(pausedTimestamp == 0, "emergencyPauseNewDeposits can only be called once");
+        uint256 diff = block.timestamp - initializedTimestamp;
```
Shadowfiend, line 176: I mean I’ll say if you’re going to pull this out of the context of the require, giving it a super generic name like `diff` is kinda meh, but we can roll with it 😉

PR #654, opened 2020-05-26
Initial fix for lot size delay circumvention

### LINKED ISSUE #648: Governance delay is ignored when updating lot sizes
# Description
When updating lot sizes, both `newLotSizesSatoshis` and `lotSizesSatoshis` are written to, defeating the purpose of the governance delay.

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/system/TBTCSystem.sol#L205-L227

# Exploit Scenario
An attacker or rogue admin hijacks the governance key and instantly updates the lot size, allowing them to grief the system. They may also be able to exploit other code by enabling new lot sizes.

# Suggested Remediation
Delete the line setting `lotSizesSatoshis`.

mhluongo:
Working on properly testing the governance delay invariant across all
tests.

Closes #648


Shadowfiend:
If you're in a refactoring mood, I would love to pull governance tests out of `TBTCSystemTest.js` into their own file.

High-level for the test, I think we need to split `verifyFinalization` into the event expectation and the state expectations (right now the one function does both), and then expect the state assertions to fail (via `expect(...).to.throw` as a test under [`when initiating the update`](https://github.com/keep-network/tbtc/blob/master/solidity/test/TBTCSystemTest.js#L236) or something of the sort. Hope that helps a little <_<

mhluongo:
> Hope that helps a little <_<

A little :grinning:

Shadowfiend:
@mhluongo Did this for you. @NicholasDotSol could use your eyes on it. You can spot-check https://github.com/keep-network/tbtc/pull/654/commits/293ef0e84a115f577da6976e5d89a58cbedd4ef0 if you want, but it's a mechanical move commit that splits governance tests to their own file. The meat is in the following two commits, where tests are adjusted to check that state verification fails after initialization finishes, and succeeds after finalization, for every change.

Notably, I undid the fix here to make sure the tests failed :wink:

mhluongo:
Thank you! I was pretty unhappy with the various test rearrangements I'd gone through.

NicholasDotSol:
I like the changes and happy to see this reorganized since we didn't fit it in the first PR. Thanks
@mhluongo and @Shadowfiend :)

mhluongo:
I'll merge, since this is basically @Shadowfiend 's PR now :smile:

Issue #651, opened 2020-05-26
The owner of a TDT which has been courtesy called needs to own TBTC to redeem it


samczsun:
# Description

If a TDT is in courtesy call, then it will calculate a redemption requirement equal to `lotSize`
https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L58

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositUtils.sol#L469-L476

Then the TDT will transfer `lotSize` TBTC from the sender to the owner of the TDT
https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L76-L87

If the sender is also the owner, then they are required to own `lotSize` TBTC in order to redeem a TDT which has entered the courtesy call state, even though they are the owner.

# Exploit Scenario
Alice is holding a TDT worth 1 BTC which has just been courtesy called. In order to redeem her TDT, she will need to temporarily acquire 1 TBTC.

PR #656, opened 2020-05-27
Redemption requirement refactor

NicholasDotSol:
splitting logic to calculate redemption requirement into 3 internal functions:

`mainCharge` - the TBTC owed to redeem
`frtAdjust` -  fee rebate owed
`signerFeeAdjust` -  signer fee owed

This should lay out the logic more clearly, help with removing logic overlap, remove complex assumptions and allow for fixing unintended issues.

closes: #651


solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -448,30 +448,68 @@ library DepositUtils {
...
-                return fee;
-            }
-        }
-        uint256 contractTbtcBalance = _d.tbtcToken.balanceOf(address(this));
-        if(contractTbtcBalance < fee) {
-            return fee.sub(contractTbtcBalance);
-        }
-        return 0;
+        uint256 frtAdjust = frtAdjust(_d, _redeemer);
+        uint256 signerFeeAdjust = signerFeeAdjust(_d, _redeemer);
```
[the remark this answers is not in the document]
NicholasDotSol: I'm good with those names

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -448,30 +448,68 @@ library DepositUtils {
...
+        uint256 mainCharge = mainCharge(_d, _redeemer);
+        uint256 frtAdjust = frtAdjust(_d, _redeemer);
+        uint256 signerFeeAdjust = signerFeeAdjust(_d, _redeemer);
+        return mainCharge.add(frtAdjust).add(signerFeeAdjust);
+    }
+
+    /// @notice             Get the base TBTC amount needed to redeem.
+    /// @param _redeemer    The deposit redeemer.
+    /// @return             The amount in TBTC.
+    function mainCharge(DepositUtils.Deposit storage _d, address _redeemer) internal view returns (uint256){
```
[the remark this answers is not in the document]
NicholasDotSol: I thought of doing exactly this, but decided against it since I thought it would just be more code. On second thought it's really not too bad, and does make things a little more clear

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -448,30 +448,68 @@ library DepositUtils {
...
+
+    /// @notice             Get the fee rebate amount needed to redeem.
+    /// @param _redeemer    The deposit redeemer.
+    /// @return             The amount in TBTC.
+    function frtAdjust(DepositUtils.Deposit storage _d, address _redeemer) internal view returns (uint256){
+        if(
+            _d.inCourtesyCall() ||
+            remainingTerm(_d) == 0 ||
+            feeRebateTokenHolder(_d) == address(0) ||
+            feeRebateTokenHolder(_d) == _redeemer
```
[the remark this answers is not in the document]
NicholasDotSol: `bool redeemerPaysFee` -> `bool redeemerPaysNoFee`
since no fee is owed if wither is true

Not sure if passing this as a param is really going to help clarify things, we just move the 2 lines to parent.

Reusing this bool on the signer-fee adjustment calculation makes this more palatable 

We would also still need to check if FRT exists since we can't rely on just `redeemerHoldsFrt`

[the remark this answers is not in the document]
NicholasDotSol: also:
`ownerRedemption` -> `tdtOwnerRedemption`
`redeemerHoldsFrt` -> `frtOwnerRedemption`

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -458,56 +464,70 @@ library DepositUtils {
...
-    function mainCharge(DepositUtils.Deposit storage _d, address _redeemer) internal view returns (uint256){
         bool inCourtesy = _d.inCourtesyCall();
-        bool ownerRedemption = depositOwner(_d) == _redeemer;
         bool atTerm = remainingTerm(_d) == 0;
+        bool redeemerPaysNoFee = inCourtesy || atTerm;
+
+        require(inCourtesy || tdtOwnerRedemption || atTerm, "Only TDT holder can redeem unless deposit is at-term or in COURTESY_CALL");
 
-        require(inCourtesy || ownerRedemption || atTerm, "Only TDT holder can redeem unless deposit is at-term or in COURTESY_CALL");
+        uint256 mainCharge = computeBaseRedemptionCharge(_d, tdtOwnerRedemption);
```
Shadowfiend, line 477: `mainCharge` here should be `baseRedemptionCharge` for consistency. Same with `frtAdjust` -> `frtAdjustment`, `signerFeeAdjust` -> `signerFeeAdjustment`.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -458,56 +464,70 @@ library DepositUtils {
...
-    function frtAdjust(DepositUtils.Deposit storage _d, address _redeemer) internal view returns (uint256){
+    /// @notice                    Get the fee rebate amount needed to redeem.
+    /// @param _redeemerPaysNoFee  True if Deposit is at term or in courtesy call.
+    /// @param _frtOwnerRedemption True if the redeemer is the FRT holder.
+    /// @param _frtExists          True if the FRT exists.
+    /// @return                    The amount in TBTC.
+    function computeFrtAdjustment(
+        DepositUtils.Deposit storage _d,
+        bool _redeemerPaysNoFee,
+        bool _frtOwnerRedemption,
```
Shadowfiend, line 504: Thoughts on aligning the structure here for clarity? `_redeemerPaysNoFee`, `_redeemerHoldsFrt`, `_frtExists`. I would also rather say `Owes` rather than `Pays`. This whole path is computing what is owed, someone else is in charge of payment. Semantic distinction, I know, but I think it also boosts clarity a touch.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -458,56 +464,70 @@ library DepositUtils {
...
+    function computeSignerFeeAdjustment(
+        DepositUtils.Deposit storage _d,
+        bool _redeemerPaysNoFee,
+        bool _frtExists
+    ) internal view returns (uint256){
+        if(_redeemerPaysNoFee){
             return 0;
         }
-        if(feeRebateTokenHolder(_d) != address(0)){
+        if(_frtExists){
```
Shadowfiend, line 530: We should settle on either the structure in `computeFrtAdjustment` (single if that returns 0) or the one here (multiple ifs that all return 0), but let's stick to one or the other.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -458,56 +464,70 @@ library DepositUtils {
...
-    /// @notice             Get the base TBTC amount needed to redeem.
-    /// @param _redeemer    The deposit redeemer.
-    /// @return             The amount in TBTC.
-    function mainCharge(DepositUtils.Deposit storage _d, address _redeemer) internal view returns (uint256){
         bool inCourtesy = _d.inCourtesyCall();
-        bool ownerRedemption = depositOwner(_d) == _redeemer;
         bool atTerm = remainingTerm(_d) == 0;
+        bool redeemerPaysNoFee = inCourtesy || atTerm;
+
+        require(inCourtesy || tdtOwnerRedemption || atTerm, "Only TDT holder can redeem unless deposit is at-term or in COURTESY_CALL");
```
Shadowfiend, line 475: Let's put the owner redemption check up front, and I'm interested here as well in a common structure to the others, which would be `redeemerHoldsTdt` or similar.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -448,8 +448,14 @@ library DepositUtils {
...
-        uint256 frtAdjust = frtAdjust(_d, _redeemer);
-        uint256 signerFeeAdjust = signerFeeAdjust(_d, _redeemer);
+        bool inCourtesy = _d.inCourtesyCall();
+        bool atTerm = remainingTerm(_d) == 0;
+        bool redeemerPaysNoFee = inCourtesy || atTerm;
+        bool frtOwnerRedemption = feeRebateTokenHolder(_d) == _redeemer;
+        bool frtExists = feeRebateTokenHolder(_d) != address(0);
+
+        uint256 frtAdjust = computeFrtAdjustment(_d, redeemerPaysNoFee, frtOwnerRedemption, frtExists);
+        uint256 signerFeeAdjust = computeSignerFeeAdjustment(_d, redeemerPaysNoFee, frtExists);
```
Shadowfiend, line 458: Am I reading this wrong that this is identical to `getRedemptionTbtcRequirement` except with `tdtOwnerRedemption` always `true`? If so, I think we can stack these?

```solidity
function getOwnerRedemptionTbtcRequirement(DepositUtils.Deposit storage _d, address _redeemer) internal pure returns (uint256) {
    // The TDT owner only ever owes fee adjustments.
    return computeRedemptionFeeAdjustments(_d, _redeemer);
}

function getRedemptionTbtcRequirement(DepositUtils.Deposit storage _d, address _redeemer) internal view returns (uint256) {
    bool redeemerHoldsTdt = depositOwner(_d) == _redeemer;
    bool inCourtesy = _d.inCourtesyCall();
    bool atTerm = remainingTerm(_d) == 0;

    require(redeemerHoldsTdt || inCourtesy || atTerm, "Only TDT holder can redeem unless deposit is at-term or courtesy called");

    uint256 mainCharge = computeBaseRedemptionCharge(_d, redeemerHoldsTdt);    
    return mainCharge.add(computeRedemptionFeeAdjustments(_d, _redeemer));
}

function computeRedemptionFeeAdjustments(DepositUtils.Deposit storage _d, address _redeemer) internal pure returns (uint256) {
    bool inCourtesy = _d.inCourtesyCall();
    bool atTerm = remainingTerm(_d) == 0;
    bool redeemerOwesNoFee = inCourtesy || atTerm;
    bool redeemerHoldsFrt = feeRebateTokenHolder(_d) == _redeemer;
    bool frtExists = feeRebateTokenHolder(_d) != address(0);

    uint256 frtAdjustment = computeFrtAdjustment(_d, redeemerOwesNoFee, redeemerHoldsFrt, frtExists)
    uint256 signerFeeAdjustment = computeSignerFeeAdjustment(_d, redeemerOwesNoFee, frtExists)

    return frtAdjustment.add(signerFeeAdjustment)
}
```

Shadowfiend: In fact, I would go so far as to make even `getRedemptionTbtcRequirement` be `pure`, and move the redeemerHoldsTdt || inCourtesy || atTerm require all the way up to `_requestRedemption`, honestly, but that may be too much right now.

NicholasDotSol: I'm not sure if grouping the fee adjustment functions is good. Won't it just make things more expensive? also not sure if it helps in terms of clarity

Shadowfiend: Having two paths that do the exact same thing is definitely counter to clarity heh. We want our code to reflect the path similarity, and the split between getOwner and getRedemptionTbtcRequirement is artificial in the first place in service of the vending machine. It's considerably more difficult without this change to look at the code and understand just how similar these paths are, IMO.

NicholasDotSol: > It's considerably more difficult without this change to look at the code and understand just how similar these paths are, IMO.

you're right here

NicholasDotSol: re-calculating those bools within the same function chain bugs me a little

Shadowfiend: I agree 😬 Just didn’t see an immediately obvious cleanup. I think it’s reflective of the broader problem of *needing* these two functions, tbh. The reason we need them is that `require`, which we don’t want to trigger in the owner function. The owner function is a simple calculation, the other one is imposing restrictions. This is where pulling the `require` further up would allow this code to be focused purely on calculation, but it also has some knock on effects that we may not want to play all the way out. If you want to give it a shot though, you’re welcome to :)

Shadowfiend:
@NicholasDotSol let me know when you want to pull this out of draft; it feels close (though I haven't gone back through the tests again yet).

Shadowfiend:
This is on my merge list for today.

solidity/test/DepositRedemptionTest.js
```diff
@@ -182,37 +169,46 @@ describe("DepositRedemption", async function() {
...
-      const expectedFee = new BN(100)
-      await feeRebateToken.transferFrom(frtHolder, tdtHolder, tdtId, {
+    it("returns SignerFee if deposit is pre-term, FRT does not exist and signer fee is partially escrowed", async () => {
+      await feeRebateToken.burn(web3.utils.toBN(testDeposit.address), {
         from: frtHolder,
       })
-      await tbtcToken.forceMint(testDeposit.address, signerFee.sub(expectedFee))
 
-      const tbtcOwed = await testDeposit.getOwnerRedemptionTbtcRequirement.call(
+      await tbtcToken.forceMint(testDeposit.address, signerFee)
```
Shadowfiend, line 206: Flagging: this will change in #680.

solidity/test/DepositRedemptionTest.js
```diff
@@ -234,15 +230,38 @@ describe("DepositRedemption", async function() {
...
-    it("returns SignerFee if we are at-term, caller is TDT holder, and fee is not escrowed", async () => {
+    it("returns zero if we are at-term, caller is TDT holder, and fee is not escrowed", async () => {
       await increaseTime(depositTerm)
+      await feeRebateToken.burn(web3.utils.toBN(testDeposit.address), {
+        from: frtHolder,
+      })
+      const tbtcOwed = await testDeposit.getRedemptionTbtcRequirement.call(
+        tdtHolder,
+      )
+      expect(tbtcOwed).to.eq.BN(new BN(0))
```
Shadowfiend, line 241: 🤔 who pays the signers in this scenario? This disagrees with row 6 of [the spec’s table](https://docs.keep.network/tbtc/#_redemption_payment_and_disbursal_scenarios) and is because I did the flow diagram wrong. The TDT holder always owes the signer fee, even at term, if it’s not escrowed.

On the flip side, we can eliminate the FRT logic if we start adjusting the owed TBTC by the already-escrowed balance in the general case (as we’ve discussed doing for #680), since the existence of the FRT is only being used as a signal of whether or not the fee is escrowed. Relying on the actual escrow amount seems safer in the first place.

solidity/test/DepositRedemptionTest.js
```diff
@@ -234,15 +230,38 @@ describe("DepositRedemption", async function() {
...
 
+    it("returns zero if we are in courtesy_call, caller is TDT holder and fee is not escrowed", async () => {
+      await testDeposit.setState(states.COURTESY_CALL)
+      await feeRebateToken.burn(web3.utils.toBN(testDeposit.address), {
+        from: frtHolder,
+      })
+      const tbtcOwed = await testDeposit.getRedemptionTbtcRequirement.call(
+        tdtHolder,
+      )
+      expect(tbtcOwed).to.eq.BN(new BN(0))
```
Shadowfiend, line 262: This would also be incorrect, if the above is. Should be signerFee if that’s true.

solidity/test/DepositRedemptionTest.js
```diff
@@ -340,16 +359,17 @@ describe("DepositRedemption", async function() {
       expect(events[0].returnValues.value).to.eq.BN(depositValue)
     })
 
-    it("escrows correct fee if deposit is pre-term, owner is TDT holder and signer fee is partially escrowed", async () => {
-      const expectedFee = new BN(100)
-      await tbtcToken.forceMint(testDeposit.address, signerFee.sub(expectedFee))
-      await feeRebateToken.transferFrom(accounts[4], owner, tdtId, {
-        from: accounts[4],
+    it("escrows signerFee if deposit is pre-term, owner is TDT holder and signer fee is partially escrowed", async () => {
```
Shadowfiend, line 362: This will also need to be adjusted with the work we’re doing on #680.

solidity/test/liquidation-tests/redemptionProofTimeoutTest.js
```diff
@@ -35,7 +35,7 @@ describe("Integration -- Redemption-proof timeout", async function () {
     it("liquidates correctly", async () => {
      await toAwaitingWithdrawalProof(testDeposit)
       //  AWAITING_WITHDRAWAL_SIGNATURE -> AWAITING_WITHDRAWAL_PROOF
-      const requirement = await testDeposit.getOwnerRedemptionTbtcRequirement.call(depositInitiator)
+      const requirement = await testDeposit.getRedemptionTbtcRequirement.call(depositInitiator)
```
Shadowfiend, line 38: Why are we expecting on this value below at all? It seems irrelevant to the test, no? Ditto for the two below.

solidity/test/DepositRedemptionTest.js
NicholasDotSol: will adjust these with the relevant PR

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -489,8 +489,10 @@ library DepositUtils {
         bool frtExists = feeRebateTokenHolder(_d) != address(0);
 
         uint256 frtAdjustment = computeFrtAdjustment(_d, redeemerOwesNoFee, redeemerHoldsFrt, frtExists);
```
Shadowfiend, line 491: Sadly this means redeemerOwesNoFee is no longer the correct term here. Maybe thirdPartyRedeemerOwesNoFee?

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -489,8 +489,10 @@ library DepositUtils {
...
-        uint256 signerFeeAdjustment = computeSignerFeeAdjustment(_d, redeemerOwesNoFee, frtExists);
-
+        uint256 signerFeeAdjustment = computeSignerFeeAdjustment(_d, frtExists);
+        if(depositOwner(_d) != _redeemer){
+            return 0;
+        }
```
Shadowfiend, line 495: Not following this bit...

NicholasDotSol: understandable, I also wanted a sanity check here.
If we don't return 0 here and redeemer is deposit owner, we could have a situation where `getRedemptionTbtcRequirement` returns a value greater than lotSize (eg: when the fee rebate token does not exist and TDT owner is the redeemer)

to be clear, I really don't like it here

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: ah yes

solidity/contracts/deposit/DepositUtils.sol
Shadowfiend: I think the only scenario here is if a non-TDT holder (owes lot size) redeems pre term/non courtesy call (owes fee) and there is no escrow (no FRT adjustment). But we prohibit that scenario because only the TDT holder is allowed to redeem pre-term. Does that sound right? See the table I linked above for the scenarios, as well.

NicholasDotSol: This is regarding at-term and courtesy-call cases.
If redeemer is not TDT owner, and FRT does not exist.
`computeSignerFeeAdjustment` just checks FRT existance and returns fee is it doesn't exist.
so `computeBaseRedemptionCharge` will return `LostSize`, and without the check `computeRedemptionFeeAdjustments` will return `signerFee`. the fee shouldn't be greater then `lotSize`

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: how about `redeemerOwesNoFeeRebate`?

solidity/contracts/deposit/DepositUtils.sol
Shadowfiend: I see. Let me think through this a little more.

Also, I just realized `computeFrtAdjustment` and `computeSignerFeeAdjustment` have their names reversed, , which is making this harder to reason about. Let's fix that.

Shadowfiend: Reconsidering the whole thing here, how about we drop this FRT adjustment vs signer fee adjustment thing altogether, and instead go for a single `computeSignerFee` function:

```solidity
function computeSignerFee(
    DepositUtils.Deposit storage _d,
    bool _thirdPartyRedeemerOwesNoFee,
    bool _redeemerHoldsFrt,
    bool _frtExists
) internal view returns (uint256) {
    bool signerFeeNotOwed;
    if (_preTerm) {
        signerFeeNotOwed = _redeemerHoldsFrt;
    } else {
        signerFeeNotOwed = _frtExists || _redeemerIsThirdParty;
    }

    if (signerFeeNotOwed) {
        return 0;
    } else {
        return signerFee(_d);
    }
}
```

NicholasDotSol: This sounds like a great change overall

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -446,30 +446,76 @@ library DepositUtils {
...
+    /// @param _frtExists            True if the FRT exists.
+    /// @return                      The fees owed in TBTC.
+    function computeSignerFee(
+        DepositUtils.Deposit storage _d,
+        bool _preTerm,
+        bool _redeemerHoldsFrt,
+        bool _redeemerIsThirdParty,
+        bool _frtExists
+    ) internal view returns (uint256) {
+        bool signerFeeNotOwed;
```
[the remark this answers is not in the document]
NicholasDotSol: I think I prefer `redeemerOwesNoFee`

PR #659, opened 2020-05-27
Parametrize relay deployment

lukasz-zimnoch:
Here we introduce some improvements to the Truffle deployment script which allows passing some relay maintainer parameters from the outside. This feature will be used by https://github.com/keep-network/local-setup.


solidity/migrations/relay-config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "forceOnDemandSPV": false,
+  "forceTestnetRelay": false,
+  "genesis": {
```
nkuba, line 4: We already have `genesis` property under `bitcoinMain` and `bitcoinTest`.
```suggestion
  "init": {
```

solidity/migrations/relay-config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "forceOnDemandSPV": false,
+  "forceTestnetRelay": false,
```
nkuba, line 3: I don't like having these two separate properties as we use them in the script in a manner similar to cases for a switch statement. Thoughts on having one property here with the following possibilities:
```json
  "forceRelay": "OnDemandSPV",
```
```json
  "forceRelay": "TestnetRelay",
```
By default, the property won't be provided in the `relay-config.json`, we will set this only when we need it.

lukasz-zimnoch: Good point! I will change.

Issue #653, opened 2020-05-26
Users can bypass paying the fee rebate to the FRT holder


samczsun:
# Description
Typically, a user will purchase a TDT from the vending machine then redeem the TDT, paying `signerFee` to the holder of the FRT. However, the fee is charged depending on the redeemer, which can be set by using `transferAndRequestRedemption`

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/Deposit.sol#L163-L174

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L148-L157

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L93-L104

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositUtils.sol#L450-L463

# Exploit scenario
Alice creates a deposit for 1 BTC. She escrows the TDT with the vending machine receiving 0.9995 TBTC. Eve purchases the TDT for 1 TBTC and calls `transferAndRequestRedemption` with Alice's address. Eve receives the 1 BTC and Alice receives the finalized TDT. Alice does not receive the 0.0005 TBTC that she should have received.

Note that by choosing this, Eve takes on the risk of the keep performing fraud and not signing the withdrawal transaction. Eve also loses out on potential bonds if the redemption proof is not submitted in time.

# Suggested Remediation
Remove the ability to perform a transfer and redemption, or change the logic involved in calculating fees owed.

PR #661, opened 2020-05-28
Restrict transferAndRequestRedemption to VendingMachine caller

NicholasDotSol:
closes: https://github.com/keep-network/tbtc/issues/653

The linked issue above outlines an exploit where `transferAndRequestRedemption` is used to avoid paying a fee rebate.
Restricting this function to the `VendingMaching` and controlling the input should remove this possibility for now.


solidity/contracts/deposit/Deposit.sol
```diff
@@ -165,6 +165,10 @@ contract Deposit is DepositFactoryAuthority {
         bytes memory _redeemerOutputScript,
         address payable _finalRecipient
     ) public returns (bool) {
+        require(
+            msg.sender == self.vendingMachineAddress,
+            "Only the vendingMachine can call transferAndRequestRedemption"
```
Shadowfiend, line 170: Let's call this “the vending machine”, maybe?

solidity/contracts/system/VendingMachine.sol
```diff
@@ -203,6 +201,6 @@ contract VendingMachine is TBTCSystemAuthority{
             tbtcToken.approve(_depositAddress, tbtcOwed);
         }
 
-        _d.transferAndRequestRedemption(_outputValueBytes, _redeemerOutputScript, _finalRecipient);
+        _d.transferAndRequestRedemption(_outputValueBytes, _redeemerOutputScript, msg.sender);
```
Shadowfiend, line 204: Just a sanity check: our tests are sufficient to ensure this doesn't break anything, right?

solidity/test/DepositRedemptionTest.js
```diff
@@ -674,6 +674,80 @@ describe("DepositRedemption", async function() {
...
+        from: owner,
+      })
+    })
+
+    afterEach(async () => {
+      await restoreSnapshot()
+    })
+
+    it("updates state successfully and fires a RedemptionRequested event", async () => {
+      await testDeposit.setVendingMachineAddress(owner)
```
Shadowfiend, line 713: Why make all the changes to be able to `setVendingMachineAddress` rather than just reading the vending machine address and making the transactions `{from: vendingMachine.address}`?

solidity/test/DepositRedemptionTest.js
```diff
@@ -674,6 +674,80 @@ describe("DepositRedemption", async function() {
...
+      const blockNumber = await web3.eth.getBlockNumber()
+
+      await testDeposit.setSigningGroupPublicKey(keepPubkeyX, keepPubkeyY)
+
+      // the fee is 2.86331153 BTC
```
Shadowfiend, line 718: Is this… Relevant?

solidity/test/DepositRedemptionTest.js
```diff
@@ -674,6 +674,80 @@ describe("DepositRedemption", async function() {
...
+  describe("transferAndRequestRedemption", async () => {
+    const sighash =
+      "0xb68a6378ddb770a82ae4779a915f0a447da7d753630f8dd3b00be8638677dd90"
+    const outpoint = "0x" + "33".repeat(36)
+    const valueBytes = "0x1111111111111111"
+    const keepPubkeyX = "0x" + "33".repeat(32)
+    const keepPubkeyY = "0x" + "44".repeat(32)
+    // Override redeemer output script for this test.
+    // No real reason here, it's just how we derived the below values.
+    const redeemerOutputScript = "0x160014" + "33".repeat(20)
```
Shadowfiend, line 686: I'm guessing this isn't a new vector… Where are we reusing it from, and how can we avoid duplicating the test vector?

solidity/contracts/system/VendingMachine.sol
NicholasDotSol: should be, yes.
This just restricts functionality, and instasnces where the function was called we were passing owner as `_finalRecipient` (so the same functinality as msg.sender). Basically, we should be better than we were before

solidity/test/DepositRedemptionTest.js
NicholasDotSol: don't think we can send a transaction from an address we do not control

solidity/test/DepositRedemptionTest.js
NicholasDotSol: slightly, the test vector is directly copy from `requestRedemption`, since they use pretty much the exact same logic, I figured not changing the test scenario was appropriate (including comments)

solidity/test/DepositRedemptionTest.js
NicholasDotSol: it is indeed not. It's from `requestRedemption` right above. Why do we want to avoid duplicating this test vector? It's a path that can be taken towards redemption and it should be run fully imo. It also ensures any future changes to this function don't interfere with the redemption intent, or at least catches intent

Shadowfiend: For me, understanding all the test vectors we use is a way to understand how good our coverage is of various paths. Duplicate vectors testing different paths basically give us an opportunity to drift vectors away from each other without questioning why the divergence is happening. Having them centralized lets us inspect any decision to change a vector in the context of what other tests are using that vector, and understand why both tests can bear the change or why we might need a completely separate vector in a given case. 

Somewhat relatedly, it also lets the test be as descriptive as possible by making the code focus on test intent rather than setup. 

Of course all of this assumes the vectors are named reasonably well as well—encapsulating what the vector represents. I’ve tried to do that with our existing shared vectors in utils, though we still have a generically-named “TX” that “does not spark joy” heh.

Shadowfiend: We never circled back here---I'm comfortable with merging this as is and doing a PR to fix the duplication, though.

PR #667, opened 2020-06-01
Helper to get funding UTXO info

### LINKED ISSUE #660: Naming utxoSize
This is completely subjective, so feel free to flag as wont-fix.

I would avoid referring to "size" when you are talking about value, since this can lead to some misconceptions that could end up leading to bugs.

I've seen this for `lotSizeSatoshis`, `lotSizeTbtc` and `utxoSize`, but it is specially relevant for the latter. Since a UTXO is a struct containing both a script and a value, the size is normally used to refer to the length in bytes of the struct, but not to the value hold by it.

### LINKED ISSUE #666: Expose funding UTXO info on Deposit
Currently, extracting the funding UXTO information is cumbersome. Exposing this would be useful for building explorers etc. 

See [comment in tbtc.js for context](https://github.com/keep-network/tbtc.js/issues/33).

liamzebedee:
Currently, extracting the funding UXTO information is cumbersome. Exposing this would be useful for building explorers etc.

Closes #666, closes #660.


Shadowfiend:
Ok well I went slightly crazy here and changed `getFundingInfo` to `fundingInfo` (to align with other getters), exposed this all the way on the deposit contract itself, and went ahead and rolled in #660 because it was there for the taking with these changes.

Back to y'all for review + merge.

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -472,4 +472,10 @@ library DepositUtils {
         require(remainingTerm(_d) == 0 || inCourtesy, "Only TDT holder can redeem unless deposit is at-term or in COURTESY_CALL");
         return lotSizeTbtc(_d);
     }
+
+    /// @notice Returns information associated with the funding UXTO.
+    /// @return A tuple of (uxtoSizeBytes, fundedAt, uxtoOutpoint)
+    function fundingInfo(Deposit storage _d) public view returns (bytes8 utxoValueBytes, uint256 fundedAt, bytes memory utxoOutpoint) {
```
NicholasDotSol, line 478: we would only be using this as a return for `Deposit.fundingInfo` if this PR didn't remove `getUTXOInfo`.
Why not keep `getUTXOInfo` and implement `fundingInfo` directly in `Deposit` without having to call `depositUtils`?

NicholasDotSol: Also, the returned tuple can contain zero-values, do we want to include a `@dev` comment about these cases?

Shadowfiend: Adding a comment sounds like a great idea.

Not sure I followed the other bit. `getUTXOInfo` was only removed from the `TestDeposit` contract. Are you suggesting that rather than add anything to `fundingInfo`, we just have the one at the `Deposit` level return `self.utxoValueBytes, self.fundedAt, self.utxoOutpoint`? If so, I'm not necessarily opposed, but it gets away from an existing pattern we've used for `utxoValue`, `lotSizeTbtc`, etc, so I'd like to have a clear idea of when we follow the pattern of implementing on `DepositUtils` and when we don't before we make the change.

NicholasDotSol: > we just have the one at the Deposit level return self.utxoValueBytes, self.fundedAt, self.utxoOutpoint?

right, we wouldn't need the `depositUtils` function. 

> but it gets away from an existing pattern we've used for utxoValue

How so? we're returning directly from the Deposit contract without involving the library.

> I'd like to have a clear idea of when we follow the pattern of implementing on DepositUtils and when we don't before we make the change.

implementing on DepositUtils makes sense when the function will be used interanlly, but when it is a simple read function that has no interaction I think it makes sense to keep it simple and implement directly on Deposit. We do the same for commercialization values.

Shadowfiend: You are completely right :laughing: My original reaction to your thought was “yes!”, then I went through the code to see if we'd used the same pattern elsewhere, and I think I got turned around between Deposit and DepositUtils.

Kk, will push some commits to make both of these changes… Hopefully today.

Shadowfiend:
Ok, pushed [those changes](https://github.com/keep-network/tbtc/pull/667#discussion_r442380639).

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: re comment, we can add something like
```
/// @dev The returned tuple might contain one or more zero-values
///      since the values are all set at different points in 
///      the deposit life-cycle
```

Shadowfiend: Whoops, yep.

Shadowfiend: Hmmm… I wonder if we should `require` that the state not be `START`, `AWAITING_SIGNER_SETUP`, or `AWAITING_BTC_FUNDING_PROOF`?

Shadowfiend: Going with yes, push coming momentarily.

Shadowfiend: This should be pushed as well.

PR #668, opened 2020-06-02
Pre-script-ion: Various adjustments to accepted funding and redemption scripts

### LINKED ISSUE #658: requestRedemption accepts invalid _redeemerOutputScripts
# Description 

Invalid `_redeemerOutputScripts` looking like properly formatted redeemscripts can be passed to `requestRedemption` without being caught. This affects to the 4 output types supported (`P2PKH`, `P2SH`, `P2WPKH` and `P2WSH`).

This is supposed to be guarded by [BCUtils.extractHash](https://github.com/summa-tx/bitcoin-spv/blob/master/solidity/contracts/BTCUtils.sol#L406), but the actual size of the script is never checked.

The root of the issue is the same for segwit and legacy outputs, the encoding of the script size is assumed to be properly set. For legacy, it is done by checking that 20 bytes are pushed to the stack (`0x14`) [but not checking the actual size of the script](https://github.com/summa-tx/bitcoin-spv/blob/master/solidity/contracts/BTCUtils.sol#L418-L434). For Segwit scripts, the size of the script is parsed from the output, but [the actual size is not computed either](https://github.com/summa-tx/bitcoin-spv/blob/master/solidity/contracts/BTCUtils.sol#L406-L417) (the expected size is compared with the parsed size, but not with the actual size).

# Exploit Scenario

Call requestRedemption with a script using the following format:

- P2PKH: `1976a914 <script_with_more_than_20_bytes> 88ac` 
- P2SH: `17a914 <script_with_more_than_20_bytes> 87` 
- P2PKH: `0014 <script_with_more_than_20_bytes>` 
- P2PKH: `0020 <script_with_more_than_32_bytes>` 

### Invalid P2PKH example
```
_outputValueBytes = 4062b00700000000
_redeemerOutputScript = 1976a914f86f0bc0a2232970ccdf4569815db500f12683618888ac
```

This script contains an additional byte (`88`) before the script suffix (`88ac`).

# Mitigation

Force a check of the script size on top of checking that the script template is properly formatted (this should be done in BCUtils.extractHash. Will open a linked issue there too).

# Extra information

Notice `extractHash` is also used in several other parts of the code. However, it is normally guarded by `validateVout`, which will catch a transaction including this type of script, since they are incorrectly formatted. Also, this transaction won't be accepted in a block, so this should not trigger if the provided data comes from a block or can be checked agains a block / known valid transaction. However, in this case the data is provided directly from the user, so no additional checks are run.

### LINKED ISSUE #660: Naming utxoSize

### LINKED ISSUE #664: P2PKH and P2SH outputs are accepted for funding proofs
# Description
When providing a funding proof, three types of outputs are accepted: P2WPKH, P2PKH, and P2SH outputs. The system is designed to handle the first and fails for the second and third.

If a P2PKH output is provided, the signing group is able to generate a scriptSig which can spend the output. However, that will trigger the ECDSA fraud flow as the signature wasn't authorized.

If a P2SH output is provided, then the public key of the keep will be interpreted as a script. This most likely results in a locked output.

# Exploit Scenario
Eve creates a deposit, funds it using a P2PKH or P2SH output, and requests redemption. The signing group is unable to provide a redemption proof, and eventually Eve seizes the group's bonds.

# Suggested Remediation
Disallow P2PKH and P2SH outputs when providing funding proofs.

Shadowfiend:
Restrict funding proofs to accepting only p2wpkh outputs. Also, further restrict
redemption output scripts to ensure the length of the output script is validated
by bumping to bitcoin-spv-sol 3.1.0, which includes this check in `extractHash`.
Prior, output scripts were checked for adherence to the standard templates, but
their declared lengths were not verified.

Fixes #658, fixes #664.


Shadowfiend:
Won't be moving on the rest of this until a week from Monday most likely.

Shadowfiend:
Okayyy, finally circled back here. Updated the PR description, and we are ready for review.

Shadowfiend:
Hm… Wonder why my local prettier pass didn't catch the formatting issues 🤔

mhluongo:
@sr-gi can you give this a look as well?

prestwich:
if you're ready, I can tag a release of bitcoin-spv that also improves `extractHash` to make length checks

Shadowfiend:
Yep, we can roll that in here @prestwich.

Does that change obviate the need for either of these changes? If so, I can always drop the code + keep the tests.

prestwich:
it'll fold the following check into `extractHash`
```
require(uint8(_output[8]) + 9 == _output.length, "Incorrect output script length");
```

prestwich:
okay I tagged and published `@summa-tx/bitcoin-spv-sol@3.1.0`

sr-gi:
Checking this atm

sr-gi:
lgtm wrt #658

Shadowfiend:
Kk will do the bitcoin-spv bump, drop our require, and see how the tests fare.

Thanks for the review @sr-gi as well!

Shadowfiend:
Hmmm @prestwich do we need a new version of relay-sol as well here?

```
    @summa-tx/relay-sol/contracts/OnDemandSPV.sol:257:9: Warning: Return value of low-level calls not used.
 
        address(c).call.gas(remoteGasAllowance)(
 
        ^ (Relevant source part starts here and spans across multiple lines).
 

 
@summa-tx/relay-sol/contracts/OnDemandSPV.sol:342:32: TypeError: Member "extractOutputScriptLen" not found or not visible after argument-dependent lookup in bytes memory.
 
            uint8 _len = uint8(_out.extractOutputScriptLen()[0]);
 
                               ^-------------------------^
```

This seems to be pulled in via `TestnetRelay.sol`, which we're importing to get some related artifacts we need for the mock relay for tests.

prestwich:
The warning about unused return values is normal. I'll take a look at the other thing

prestwich:
Bump the relay to v2.0.2

Shadowfiend:
2.0.2 doesn't appear to be published to npm <_<

prestwich:
uhhhh, check again and I definitely didn't just push it why would you think that :sweat_smile:

Shadowfiend:
Reddit reports that we may have done it.

Shadowfiend:
Okayyy we are finally good. @NicholasDotSol final merge is to you.

PR #669, opened 2020-06-03
Renaming of notifyUndercollateralizedLiquidation

elishadrion:
The function `notifyUndercollateralizedLiquidation` can initially create confusion as it's actually callable when the Deposit is **severely** undercollateralized.

My proposition is to rename it from `notifyUndercollateralizedLiquidation` to `notifySevereUndercollateralizedLiquidation`.

### DIFF
--- a/solidity/contracts/deposit/Deposit.sol
+++ b/solidity/contracts/deposit/Deposit.sol
@@ -425,10 +425,10 @@ contract Deposit is DepositFactoryAuthority {
         self.exitCourtesyCall();
     }
 
-    /// @notice     Notify the contract that the signers are undercollateralized.
+    /// @notice     Notify the contract that the signers are severely undercollateralized.
     /// @dev        Calls out to the system for oracle info.
-    function notifyUndercollateralizedLiquidation() public {
-        self.notifyUndercollateralizedLiquidation();
+    function notifySevereUndercollateralizedLiquidation() public {
+        self.notifySevereUndercollateralizedLiquidation();
     }
 
     /// @notice     Notifies the contract that the courtesy period has elapsed.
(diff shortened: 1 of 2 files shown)


Shadowfiend:
Going to give this a deeper look in a but, but at a glance this looks like a 💯 🎉 🚀 change on my just-made-up change awesomeness emoji scale.

elishadrion:
The changes are quite extensive. If you don't understand something, don't hesitate to hit me up! 💯 🎉 🚀

solidity/contracts/deposit/Deposit.sol
```diff
@@ -425,10 +425,10 @@ contract Deposit is DepositFactoryAuthority {
         self.exitCourtesyCall();
     }
 
-    /// @notice     Notify the contract that the signers are undercollateralized.
+    /// @notice     Notify the contract that the signers are severely undercollateralized.
     /// @dev        Calls out to the system for oracle info.
-    function notifyUndercollateralizedLiquidation() public {
-        self.notifyUndercollateralizedLiquidation();
+    function notifySevereUndercollateralizedLiquidation() public {
+        self.notifySevereUndercollateralizedLiquidation();
```
Shadowfiend, line 431: Actually quite tempted to take this opportunity to revamp all of our `notify*` naming, as it's not particularly consistent in how it describes what it's notifying… e.g. here, we're really doing `notifySeverlyUndercollateralized`, which *triggers* liquidation. Below, `notifyCourtesyTimeout` also triggers liquidation, but doesn't mention it in its name.

Let me see what timing overall we have, and then see if we merge as-is and follow up with broader renaming, or if we just do the broader renaming and close this PR. Thanks for calling attention here!

Shadowfiend:
I'm doing a mega-branch with a fairly large cleanup of `Deposit.sol` from an API + docs perspective, captured in #695. Closing this issue in its favor; expect to see this early next week, though you can have a sneak peek at https://github.com/keep-network/tbtc/tree/notification-station .

PR #672, opened 2020-06-04
Enforce receiveApproval caller is associated token

### LINKED ISSUE #670: Scripts do not validate caller when receiving approval
# Description
Both `FundingScript` and `RedemptionScript` allow arbitrary users to call `receiveApproval` with an arbitrary `_from` value. In the `FundingScript` case, this isn't so bad as the expected caller is an NFT so the user providing approval will likely use the approval in the same transaction. However, in the `RedemptionScript` case the expected token is an ERC20. As such, users who like to use "infinite approvals" may be vulnerable to theft.

# Exploit Scenario
Alice wants to use the redemption script to redeem her TBTC for BTC. She performs an infinite approval using `approveAndCall`. Later, when Alice has acquired more TBTC, Eve manually calls `receiveApproval` to redeem Alice's bitcoins for BTC.

# Suggested Remediation
Check that the caller is the token

NicholasDotSol:
Token holders can approve “infinitely”, which is to say they can approve a spender
to spend on their behalf for a balance greater than their current amount. In these
cases, that approval continues to apply when they gain a later balance. Some users
use this function to reduce the number of approval transactions they have to
perform.

If that is used in conjunction with the `RedemptionScript`, a third party can call
`RedemptionScript`'s `receiveApproval` directly to redeem another user's inifinitely
approved TBTC. To guard against this, this PR restricts `receiveApproval` caller to
only be the token handling approval.

Closes #670.

### DIFF
--- a/solidity/test/VendingMachineTest.js
+++ b/solidity/test/VendingMachineTest.js
@@ -531,6 +531,29 @@ describe("VendingMachine", async function() {
           "Bad _extraData signature. Call must be to tbtcToBtc.",
         )
       })
+
+      it("protects receiveApproval from external callers", async () => {
+        const tbtcToBtc = vendingMachine.abi.filter(
+          x => x.name == "tbtcToBtc",
+        )[0]
+        const calldata = web3.eth.abi.encodeFunctionCall(tbtcToBtc, [
+          testDeposit.address,
+          "0x1111111100000000",
+          redeemerOutputScript,
+          owner,
+        ])
+
+        await expectRevert(
+          redemptionScript.receiveApproval(
+            redemptionScript.address,
+            tdtId,
+            owner,
+            calldata,
+            {from: owner},
+          ),
+          "Only token contract can call receiveApproval",
+        )
+      })
     })
   })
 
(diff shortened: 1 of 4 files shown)


solidity/test/VendingMachineTest.js
```diff
@@ -531,6 +531,29 @@ describe("VendingMachine", async function() {
           "Bad _extraData signature. Call must be to tbtcToBtc.",
         )
       })
+
+      it("protects receiveApproval from external callers", async () => {
+        const tbtcToBtc = vendingMachine.abi.filter(
+          x => x.name == "tbtcToBtc",
+        )[0]
```
Shadowfiend, line 538: This can just be `find(...)` instead of `filter(...)[0]`.

solidity/test/VendingMachineTest.js
```diff
@@ -531,6 +531,29 @@ describe("VendingMachine", async function() {
...
+          redeemerOutputScript,
+          owner,
+        ])
+
+        await expectRevert(
+          redemptionScript.receiveApproval(
+            redemptionScript.address,
+            tdtId,
+            owner,
+            calldata,
```
Shadowfiend, line 551: Feels like this can just be an empty byte array without losing the test, right?

NicholasDotSol: yep

solidity/test/VendingMachineTest.js
NicholasDotSol: indeed

PR #679, opened 2020-06-16
Deposit Slog

### LINKED ISSUE #676: DepositLog concerns
# Description
Two concerns with `DepositLog`

1) Timestamps are logged even though you can look it up from the event block number
2) The funding txid isn't logged, but the redemption txid is

mhluongo:
Clean up deposit logging, removing unnecessary info and exposing the funding tx ID.

Closes #676.

### DIFF
--- a/solidity/contracts/DepositLog.sol
+++ b/solidity/contracts/DepositLog.sol
@@ -65,7 +65,10 @@ contract DepositLog {
     event FraudDuringSetup(address indexed _depositContractAddress);
 
     // This event is fired when we enter the ACTIVE state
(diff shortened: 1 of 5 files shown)


Shadowfiend:
See #676, I'm not convinced the timestamp thing should happen.

@NicholasDotSol can we split your changes to a separate PR that references 676? I think we can merge that change separately (though the commit will probably need adjusting).

mhluongo:
Closed as #676 is a [partial wontfix](https://github.com/keep-network/tbtc/issues/676#issuecomment-646012591)

PR #681, opened 2020-06-18
Log funding txid

NicholasDotSol:
refs: #676 

Unlike the redemption TX ID, the funding TX ID is not logged in events.


NicholasDotSol:
got some commits mixed up in here

Issue #652, opened 2020-05-26
Increasing redemption fee incorrectly updates the fee paid


samczsun:
# Description
When performing a withdrawal, the transaction fee (defined as the original amount of the UXTO minus the value transferred) is stored as a state variable.

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L110

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L123

This variable is checked when submitting a redemption proof, to make sure that the signers didn't change the fee amount (thereby causing a loss to the user).

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L313

A user is allowed to increase the redemption fee after a certain time has passed, in case the blockchain is congested. If this happens, the transaction fee is updated to be the output of `checkRelationshipToPrevious`.

https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L232-L233

However, the above function does not return the fee, but rather the value being sent to the user.
https://github.com/keep-network/tbtc/blob/100e5aae60aad209b0d3c6d495438d2b5b270fd3/solidity/contracts/deposit/DepositRedemption.sol#L259-L281

# Exploit Scenario
If all of the members of the keep are able to collude then they may be able to grief the depositor. If a miner is involved, they may be able to recover some of the BTC from the depositor. Of course, in both cases the members risk getting slashed for fraud so this exploit scenario isn't very probable.

# Suggested Remediation
Set the fee paid to the correct value.

sr-gi:
Came to report the same bug. I was wondering what the viability of this would be though, since I think it looks less severe that what it actually could be.

The signers would risk loss of funds for the minimum confirmation time, i.e. 6 blocks. However, this implies that there is a party that knows the funding UTXO and the original requested fee (either the redeemer or someone monitoring the calls to the contract) and that is also monitoring the Bitcoin blockchain in order to notice that the fraudulent transaction has a higher fee than expected.

Anyway, nice catch.

Shadowfiend:
[...]

PR #683, opened 2020-06-18
Fix fee bump

NicholasDotSol:
closes: #652 

`latestRedemptionFee` was assigned the value sent to the user instead of the updated redemption fee after the bump. This could lead to some possible attack vectors described in the linked issue.

### DIFF
--- a/solidity/test/DepositRedemptionTest.js
+++ b/solidity/test/DepositRedemptionTest.js
@@ -916,16 +916,17 @@ describe("DepositRedemption", async function() {
       )
     })
 
-    it("approves a new digest for signing, updates the state, and logs RedemptionRequested", async () => {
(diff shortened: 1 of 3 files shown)


solidity/test/DepositRedemptionTest.js
```diff
@@ -916,16 +916,17 @@ describe("DepositRedemption", async function() {
       )
     })
 
-    it("approves a new digest for signing, updates the state, and logs RedemptionRequested", async () => {
+    it("correctly increases fee, approves a new digest for signing, updates the state, and logs RedemptionRequested", async () => {
```
Shadowfiend, line 919: Feels like there's also a state test here that does a couple of increase fee bump tests and then confirms that you can redeem with the original fee, an intermediate fee, the final fee that was set, and cannot redeem with a fee higher than the one that was set. Thoughts?

PR #684, opened 2020-06-20
Update README.adoc

afmsavage:
fixed the TBTC channel name


Shadowfiend:
The linting jobs aren't relevant in this case, so I'm going to just push this on in.

PR #688, opened 2020-06-22
Introduce a supply cap manager

mhluongo:
Adds an additional `capManager` role to `TBTCSystem` to determine whether the current `VendingMachine` supply restriction can be increased to the next graduated cap.

Opening as a draft to get on the same page about the approach. I'm also not sure how we're testing role-based auth- advice welcome.

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -63,6 +63,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     uint256 constant pausedDuration = 10 days;
 
     VendingMachine public vendingMachine;
+    address public capManager;
 
     ISatWeiPriceFeed public priceFeed;
     IRelay public relay;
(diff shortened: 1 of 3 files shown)


mhluongo:
I played around with making the `VendingMachine` ownable, but this felt slightly less dirty..

In practice, we'd set this to a multisig at initialization so outside parties can decide on moving the supply schedule forward.

Shadowfiend:
Couple of things...

First, lol at the commit summary “add supply cap manager”. That’s an incidental part of what that commit does. 

Second, why make this a separate role instead of using the system owner? Including that info in the commit would also be useful.

mhluongo:
> First, lol at the commit summary “add supply cap manager”. That’s an incidental part of what that commit does.

So salty!

> Opening as a draft to get on the same page about the approach. I'm also not sure how we're testing role-based auth- advice welcome.

I'd planned to force-push wreck this branch, just wanted thoughts on the mechanism, how to better expose role-based testing in the suite, etc.

I'm having second (and third and fourth) thoughts on the idea at all, though.

mhluongo:
Meh, not feeling this. No reason to pre-program a cap curve and allow a committee to override it. We should address the feedback of the curve being too restrictive head-on, either by increasing the caps before deploy or shortening the schedule (my preference).

PR #689, opened 2020-06-23
What is time? Adjust the emergency pause and supply cap schedules

mhluongo:
Instead of allowing the emergency pause for 365 days, cut it down to 180. Instead of a supply cap for 4 months, move it to 3.

I've gone back and forth on these schedules and [other approaches](https://github.com/keep-network/tbtc/pull/688) to handle all the demand we're getting. Rather that increase the caps, I think removing the last 1000 BTC limit- which doesn't buy us much- is the right call.

### DIFF
--- a/solidity/contracts/system/VendingMachine.sol
+++ b/solidity/contracts/system/VendingMachine.sol
@@ -35,7 +35,7 @@ contract VendingMachine is TBTCSystemAuthority{
 
     /// @notice Get the maximum TBTC token supply based on the age of the contract
     ///         deployment. The supply cap starts at 2 BTC for the first day, 100 for
-    ///         the first 30 days, 250 for the next 30, 500 for the kast 30... then
+    ///         the first 30 days, 250 for the next 30, 500 for the last 30... then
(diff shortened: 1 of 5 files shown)


solidity/contracts/system/VendingMachine.sol
```diff
@@ -28,16 +28,15 @@ contract VendingMachine is TBTCSystemAuthority{
...
     function getMintedSupply() public view returns (uint256) {
         return tbtcToken.totalSupply();
     }
 
     /// @notice Get the maximum TBTC token supply based on the age of the contract
     ///         deployment. The supply cap starts at 2 BTC for the first day, 100 for
-    ///         the first 30 days, 250 for the next 30, 500 for the next 30, 1000 for
-    ///         the last 30... then finally removes the restriction, returning 21M BTC
-    ///         as a sanity check.
+    ///         the first 30 days, 250 for the next 30, 500 for the kast 30... then
```
Shadowfiend, line 38: It's the kast 30 that'll get ya.

mhluongo: Hyuck

PR #691, opened 2020-06-25
Clean up the README

mhluongo:
1. Removed outdated references to LaTeX from README
2. Switched the docs badge for green, because it's prettier 💄
3. Fixed a link


Shadowfiend:
I'm good to go here… The Solidity link will be outdated, but that needs fixing in the next week or two anyway IMO.

Issue #690, opened 2020-06-25
Move Solidity doc generation to Github Actions


mhluongo:
I'm pretty sure it's not running at all in CircleCI. Migrating is also a good opportunity to remove the outdated LaTeX / TikZ cruft.

PR #692, opened 2020-06-25
Disable the Solidity docs build

mhluongo:
~~It was missed in the migration to Github Actions.~~

Disabled the build. We'll reintroduce it in #690


Shadowfiend:
The Solidity build generates asciidoc, so it would require reintroducing asciidoc builds as well. I would rather remove it for now, honestly.

PR #693, opened 2020-06-25
Fee refactor draft

NicholasDotSol:


solidity/contracts/deposit/DepositRedemption.sol
```diff
@@ -53,41 +53,21 @@ library DepositRedemption {
...
-            // and we should always burn a full TBTC to redeem the deposit.
+        if(tbtcOwedToDeposit > 0){
             if(tdtHolder == vendingMachineAddress){
-                _d.tbtcToken.burnFrom(msg.sender, tbtcLot);
-            }
-            // if signer fee is not escrowed, escrow and it here and send the rest to TDT holder
-            else if(_d.tbtcToken.balanceOf(address(this)) < signerFee){
-                _d.tbtcToken.transferFrom(msg.sender, address(this), signerFee);
-                _d.tbtcToken.transferFrom(msg.sender, tdtHolder, tbtcLot.sub(signerFee));
+                _d.tbtcToken.burnFrom(msg.sender, tbtcOwedToDeposit);
```
Shadowfiend, line 61: I'm wondering if we need to revert if TDT holder is vending machine and this amount is not equal to lot size. Ostensibly that shouldn't happen, but it bears some consideration what we should do if it *does* happen.

solidity/contracts/deposit/DepositRedemption.sol
```diff
@@ -53,41 +53,21 @@ library DepositRedemption {
...
             }
-            // tansfer a full TBTC to TDT holder if signerFee is escrowed
-            else{
-                _d.tbtcToken.transferFrom(msg.sender, tdtHolder, tbtcLot);
-            }
-            return;
+            _d.tbtcToken.transferFrom(msg.sender, address(this), tbtcOwedToDeposit);
```
Shadowfiend, line 63: This needs to be an else, not a fallthrough,

solidity/contracts/deposit/DepositRedemption.sol
```diff
@@ -53,41 +53,21 @@ library DepositRedemption {
...
+        }
+        if(tbtcOwedToTdtHolder > 0){
```
Shadowfiend, line 65: Do we… Need this guard? I don't see anywhere that ERC20s revert on 0 transfers, I think that's only ETH `send`s?

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -489,33 +523,32 @@ library DepositUtils {
...
     ) internal view returns (uint256) {
-        bool redeemerOwesNoFee;
+        bool redeemerOwesNoFeeRebate;
         if (_preTerm) {
-            redeemerOwesNoFee = _redeemerHoldsFrt;
-        } else {
-            redeemerOwesNoFee = _frtExists || _redeemerIsThirdParty;
+            redeemerOwesNoFeeRebate = _redeemerHoldsFrt;
+        }else {
+            redeemerOwesNoFeeRebate = true;
```
Shadowfiend, line 546: You can drop this branch by setting `redeemerOwesNoFeeRebate = true` at initialization. Then we replace it with `_redeemerHoldsFrt` when `_preTerm` is true.

In fact, assuming this logic is correct, I think this can now be `bool redeemerOwesNoFeeRebate = atTerm || _redeemerHoldsFrt;` and the if statement can go away altogether.

Shadowfiend: Second part of this that doesn't hit right is that the redeemer should never really owe the fee rebate, it's just that the escrow requires it. I'm wondering if `computeSignerFee` becomes `computeRedemptionFeeEscrow`. Then:

```solidity
function computeRedemptionFeeEscrow(
    DepositUtils.Deposit storage _d,
    bool _preTerm,
    bool _redeemerHoldsFrt
) internal view returns (uint256) {
    bool escrowRequiresFeeRebate = _preTerm && !_redeemerHoldsFrt;

    if (escrowRequiresFeeRebate) {
        return signerFee(_d).mul(2);
    } else {
        return signerFee(_d);
    }
}
```

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -447,24 +447,58 @@ library DepositUtils {
     ///                     is this deposit's TDT holder.
     /// @param _redeemer    The assumed owner of the deposit's TDT.
     /// @return             The amount in TBTC needed to redeem the deposit.
-    function getOwnerRedemptionTbtcRequirement(DepositUtils.Deposit storage _d, address _redeemer) internal view returns(uint256) {
-        // The TDT owner only ever owes fee adjustments.
-        return computeRedemptionFeeAdjustment(_d, _redeemer);
+    function getOwnerRedemptionTbtcRequirement(
+        DepositUtils.Deposit storage _d,
+        address _redeemer
+    ) internal view returns(uint256 owedToDeposit, uint256 owedToRedeemer) {
```
Shadowfiend, line 453: I'm not sure why `getOwnerRedemptionTbtcRequirement` would return anything other than the amount owed to the deposit.

solidity/contracts/deposit/DepositRedemption.sol
NicholasDotSol: I don't think you are, nice

solidity/contracts/deposit/DepositRedemption.sol
NicholasDotSol: technically no, but I'd rather check an `if()` statement than do a full transfer + emit an event with zero value.

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: the simplification makes sense, and the rename also makes sense I think

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -447,24 +447,58 @@ library DepositUtils {
...
+            if(!redeemerHoldsTdt){
+                owedToTdtHolder = lotSizeTbtc(_d).sub(signerFee(_d));
+            }
+            return(baseRedemptionRequirement, 0, owedToTdtHolder);
+        }
+        if(!redeemerHoldsTdt){
+            if(currentBalance >= signerFee(_d)){
+                owedToTdtHolder = lotSizeTbtc(_d);
+            }else{
+                owedToTdtHolder = lotSizeTbtc(_d).sub(currentBalance);
```
Shadowfiend, line 493: Is this less the current balance, or is it less the difference between the signer fee and the current balance? Intuitively, I think this should be `lotSizeTbtc(_d).sub(signerFee(_d).sub(currentBalance))`; could be wrong.

Shadowfiend: Ah no, I see. Got turned around.

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: Ah yes, forgot to remove this

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: I think it should be less the current balance, as long as the current balance is less than the signer fee

NicholasDotSol: ah GH didn't refresh to comment, sorry

solidity/contracts/deposit/DepositUtils.sol
```diff
@@ -447,24 +447,58 @@ library DepositUtils {
...
+            }
+        }
+        if(currentBalance > baseRedemptionRequirement){
+            owedToRedeemer = currentBalance.sub(baseRedemptionRequirement);
+            return(0, owedToRedeemer, owedToTdtHolder);
+        }
+        owedToDeposit = baseRedemptionRequirement.sub(currentBalance);
+        return(owedToDeposit, 0, owedToTdtHolder);
```
Shadowfiend, line 501: Ok, so I think there's value to stacking these but I think the approach is a bit jumbled right now. I suggest we compute everything with respect to the initial escrow amount.

In particular, we first look up how much *must be in escrow*.

Then, we compute:

- the amount owed to the deposit as *escrow requirement - current balance*, clamped to 0.
- the amount owed to the redeemer as *current balance - escrow requirement*, clamped to 0.
- the amount owed to the TDT holder as *escrow requirement - (signer fee - fee rebate)*, with fee rebate only being included if `preTerm && ! _redeemerHoldsFrt`.
- the amount owed to the FRT holder as *fee rebate*, with a max of current balance. We don't currently return this, but that seems like an oversight?

I would introduce a helper here:

```
function subtractWithFloor(uint256 left, uint256 right, uint256 min) returns (uint256) {
    if (left > right) {
        return left.sub(right);
    } else {
        return min;
    }
}
```

To make this read more clearly.

Thoughts? This removes almost all conditionals from the body of the function, except for the one around fee rebate.

NicholasDotSol: I'm not sure I totally follow

NicholasDotSol:
```
        uint256 mainCharge = computeBaseRedemptionCharge(_d, redeemerHoldsTdt);
        uint256 baseRedemptionRequirement = mainCharge.add(computeRedemptionFeeAdjustment(_d, _redeemer));
        uint256 currentBalance = _d.tbtcToken.balanceOf(address(this));

        owedToDeposit = subtractWithFloor(baseRedemptionRequirement, currentBalance, 0);
        owedToRedeemer = subtractWithFloor(currentBalance, baseRedemptionRequirement, 0);
        if(!redeemerHoldsTdt){
            owedToTdtHolder = lotSize.sub(subtractWithFloor(signerFee, currentBalance, 0));
        }
        return(owedToDeposit, owedToRedeemer, owedToTdtHolder);
```
is a simplified state using the helper that works as well

NicholasDotSol:
> the amount owed to the FRT holder as fee rebate, with a max of current balance. We don't currently return this, but that seems like an oversight?

Not fully clear on what the FRT issue is. All we want to do is escrow what is needed. The fee is distributed (if it is escrowed) later on in the redemption flow. What is missing from the FRT handling logic?

solidity/contracts/deposit/DepositUtils.sol
NicholasDotSol: > the amount owed to the FRT holder as fee rebate, with a max of current balance. We don't currently return this, but that seems like an oversight?

Not fully clear on what the FRT issue is. All we want to do is escrow what is needed. The fee is distributed (if it is escrowed) later on in the redemption flow. What is missing from the FRT handling logic?

Shadowfiend: Do we pay all of these pieces out at the beginning of the redemption flow? I'm trying to see where the logical separation is between the owed amounts we return here and the owed amount for the FRT. If we can't compute that because it's managed later in the flow, we have an opportunity to note that in the function docs.

Shadowfiend: Also---the simplified logic with the helper looks much more readable to me, other than a few nits I have with a couple of the variable names---wdyt?

NicholasDotSol: during `provideRedemptionProof`, we'll `distributeSignerFee`, then `distributeFeeRebate`. These are callen in succession and they read from the Deposit balance. `distributeSignerFee` should always do something since we should always have `signerFee` escrowed at this time, but `distributeFeeRebate` is not guaranteed to do something. These both depend on the return value from `getRedemptionTbtcRequirement`. The owed amounts should include the FRT. `computeRedemptionFeeEscrow` returns `signerFee` amount as base return value, or  `2*signerFee` if fee rebate is owed. 
`performRedemptionTBTCTransfers` enforces that these values are escrowed in the Deposit, and `provideRedemptionProof` will finally handle sending the `signerFee` and `FeeRebate`

Shadowfiend: It's worth explicitly writing these invariants down somewhere, probably in the docs for `performRedemptionTBTCTransfers` (something along the lines of “once these transfers complete, the deposit balance should be sufficient to pay out signer fees plus any fee rebate owed”).

NicholasDotSol:
closing in favor of #694

PR #696, opened 2020-07-07
Give Ropsten development constants

Shadowfiend:
Currently, TBTCConstants requires proof of 6 blocks; unfortunately, for
BTC testnet this can be a challenge with the difficulty continuity that the
contracts impose. Shifting Ropsten to development constants makes this
feasible again.


Shadowfiend:
@sthompson22 let's do a test deploy of this on Ropsten and see if it makes the dApp complete a little more straightforwardly.

PR #698, opened 2020-07-07
Mitigate supply cap exhaustion attacks

mhluongo:
The additional deposit restriction opened up the system to a variety of subtle cap exhaustion attacks. This PR removes the deposit restriction, so racing to the cap to prevent additional deposit creation can no longer be used to game distribution.

Note that depositing "past the cap" has always been possible- if at system instantiation I opened 100 1-BTC deposits but never minted, there could be 200 BTC in the system in the first 30 days. This PR just removes the edge case where 100 TBTC have been minted, allowing deposits to be opened and closed independently.


mhluongo:
I don't love that this kills tests without replacement, but the test it removes was in an odd spot. Suggestions on how best to increase coverage here are welcome.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -153,8 +149,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     /// @return True if new deposits should be allowed, both by the emergency pause button
     ///         and respected the max supply schedule.
```
NicholasDotSol, line 150: `@return `doc is out of date now

solidity/test/DepositFundingTest.js
```diff
@@ -66,194 +60,6 @@ describe("DepositFunding", async function() {
     await restoreSnapshot()
   })
 
-  describe("createNewDeposit", async () => {
```
NicholasDotSol, line 69: Can you clarify why this whole test killed instead of just the supply-cap specific ones

NicholasDotSol:
@mhluongo, so the supply cap does not affect BTC -> TDT minting, just TDT -> TBTC?

solidity/contracts/system/TBTCSystem.sol
mhluongo: Taking care of that + linting

solidity/test/DepositFundingTest.js
mhluongo: :facepalm:

mhluongo:
> so the supply cap does not affect BTC -> TDT minting, just TDT -> TBTC?

Correct, that's how the cap has always worked- you could create tons of deposits before the minting cap was hit, since deposits aren't required to be used to mint.

mhluongo:
Closing in favor of #700

PR #700, opened 2020-07-08
Mitigate supply cap exhaustion attacks

mhluongo:
The additional deposit restriction opened up the system to a variety of subtle cap exhaustion attacks. This PR removes the deposit restriction, so racing to the cap to prevent additional deposit creation can no longer be used to game distribution.

Note that depositing "past the cap" has always been possible- if at system instantiation I opened 100 1-BTC deposits but never minted, there could be 200 BTC in the system in the first 30 days. This PR just removes the edge case where 100 TBTC have been minted, allowing deposits to be opened and closed independently.

Original discussion on #698.

### DIFF
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -20,7 +20,7 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "./KeepFactorySelection.sol";
 
-/// @title  TBTC System.
+/// @title TBTC System.
 /// @notice This contract acts as a central point for access control,
 ///         value governance, and price feed.
 /// @dev    Governable values should only affect new deposit creation.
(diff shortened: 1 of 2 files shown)


solidity/contracts/system/TBTCSystem.sol
```diff
@@ -143,18 +140,15 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
-        vendingMachine = _vendingMachine;
         setTbtcDepositToken(_tbtcDepositToken);
         initializedTimestamp = block.timestamp;
         allowNewDeposits = true;
     }
 
     /// @notice Returns whether new deposits should be allowed.
-    /// @return True if new deposits should be allowed, both by the emergency pause button
-    ///         and respected the max supply schedule.
+    /// @return True if new deposits should be allowed by the emergency pause button
```
Shadowfiend, line 149: “Should“ is pretty funny phrasing considering that if this returns false, new deposits *are* not allowed :grin: But I think we were already using that anyway.

PR #705, opened 2020-07-15
Bumper Cars: Introduce a max initial fee and minimum final UTXO value during redemption

### LINKED ISSUE #671: A redemption with zero value is allowed by tbtc but rejected by the mempool
# Description
Bitcoin rejects transactions that send "dust", and an output with zero sats very comfortably meets the definition of dust. Such a transaction isn't rejected by tbtc, creating a discrepancy which can be exploited to exchange BTC for signer ETH.

# Exploit Scenario
Eve creates a deposit and requests a redemption with 0 value. The signers are unable to get this transaction included in a block without going directly to a miner. Assuming they fail to do so, Eve gets 150% of her deposit back in the form of ETH.

# Suggested Remediation
A simple solution would be to implement an upper cap on fees. A better one might be to reimplement the dust calculations, although the dust threshold is set on a per-node basis. As both solutions are slightly brittle, a third solution might be to redesign the system such that invalid transactions that can't be accepted into the mempool can be used to penalize the depositor in some way.

Shadowfiend:
To avoid approving signatures for transactions that are very unlikely
to go through on the Bitcoin side, this PR imposes a maximum on the
fee permitted for the Bitcoin redemption transaction. This fee maximum
is imposed as a minimum on the remaining UTXO value, defined as a
constant in `TBTCConstants.MINIMUM_UTXO_VALUE`.

Additionally, to extend the time that Bitcoin fees would have to remain high
with respect to the deposit's value in outlier cases, the initial fee specified at
redemption request time is limited to half of the deposit's value.

Last but not least, if a fee bump would push the UTXO value below the minimum,
the bump is permitted, but the resulting fee and UTXO values are clamped to
ensure the minimum is respected. This allows one more fee bump than might
otherwise be permitted, and combined with the previous change ensures that
a redemption fee can be bumped at least once.

-------

See the linked issue for additional details on potential attacks and how this
particular solution was zeroed in on. In particular:

> if the minimum UTXO value is 2000 sats, and the minimum available lot size is 50000 sats, the occurrence would require a fee event that would, at minimum, cause loss of 96% of the UTXO value in fees. An honest redeemer can be expected to start their redemption at the lowest reasonable fee value, and to avoid high-fee scenarios if possible. A dishonest redeemer could push for high fees to grief, but assuming the minimum lot size of 50k sats is available and the minimum UTXO value is 2k sats, my understanding is that would require fees of ~480 sats/vbyte, sustained for ~5 hours (due to the limited initial fee + allowing an additional fee bump, and assuming an hour to get the requisite confirmations). If this feels insufficient, we can raise the minimum allowed lot size to also raise the max fee.
>
> Worth noting: there is an effective (non-governance) minimum lot size of over 100k sats at the moment due to the ETH/BTC price and #634 .

Closes #671.

### DIFF
--- a/solidity/contracts/deposit/DepositRedemption.sol
+++ b/solidity/contracts/deposit/DepositRedemption.sol
@@ -234,7 +234,16 @@ library DepositRedemption {
         require(block.timestamp >= _d.withdrawalRequestTime.add(TBTCConstants.getIncreaseFeeTimer()), "Fee increase not yet permitted");
 
         uint256 _newOutputValue = checkRelationshipToPrevious(_d, _previousOutputValueBytes, _newOutputValueBytes);
+
+        // If the fee bump shrinks the UTXO value below the minimum allowed
+        // value, clamp it to that minimum. Further fee bumps will be disallowed
+        // by checkRelationshipToPrevious.
+        if (_newOutputValue < TBTCConstants.getMinimumUtxoValue()) {
+            _newOutputValue = TBTCConstants.getMinimumUtxoValue();
+        }
+
         _d.latestRedemptionFee = _d.utxoValue().sub(_newOutputValue);
+
         // Calculate the next sighash
         bytes32 _sighash = CheckBitcoinSigs.wpkhSpendSighash(
             _d.utxoOutpoint,
@@ -256,7 +265,7 @@ library DepositRedemption {
             _sighash,
             _d.utxoValue(),
             _d.redeemerOutputScript,
-            _d.utxoValue().sub(_newOutputValue),
+            _d.latestRedemptionFee,
             _d.utxoOutpoint);
     }
 
--- a/solidity/contracts/deposit/DepositRedemption.sol
+++ b/solidity/contracts/deposit/DepositRedemption.sol
@@ -108,7 +108,12 @@ library DepositRedemption {
         // Convert the 8-byte LE ints to uint256
         uint256 _outputValue = abi.encodePacked(_outputValueBytes).reverseEndianness().bytesToUint();
         uint256 _requestedFee = _d.utxoValue().sub(_outputValue);
+
         require(_requestedFee >= TBTCConstants.getMinimumRedemptionFee(), "Fee is too low");
+        require(
+            _requestedFee < _d.utxoValue() / 2,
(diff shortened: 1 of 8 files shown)


solidity/contracts/deposit/DepositRedemption.sol
```diff
@@ -108,7 +108,12 @@ library DepositRedemption {
         // Convert the 8-byte LE ints to uint256
         uint256 _outputValue = abi.encodePacked(_outputValueBytes).reverseEndianness().bytesToUint();
         uint256 _requestedFee = _d.utxoValue().sub(_outputValue);
+
         require(_requestedFee >= TBTCConstants.getMinimumRedemptionFee(), "Fee is too low");
+        require(
+            _requestedFee < _d.utxoValue() / 2,
```
NicholasDotSol, line 114: we could use SafeMath here, but its not necessary

solidity/test/DepositRedemptionTest.js
```diff
@@ -996,47 +1020,68 @@ describe("DepositRedemption", async function() {
...
-        await increaseTime(feeIncreaseTimer)
-
         const updatedFee = await testDeposit.getLatestRedemptionFee.call()
-        expect(updatedFee).to.eq.BN(
-          new BN(increment).mul(new BN(2).add(new BN(i))),
-        )
+        expect(updatedFee).to.eq.BN(feeIncrement.muln(2 + i))
       }
+
+      // The last fee increase pushes the deposit below the minimum UTXO value,
```
NicholasDotSol, line 1072: :+1:

solidity/contracts/deposit/DepositRedemption.sol
Shadowfiend: Talked about this a bit further. Strictly speaking SafeMath can only give us a clearer error message when dividing by 0; otherwise it's identical to a regular division. Given that we have the divisor set to 2, and it would need to become dynamic and be subject to a 0 value for divide-by-0 to be an issue, we went back and forth on it.

I pondered this a bit long and I think in the interest of time here, I'm going to suggest we punt and merge this as-is @NicholasDotSol . If you're comfortable with that, merging is in your court :)

PR #709, opened 2020-07-21
Circle: Create sortition pool

sthompson22:
### Intro 

After a fresh migration of tbtc contracts we need to create a sortition pool for the `TBTCSystem` application.  This is a one time endeavor, and doesn't need to be done again until we do another tbtc contract migration.

We currently handle sortition pool creation via the automated init process on keep-ecdsa client deployment.  Specifically, it's handled [here](https://github.com/keep-network/keep-ecdsa/blob/master/infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js#L210).  

While this works the configuration is a bit out of place.  It's true that keep-ecdsa requires there to be a sortition pool present for the given `TBTCSystem` to run, but a sortition pool not a product of keep-ecdsa configuration.  Rather, it's a product of environment initialization after a fresh migration.

Application deployment InitContainers should operate within the context of provisioning the application.

### What We Did

To try and correct the ^ scenario we're moving sortition pool creation to a CircleCI job that will be part of our contract deployment workflow.  Once tbtc migrations are done, the included script will create a sortition pool for the  new `TBTCSystem`.

### Testing

Ran the script against a local deployment prepared by the [local-setup](https://github.com/keep-network/local-setup) repo.  I modified the `lcl-initialization` script that's part of the `install-tbtc` workflow to exclude the creation of a sortition pool, which typically happens automatically via the `local-setup` repo scripts.

Script Output
```
sthompson22 ~/Projects/keep/local-setup/tbtc/solidity (master) $ node create-sortition-pool.js 
creating sortition pool for BondedECDSAKeepFactory: [0x4849e972bd3CC5b9cFd45A9EcA0d2683dC0F49E2]
sortition pool creation tx: [0x73faa5ad797c04dc61b7fe4f6cbfaae13b5018947477b172281e18908ed900cb]
sortition pool address: 0xef6F456daA1b122665d9E9B20eE82c19E17092eC
```

NOTE:  I won't be able to exercise the CircleCI bits until the next Ropsten release, I fully expect to need some tweaks in a follow  up PR.

### Dependencies

Needs to be merged with https://github.com/keep-network/keep-ecdsa/pull/507


solidity/package.json
```diff
@@ -41,31 +41,32 @@
...
     "mocha": "^7.0.1",
-    "@types/mocha": "^7.0.2",
     "prettier": "^1.19.1",
     "prettier-plugin-solidity": "^1.0.0-alpha.40",
+    "solc": "0.5.17",
     "solium-config-keep": "git+https://github.com/keep-network/solium-config-keep.git#0.1.2",
     "truffle": "^5.1.13",
-    "solc": "0.5.17",
-    "@asciidoctor/core": "2.2.0"
+    "web3": "^1.2.11"
```
sthompson22, line 69: Worth calling out if there's a  way to handle this script execution without importing biggun' web3, I'm all for it...truffle maybe?

.circleci/config.yml
```diff
@@ -200,6 +200,18 @@ jobs:
...
+    executor: docker-node
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Create sortition pool
+          working_directory: ~/project/solidity
+          command: |
+            cp ./scripts/create_sortition_pool.js ./
+            node create-sortition-pool.js
```
nkuba, line 213: Do we have to do `cp`?
```suggestion
          working_directory: ~/project/solidity/scripts
          command: node create-sortition-pool.js
```

solidity/package.json
nkuba: This would require some updates to the script to be handled by truffle. Executing with truffle may simplify the script, but using contracts from keep-ecdsa will need some magic anyways.
Truffle injects `web3` instance so we could still use `web3` in the script.

solidity/scripts/circleci-create-sortition-pool.js
```diff
@@ -0,0 +1,94 @@
...
+BondedECDSAKeepFactory.options.handleRevert = true
+
+// We need the TBTCSystem address to pass to the create pool function.
+const TBTCSystemContractAddress =
+  TBTCSystemJson.networks[ethereumNetworkId].address
+
+async function createTBTCSystemSortitionPool() {
+  try {
+    console.log(
+      `creating sortition pool for BondedECDSAKeepFactory: [${BondedECDSAKeepFactoryJsonAddress}]`,
```
nkuba, line 67:
```suggestion
      `creating sortition pool for BondedECDSAKeepFactory: [${BondedECDSAKeepFactoryAddress}] and TBTCSystem: [${TBTCSystemContractAddress}]`,
```

solidity/scripts/circleci-create-sortition-pool.js
```diff
@@ -0,0 +1,94 @@
...
+const TBTCSystemJson = require("./build/contracts/TBTCSystem.json")
+
+// Boilerplate setup for access contract functions.
+const BondedECDSAKeepFactoryJsonAbi = BondedECDSAKeepFactoryJson.abi
+const BondedECDSAKeepFactoryJsonAddress =
+  BondedECDSAKeepFactoryJsonJson.networks[ethereumNetworkId].address
+const BondedECDSAKeepFactory = new web3.eth.Contract(
+  BondedECDSAKeepFactoryJsonAbi,
+  BondedECDSAKeepFactoryJsonAddress,
+)
```
nkuba, line 57: Maybe we could simplify naming and drop `Json` from some of the variables names?
```suggestion
const BondedECDSAKeepFactoryAbi = BondedECDSAKeepFactoryJson.abi
const BondedECDSAKeepFactoryAddress =
  BondedECDSAKeepFactoryJson.networks[ethereumNetworkId].address
const BondedECDSAKeepFactory = new web3.eth.Contract(
  BondedECDSAKeepFactoryAbi,
  BondedECDSAKeepFactoryAddress,
)
```

.circleci/config.yml
sthompson22: I ran down this road, but assumed there would be a problem with sourcing  the script directly from the `/scripts` dir, since the script is currently setup in a way to lean on the installed packages in the `/solidity` dir.  

I could update the paths for each requires, or provide a separate package file for this script.  All of that seemed less straight forward than just moving the file up a level to lean on the existing packages.

solidity/scripts/circleci-create-sortition-pool.js
sthompson22: This looks like some copy/paste fuckery that went awry, thanks for this.

sthompson22:
Closing this in favor of an alternative approach [here](https://github.com/keep-network/tbtc/pull/715)

PR #714, opened 2020-07-24
Refresh the minimum bondable value

pdyraga:
Refs https://github.com/keep-network/keep-ecdsa/issues/506

~~Depends on https://github.com/keep-network/keep-ecdsa/pull/508~~
~~Depends on https://github.com/keep-network/sortition-pools/pull/84~~

Deep inside ECDSA keep factory implementation, `BondedSortitionPool` was updating the minimum bondable value for each new group selection. The idea was to automatically remove from the pool operators that are no longer able to satisfy application needs about the minimum available bondable value and prevent operators with too low value griefing signer selection.

We now have two minimum bond values: one defining the minimum unbonded value the operator needs to have so that it can join and stay in the pool, and another defining the required bond value for the current selection. If the given operator has enough unbonded value to stay in the pool but it does not have enough unbonded value for the current group selection, it is skipped.

In ECDSA keep factory, sortition pool is created with a minimum bond of 20 ETH to avoid small operators joining and griefing future selections before the minimum bond is set to the right value by the application. Anyone can create a sortition pool for an application with the default minimum bond value but the application can change this value later, at any point. 20 ETH per operator should be more than enough to cover 1 BTC deposit in 3-operators group.

The default is fine for most cases <now> but the application is expected to refresh the minimum bondable value, depending on the current situation and needs.

For tBTC, the minimum bond value proposed here is be the minimum lot size expressed in wei given the current BTC price with 150% collateralization. For such a value, operator should be able to participate in at least 3 keeps.

Here we add a function allowing to refresh the minimum bondable value given the current BTC price. The function can be called by anyone at any moment. It is recommended to call this function on tBTC initialization and after the minimum lot size update.

### DIFF
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -91,20 +91,28 @@ library KeepFactorySelection {
 
     /// @notice Sets the minimum bondable value required from the operator to
     /// join the sortition pool for tBTC.
-    /// @param _minimumBondableValue The minimum unbonded value allowing
-    /// an operator to join and stay in the sortition pool for the application.
+    /// @param _minimumBondableValue The minimum bond value the application
+    /// requires from a single keep.
+    /// @param _groupSize Number of signers in the keep.
+    /// @param _honestThreshold Minimum number of honest keep signers.
     function setMinimumBondableValue(
         Storage storage _self,
-        uint256 _minimumBondableValue
+        uint256 _minimumBondableValue,
+        uint256 _groupSize,
+        uint256 _honestThreshold
     ) public {
         if (address(_self.keepStakeFactory) != address(0)) {
             _self.keepStakeFactory.setMinimumBondableValue(
-                _minimumBondableValue
+                _minimumBondableValue,
+                _groupSize,
+                _honestThreshold
             );
         }
         if (address(_self.ethStakeFactory) != address(0)) {
             _self.ethStakeFactory.setMinimumBondableValue(
-                _minimumBondableValue
+                _minimumBondableValue,
+                _groupSize,
+                _honestThreshold
             );
         }
     }
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -49,7 +49,7 @@ library KeepFactorySelection {
     function initialize(
         Storage storage _self,
         IBondedECDSAKeepFactory _defaultFactory
-    ) internal {
+    ) public {
         require(
             address(_self.keepStakeFactory) == address(0),
             "Already initialized"
@@ -87,6 +87,26 @@ library KeepFactorySelection {
         refreshFactory(_self);
 
         return factory;
+    }
+
+    /// @notice Sets the minimum bondable value required from the operator to
+    /// join the sortition pool for tBTC.
+    /// @param _minimumBondableValue The minimum unbonded value allowing
+    /// an operator to join and stay in the sortition pool for the application.
+    function setMinimumBondableValue(
+        Storage storage _self,
+        uint256 _minimumBondableValue
+    ) public {
+        if (address(_self.keepStakeFactory) != address(0)) {
+            _self.keepStakeFactory.setMinimumBondableValue(
+                _minimumBondableValue
+            );
+        }
+        if (address(_self.ethStakeFactory) != address(0)) {
+            _self.ethStakeFactory.setMinimumBondableValue(
+                _minimumBondableValue
+            );
+        }
     }
 
     /// @notice Refreshes the keep factory choice. If either ETH-stake factory
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -566,7 +566,9 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     /// after minimum lot size update.
     function refreshMinimumBondableValue() public {
         keepFactorySelection.setMinimumBondableValue(
-            calculateBondRequirementWei(getMinimumLotSize())
+            calculateBondRequirementWei(getMinimumLotSize()),
+            keepSize,
+            keepThreshold
         );
     }
 
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -162,7 +162,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
 
     /// @notice Return the largest lot size currently enabled for deposits.
     /// @return The largest lot size, in satoshis.
-    function getMaxLotSize() public view returns (uint256) {
+    function getMaximumLotSize() public view returns (uint256) {
         return lotSizesSatoshis[lotSizesSatoshis.length - 1];
     }
 
--- a/solidity/contracts/system/TBTCSystem.sol
+++ b/solidity/contracts/system/TBTCSystem.sol
@@ -506,13 +506,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
             "Caller must be a Deposit contract"
         );
 
-        uint256 price = priceFeed.getPrice();
-        if (price == 0 || price > 10 ** 18) {
-            // This is if a sat is worth 0 wei, or is worth >1 ether. Revert at
-            // once.
-            revert("System returned a bad price");
-        }
-        return price;
+        return _fetchBitcoinPrice();
     }
 
     // Difficulty Oracle
@@ -566,6 +560,24 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         );
     }
 
+    /// @notice Refreshes the minimum bondable value required from the operator
+    /// to join the sortition pool for tBTC. The minimum bondable value is
+    /// equal to the current minimum lot size collateralized 150% multiplied by
+    /// the current BTC price.
+    /// @dev It is recommended to call this function on tBTC initialization and
+    /// after minimum lot size update.
+    function refreshMinimumBondableValue() external {
+        uint256 minimimLotSizeSatoshis = lotSizesSatoshis[0];
+        uint256 bondRequirementSatoshis = minimimLotSizeSatoshis.mul(
+            initialCollateralizedPercent
+        ).div(100);
+        uint256 bondRequirementWei = _fetchBitcoinPrice().mul(
+            bondRequirementSatoshis
+        );
+
+        keepFactorySelection.setMinimumBondableValue(bondRequirementWei);
+    }
+
     /// @notice Returns the time delay used for governance actions except for
     ///         price feed additions.
     function getGovernanceTimeDelay() public pure returns (uint256) {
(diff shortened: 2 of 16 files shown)


Shadowfiend:
Is this something we would also expect `TBTCSystem` to call when updating the available lot sizes?

pdyraga:
> Is this something we would also expect `TBTCSystem` to call when updating the available lot sizes?

Yes, I think it's a good idea. I have not originally added it because I thought not every lot sizes update needs to modify the minimum lot. On the other hand, if we do the refresh and the minimum lot size is updated, we'll save gas on not submitting another transaction. Also, this could be a good opportunity to update the minimum unbonded value given the most recent ETBBTC price. It's done - please see the last two commits.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -89,6 +89,26 @@ library KeepFactorySelection {
         return factory;
     }
 
+    /// @notice Sets the minimum bondable value required from the operator to
+    /// join the sortition pool for tBTC.
+    /// @param _minimumBondableValue The minimum unbonded value allowing
+    /// an operator to join and stay in the sortition pool for the application.
+    function setMinimumBondableValue(
+        Storage storage _self,
+        uint256 _minimumBondableValue
```
Shadowfiend, line 98: Should we compute this rather than pass it in? i.e., rather than making it a setter, make it literally a `refresh` that takes the lowest lot size and multiplies it by the current price, or some factor thereof?

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -566,6 +562,33 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+        ).div(100);
+        uint256 bondRequirementWei = _fetchBitcoinPrice().mul(
+            bondRequirementSatoshis
+        );
+        keepFactorySelection.setMinimumBondableValue(bondRequirementWei);
+    }
+
+    /// @notice Returns the minimum lot size in satoshi.
+    function getMinimumLotSizeSatoshi() public view returns (uint256) {
+        uint256 minimum = lotSizesSatoshis[0];
```
Shadowfiend, line 583: It feels like we'd do better to ensure the lot size array is in order when we set it, rather than trying to sort it later?

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -566,6 +562,33 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+    /// @dev It is recommended to call this function on tBTC initialization and
+    /// after minimum lot size update.
+    function refreshMinimumBondableValue() public {
+        uint256 bondRequirementSatoshis = getMinimumLotSizeSatoshi().mul(
+            initialCollateralizedPercent
+        ).div(100);
+        uint256 bondRequirementWei = _fetchBitcoinPrice().mul(
+            bondRequirementSatoshis
+        );
+        keepFactorySelection.setMinimumBondableValue(bondRequirementWei);
```
Shadowfiend, line 578: Since we're setting this on the sortition pool, shouldn't it be a per-operator value? That is, shouldn't it be 50% rather than 150%?

pdyraga: The reasoning for using 150% was to set up the bar on a low level but not on the lowest level: we make sure every operator we add to the pool can create a keep only with its members if needed. Given that we talk about the minimum lot size, it sounds to me like a sane expectation especially that we need to account for ETHBTC price fluctuations. Technically, we can lower it to 50% but I am not sure if we want to. -_-

solidity/contracts/system/TBTCSystem.sol
pdyraga: Yeah, I wasn't sure if I can do that. The mandatory lot size looks to me like something we might always pass as the very first value just for convenience but given that it's not a function we call every day, expecting lot sizes to come sorter makes sense. If you agree with this change, I am happy to update the code 🙌

solidity/contracts/system/TBTCSystem.sol
Shadowfiend: The other danger I'm seeing here is that we've really blurred the lines of responsibility on who decides what about bonds:

- The threshold/member count for keeps is static, but it's managed by `DepositFactory`. But it's ultimately passed through `Deposit` to `TBTCSystem` and then finally to the ECDSA keep.
- The math on total bond amount is done in `DepositFunding.sol`, but relies entirely on information that `TBTCSystem` has at its disposal.
- The math that turns the bond amount and n/m count into a per-node bond is done in `BondedECDSAKeep`.

I would suggest that we:
- Move `n`/`m` values into `TBTCSystem`'s constructor and drop them from `DepositFactory`.
- Stop passing these through `DepositFactory` and `Deposit`.
- Have `requestNewKeep` pass those along to `BondedECDSAKeepFactory`.
- Calculate the initial bond requirement that is passed to `BondedECDSAKeepFactory` entirely in `TBTCSystem`, rather than doing it in `Deposit` and passing that info along.
- Reuse the `n`/`m` info and initial bond requirement calculation function in the implementation of `refreshMinimumBondableValue`.

solidity/contracts/system/TBTCSystem.sol
pdyraga: Done in https://github.com/keep-network/tbtc/pull/714/commits/5ec1459464c293954f86bd5edea6f35d9f92bf99.

solidity/contracts/system/TBTCSystem.sol
pdyraga: > EDIT: one more bullet. In a perfect world, we would pass the initial required bond + the n/m info to BondedECDSAKeepFactory and let it decide what the implications are on minimum bondable value.
>
> The whole point here is that the minimum bondable value is as much as possible an implementation detail of BondedECDSAKeepFactory, the Keep setup is as much as possible an implementation detail of TBTCSystem, and the deposit and deposit factory are entirely concerned with the deposit's overall collateralization info.

I am no sure if I understand correctly what you say but I think this is ~ what's happening now. `TBTCSystem` is calculating the bond required for the whole keep and `BondedECDSAKeepFactory` splits this bond between members. 

tBTC needs to be aware of the minimum unbonded value because the refresh needs to happen outside of the function opening a keep. In the previous approach we were updating the minimum unbonded value based on the bonding parameter passed to `BondedECDSAKeepFactory` when opening a keep but it doesn't work well.

I think there is something I am missing here... 🤔

pdyraga:
I marked it as a ready for review, it's ready for another chance @Shadowfiend.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -49,7 +49,7 @@ library KeepFactorySelection {
     function initialize(
         Storage storage _self,
         IBondedECDSAKeepFactory _defaultFactory
-    ) internal {
+    ) public {
```
Shadowfiend, line 52: How come?

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -151,16 +154,16 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         return allowNewDeposits;
     }
 
+    /// @notice Return the lowest lot size currently enabled for deposits.
+    /// @return The lowest lot size, in satoshis.
+    function getMinLotSize() public view returns (uint256) {
```
Shadowfiend, line 159: Hmmm… Now that I see this I kind of feel like we do `Minimum` and `Maximum` here... Thoughts?

solidity/contracts/system/KeepFactorySelection.sol
pdyraga: Oh, I forgot to flag this change. Does it make sense to keep it `internal`? After this change, we have all library functions called from the contract `public` so they are not copied anywhere and stay in the library. All functions used only from the library are `internal`.

solidity/contracts/system/TBTCSystem.sol
pdyraga: Yeah... I did `Min` to keep it consistent with the existing `Max` function but I have the same feelings about `Minimum` and `Maximum`. I'll update it.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -590,23 +594,58 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
-    /// @param _m Minimum number of honest keep members required to sign.
-    /// @param _n Number of members in the keep.
+    /// @param _lotSizeSatoshis Lot size in satoshis.
     /// @param _maxSecuredLifetime Duration of stake lock in seconds.
     /// @return Address of a new keep.
     function requestNewKeep(
-        uint256 _m,
-        uint256 _n,
-        uint256 _bond,
+        uint64 _lotSizeSatoshis,
```
Shadowfiend, line 601: Well now I feel weird about this taking the lot size 🤔 

I think it's a marked improvement though.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -590,23 +594,58 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
+        return _keepFactory.openKeep.value(msg.value)(keepSize, keepThreshold, msg.sender, bond, _maxSecuredLifetime);
+    }
+
+    /// @notice Calculates bond requirement in wei for the given lot size in
+    /// satoshis given the current BTCETH price.
+    /// @param _lotSizeSatoshis Lot size in satoshis.
+    /// @return Bond requirement in wei.
+    function calculateBondRequirementWei(
+        uint256 _lotSizeSatoshis
+    ) public view returns (uint256) {
```
Shadowfiend, line 622: I think this needs to stay `internal` or we're leaking the price info.

pdyraga: Ah, good point. Will update it.

solidity/contracts/system/KeepFactorySelection.sol
Shadowfiend: Oh very good, sorry, I forgot this is a library. That's fine.

solidity/contracts/system/TBTCSystem.sol
pdyraga: Hmm, it feels quite natural to me but I don't have the full context. `Deposit` is initialized with the lot size and `TBTCSystem` does want to work with supported lot sizes. The moment we open a keep is where lot sizes ends and wei bonds come into play.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -638,6 +625,19 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         return false;
     }
 
+    /// @notice Calculates bond requirement in wei for the given lot size in
+    /// satoshis given the current BTCETH price.
```
Shadowfiend, line 629: Super tiny nit that I'm not going to block merge on, but I would say “based on the current ETHBTC price”, mostly because we're using “given” earlier to refer specifically to something that is passed in.

pdyraga: Wait wait, I can edit it 😆

solidity/contracts/system/TBTCSystem.sol
Shadowfiend: It feels like by the time we reach the concept of a keep, we should be requesting the keep with a particular bond for the keep, rather than passing the lot size. But I like that the price feed and bond calculation is here now. So I'm still a little ways away from figuring out what's bothering me about it :) Regardless, I don't think it's worth doing anything more here for now.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -626,7 +626,7 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     }
 
     /// @notice Calculates bond requirement in wei for the given lot size in
-    /// satoshis given the current BTCETH price.
+    /// satoshis based on the current ETHBTC price.
```
Shadowfiend, line 629: :boar:

solidity/contracts/system/TBTCSystem.sol
Shadowfiend: Taking this discussion offline briefly, will summarize the outcome.

pdyraga: We decided to allow operators in a sortition pool who can support only one keep (one member in one keep to be precise). For this reason, we decided for `setMinimumBondValue(minimumBondValue, keepThreshold, keepSize)`, in the `BondedECDSAKeepFactory` that will basically take that and divide `minimumBondValue` by `keepSize` to determine the correct value to set on the sortition pool. This lets the factory be the center of determining how it spreads bonds across operators (which it already is when opening a new keep), while passing along the full info needed for the final setting in the sortition pool.

`keep-ecdsa` change is here: https://github.com/keep-network/keep-ecdsa/pull/510

pdyraga:
Ready for another chance @Shadowfiend 🙌

PR #722, opened 2020-08-25
Use vendors for Keep factories and lock them

nkuba:
Keep ECDSA defines a vendor contract that is used to register factories and upgrade them. It folds a reference to the latest version of a factory. We were using the vendor contract previously but decided to use a fixed factory address in https://github.com/keep-network/tbtc/pull/610.
In this PR we are adding back usage of the vendor contract. The contract will be called to obtain the latest version of a Keep factory for each type: KEEP staked and ETH only backed keeps.

We are adding also a locking mechanism for Keep factories version updates. We have a possibility to update the factory address via keep vendor contract to the latest one. We want to have a way to lock the versions at some point in the future when keep-ecdsa is considered to be in a state that won't require any updates. When the lock is set we will use only the versions of the factories that were current ones in the moment of locking.

Open questions:
- Do we want to have two separate locks for KEEP backed and ETH only keeps factories?
- Do we want to have a two-step update process with a standard governance period for the locking the factories?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
             echo $GCLOUD_SERVICE_KEY > ~/gcloud-service-key.json
             gcloud auth activate-service-account --key-file ~/gcloud-service-key.json
             export UNISWAP_CONTRACT_DATA="uniswap-contract-data.txt"
-            export BONDED_ECDSA_KEEP_FACTORY_CONTRACT_DATA="BondedECDSAKeepFactory.json"
+            export BONDED_ECDSA_KEEP_VENDOR_CONTRACT_DATA="BondedECDSAKeepVendor.json"
             solidity/scripts/circleci-provision-external-contracts.sh
       - run:
           name: Migrate Contracts
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -12,12 +12,12 @@ interface KeepFactorySelector {
     /// @notice Selects keep factory for the new deposit.
     /// @param _seed Request seed.
     /// @param _keepStakeFactory Regular, KEEP-stake based keep factory.
-    /// @param _ethStakeFactory Fully backed, ETH-bond-only based keep factory.
+    /// @param _fullyBackedFactory Fully backed, ETH-bond-only based keep factory.
     /// @return The selected keep factory.
     function selectFactory(
         uint256 _seed,
         IBondedECDSAKeepFactory _keepStakeFactory,
-        IBondedECDSAKeepFactory _ethStakeFactory
+        IBondedECDSAKeepFactory _fullyBackedFactory
     ) external view returns (IBondedECDSAKeepFactory);
 }
 
@@ -44,8 +44,8 @@ library KeepFactorySelection {
         IBondedECDSAKeepFactory keepStakeFactory;
 
         // Fully backed ECDSA keep vendor and factory: ETH bond only.
-        IBondedECDSAKeepVendor ethStakeVendor;
-        IBondedECDSAKeepFactory ethStakeFactory;
+        IBondedECDSAKeepVendor fullyBackedVendor;
+        IBondedECDSAKeepFactory fullyBackedFactory;
 
         // Lock for factories versions freeze. When set to true vendor won't be
         // called to obtain a new factory address but a version of a factory
@@ -123,8 +123,8 @@ library KeepFactorySelection {
                 _honestThreshold
             );
         }
-        if (address(_self.ethStakeFactory) != address(0)) {
-            _self.ethStakeFactory.setMinimumBondableValue(
+        if (address(_self.fullyBackedFactory) != address(0)) {
+            _self.fullyBackedFactory.setMinimumBondableValue(
                 _minimumBondableValue,
                 _groupSize,
                 _honestThreshold
@@ -143,7 +143,7 @@ library KeepFactorySelection {
         IBondedECDSAKeepFactory keepStakeFactory = getKeepStakedFactory(_self);
 
         if (
-            address(_self.ethStakeFactory) == address(0) ||
+            address(_self.fullyBackedFactory) == address(0) ||
             address(_self.factorySelector) == address(0)
         ) {
             // KEEP-stake factory is guaranteed to be there. If the selection
@@ -152,7 +152,7 @@ library KeepFactorySelection {
             return;
         }
 
-        IBondedECDSAKeepFactory ethStakeFactory = getFullyBackedFactory(_self);
+        IBondedECDSAKeepFactory fullyBackedFactory = getFullyBackedFactory(_self);
 
         _self.requestCounter++;
         uint256 seed = uint256(
@@ -161,12 +161,12 @@ library KeepFactorySelection {
         _self.selectedFactory = _self.factorySelector.selectFactory(
             seed,
             keepStakeFactory,
-            ethStakeFactory
+            fullyBackedFactory
         );
 
         require(
             _self.selectedFactory == keepStakeFactory ||
-                _self.selectedFactory == ethStakeFactory,
+                _self.selectedFactory == fullyBackedFactory,
             "Factory selector returned unknown factory"
         );
     }
@@ -202,18 +202,18 @@ library KeepFactorySelection {
         returns (IBondedECDSAKeepFactory)
     {
         if (_self.factoriesVersionsLock) {
-            return _self.ethStakeFactory;
+            return _self.fullyBackedFactory;
         } else {
-            IBondedECDSAKeepFactory ethStakeFactory = IBondedECDSAKeepFactory(
-                _self.ethStakeVendor.selectFactory()
+            IBondedECDSAKeepFactory fullyBackedFactory = IBondedECDSAKeepFactory(
+                _self.fullyBackedVendor.selectFactory()
             );
 
             require(
-                address(ethStakeFactory) != address(0),
+                address(fullyBackedFactory) != address(0),
                 "Vendor returned invalid factory address"
             );
 
-            return ethStakeFactory;
+            return fullyBackedFactory;
         }
     }
 
@@ -231,20 +231,20 @@ library KeepFactorySelection {
         address _fullyBackedVendor
     ) internal {
         require(
-            address(_self.ethStakeVendor) == address(0),
+            address(_self.fullyBackedVendor) == address(0),
             "Fully backed vendor already set"
         );
         require(_fullyBackedVendor != address(0), "Invalid address");
 
-        IBondedECDSAKeepVendor ethStakeVendor = IBondedECDSAKeepVendor(
+        IBondedECDSAKeepVendor fullyBackedVendor = IBondedECDSAKeepVendor(
             _fullyBackedVendor
         );
-        address ethStakeFactory = ethStakeVendor.selectFactory();
+        address fullyBackedFactory = fullyBackedVendor.selectFactory();
 
-        require(ethStakeFactory != address(0), "Vendor returned invalid factory address");
+        require(fullyBackedFactory != address(0), "Vendor returned invalid factory address");
 
-        _self.ethStakeVendor = ethStakeVendor;
-        _self.ethStakeFactory = IBondedECDSAKeepFactory(ethStakeFactory);
+        _self.fullyBackedVendor = fullyBackedVendor;
+        _self.fullyBackedFactory = IBondedECDSAKeepFactory(fullyBackedFactory);
     }
 
     /// @notice Sets the address of the keep factory selection strategy contract.
@@ -289,14 +289,14 @@ library KeepFactorySelection {
             "KEEP backed vendor not set"
         );
         require(
-            address(_self.ethStakeVendor) != address(0),
+            address(_self.fullyBackedVendor) != address(0),
             "Fully backed vendor not set"
         );
 
         IBondedECDSAKeepFactory latestKeepStakeFactory = getKeepStakedFactory(
             _self
         );
-        IBondedECDSAKeepFactory latestEthStakeFactory = getFullyBackedFactory(
+        IBondedECDSAKeepFactory latestFullyBackedFactory = getFullyBackedFactory(
             _self
         );
 
@@ -305,12 +305,12 @@ library KeepFactorySelection {
             "Unexpected KEEP backed factory"
         );
         require(
-            address(latestEthStakeFactory) == _expectedFullyBackedFactory,
+            address(latestFullyBackedFactory) == _expectedFullyBackedFactory,
             "Unexpected fully backed factory"
         );
 
         _self.keepStakeFactory = latestKeepStakeFactory;
-        _self.ethStakeFactory = latestEthStakeFactory;
+        _self.fullyBackedFactory = latestFullyBackedFactory;
 
         _self.factoriesVersionsLock = true;
     }
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -122,19 +122,33 @@ library KeepFactorySelection {
         uint256 _groupSize,
         uint256 _honestThreshold
     ) public {
-        if (address(_self.keepStakeFactory) != address(0)) {
-            _self.keepStakeFactory.setMinimumBondableValue(
+        require(
+            address(_self.keepStakeVendor) != address(0),
+            "KEEP backed vendor not set"
+        );
+
+        IBondedECDSAKeepFactory keepStakeFactory = getKeepStakedFactory(_self);
+
+        if (address(keepStakeFactory) != address(0)) {
+            keepStakeFactory.setMinimumBondableValue(
                 _minimumBondableValue,
                 _groupSize,
                 _honestThreshold
             );
         }
-        if (address(_self.fullyBackedFactory) != address(0)) {
-            _self.fullyBackedFactory.setMinimumBondableValue(
-                _minimumBondableValue,
-                _groupSize,
-                _honestThreshold
+
+        if (address(_self.fullyBackedVendor) != address(0)) {
+            IBondedECDSAKeepFactory fullyBackedFactory = getFullyBackedFactory(
+                _self
             );
+
+            if (address(fullyBackedFactory) != address(0)) {
+                fullyBackedFactory.setMinimumBondableValue(
+                    _minimumBondableValue,
+                    _groupSize,
+                    _honestThreshold
+                );
+            }
         }
     }
 
@@ -257,10 +271,7 @@ library KeepFactorySelection {
             address(_self.factorySelector) == address(0),
             "Factory selector already set"
         );
-        require(
-            address(_factorySelector) != address(0),
-            "Invalid address"
-        );
+        require(address(_factorySelector) != address(0), "Invalid address");
 
         _self.factorySelector = KeepFactorySelector(_factorySelector);
     }
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -49,7 +49,7 @@ library KeepFactorySelection {
 
         // Lock for factories versions freeze. When set to true vendor won't be
         // called to obtain a new factory address but a version of a factory
-        // from the moment of freezing will be used. Once locked the factory
+        // from the moment of freezing will be used, once locked the factory
         // address won't be able to update anymore.
         bool factoriesVersionsLocked;
     }
@@ -213,7 +213,7 @@ library KeepFactorySelection {
     }
 
     /// @notice Returns ETH-bond-only based factory address. If factories lock is not set
-    /// it calls the ETH staked vendor to obtain the latest versions of the factory.
+    /// it calls the fully-backed vendor to obtain the latest versions of the factory.
     function getFullyBackedFactory(Storage storage _self)
         internal
         view
@@ -285,7 +285,8 @@ library KeepFactorySelection {
     /// @notice Locks versions of factories. When lock is set vendor contracts
     /// won't be called anymore to obtain the latest factories versions.
     /// It requires expected factories addresses to be provided to protect from
-    /// locking on unexpected addresses.
+    /// locking on unexpected addresses. The function can be called only once and
+    /// the lock cannot be disabled.
     /// @param _expectedKeepStakedFactory Expected KEEP-staked factory address
     /// @param _expectedFullyBackedFactory Expected ETH-bond-only factory address
     function lockFactoriesVersions(
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -51,7 +51,7 @@ library KeepFactorySelection {
         // called to obtain a new factory address but a version of a factory
         // from the moment of freezing will be used. Once locked the factory
         // address won't be able to update anymore.
-        bool factoriesVersionsLock;
+        bool factoriesVersionsLocked;
     }
 
     /// @notice Initializes the library with the default KEEP-stake-based
@@ -204,7 +204,7 @@ library KeepFactorySelection {
         view
         returns (IBondedECDSAKeepFactory)
     {
-        if (_self.factoriesVersionsLock) {
+        if (_self.factoriesVersionsLocked) {
             return _self.keepStakeFactory;
         } else {
             return
@@ -219,7 +219,7 @@ library KeepFactorySelection {
         view
         returns (IBondedECDSAKeepFactory)
     {
-        if (_self.factoriesVersionsLock) {
+        if (_self.factoriesVersionsLocked) {
             return _self.fullyBackedFactory;
         } else {
             return
@@ -293,7 +293,7 @@ library KeepFactorySelection {
         address _expectedKeepStakeFactory,
         address _expectedFullyBackedFactory
     ) internal {
-        require(!_self.factoriesVersionsLock, "Already locked");
+        require(!_self.factoriesVersionsLocked, "Already locked");
 
         require(
             address(_self.keepStakeVendor) != address(0),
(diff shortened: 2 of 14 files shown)


pdyraga:
> Do we want to have a two-step update process with a standard governance period for locking the factories?

I don't think it is required. In this particular case, we are not talking about updating to a new value but about staying with the current value forever.

> Do we want to have two separate locks for KEEP backed and ETH only keeps factories?

I do not think it is worth the complexity, to be honest.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -150,29 +170,80 @@ library KeepFactorySelection {
...
+            require(
+                address(keepStakeFactory) != address(0),
+                "Vendor returned invalid factory address"
+            );
+
+            return keepStakeFactory;
+        }
+    }
+
+    /// @notice Returns ETH-stake based factory address. If factories lock is not set
```
pdyraga, line 196: `ETH-stake based factoryh` -> `ETH-bond-only based factory`?

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -34,28 +38,38 @@ library KeepFactorySelection {
...
         KeepFactorySelector factorySelector;
 
-        // Standard ECDSA keep factory: KEEP stake and ETH bond.
+        // Standard ECDSA keep vendor and factory: KEEP stake and ETH bond.
         // Guaranteed to be set for initialized factory.
+        IBondedECDSAKeepVendor keepStakeVendor;
         IBondedECDSAKeepFactory keepStakeFactory;
 
-        // Fully backed ECDSA keep factory: ETH stake and ETH bond.
+        // Fully backed ECDSA keep vendor and factory: ETH stake and ETH bond.
```
pdyraga, line 46: It's an _ETH-bond-only_ option to be precise.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -122,7 +136,11 @@ library KeepFactorySelection {
     /// Otherwise, calls selection strategy providing addresses of both
     /// factories to make a choice. Additionally, passes the selection seed
     /// evaluated from the current request counter value.
+    /// Unless lock is set, it calls vendors to obtain the latests factories
+    /// versions.
     function refreshFactory(Storage storage _self) internal {
+        _self.keepStakeFactory = getKeepStakedFactory(_self);
```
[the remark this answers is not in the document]
pdyraga: Actually, does it make sense to write them here? The current solution is correct bug gas-inefficient. Let's assume we have `_self.factorySelector` set. Then, for each factory selection, we perform calls to two vendors and `SSTORE` their addresses. This is not efficient, because we do not need to update `_self.keepStakeFactory` and `_self.ethStakeFactory`, we can pass the in-memory references to `self.factorySelector.selectFactory`. Why to `SSTORE` them, given that are interested in `selectedFactory` in `selectFactory`.

I suggest fetching two factories from vendors and passing then to `selectFactory` but not writing anything else in this function than `selectedFactory`. Instead, we can write them in `lockFactoriesVersions`.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -150,29 +170,80 @@ library KeepFactorySelection {
         );
     }
 
+    /// @notice Returns KEEP staked factory address. If factories lock is not set
```
pdyraga, line 173: `KEEP staked factory` -> `KEEP stake based factory` perhaps?

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -150,29 +170,80 @@ library KeepFactorySelection {
...
     /// Once both fully-backed keep factory and factory selection strategy are
     /// set, KEEP-stake-based factory is no longer the default choice and it is
     /// up to the selection strategy to decide which factory should be chosen.
     /// @dev Can be called only one time!
-    /// @param _fullyBackedFactory Address of the fully-backed, ETH-stake based
-    /// keep factory.
-    function setFullyBackedKeepFactory(
+    /// @param _fullyBackedVendor Address of the fully-backed, ETH-stake based
+    /// keep vendor.
+    function setFullyBackedKeepVendor(
```
pdyraga, line 228: `setFullyBackedKeepFactoryVendor` ?

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -34,28 +38,38 @@ library KeepFactorySelection {
...
 
-        // Standard ECDSA keep factory: KEEP stake and ETH bond.
+        // Standard ECDSA keep vendor and factory: KEEP stake and ETH bond.
         // Guaranteed to be set for initialized factory.
+        IBondedECDSAKeepVendor keepStakeVendor;
         IBondedECDSAKeepFactory keepStakeFactory;
 
-        // Fully backed ECDSA keep factory: ETH stake and ETH bond.
+        // Fully backed ECDSA keep vendor and factory: ETH stake and ETH bond.
+        IBondedECDSAKeepVendor ethStakeVendor;
```
pdyraga, line 47: `fullyBackedFactoryVendor` perhaps? We do not have ETH stake.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -34,28 +38,38 @@ library KeepFactorySelection {
 
         KeepFactorySelector factorySelector;
 
-        // Standard ECDSA keep factory: KEEP stake and ETH bond.
+        // Standard ECDSA keep vendor and factory: KEEP stake and ETH bond.
         // Guaranteed to be set for initialized factory.
+        IBondedECDSAKeepVendor keepStakeVendor;
```
pdyraga, line 43: `keepStakeFactoryVendor` maybe? We are not vendoring keeps, we are vendoring factories.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -150,29 +170,80 @@ library KeepFactorySelection {
...
+        IBondedECDSAKeepVendor ethStakeVendor = IBondedECDSAKeepVendor(
+            _fullyBackedVendor
         );
+        address ethStakeFactory = ethStakeVendor.selectFactory();
 
-        _self.ethStakeFactory = IBondedECDSAKeepFactory(_fullyBackedFactory);
+        require(ethStakeFactory != address(0), "Vendor returned invalid factory address");
+
+        _self.ethStakeVendor = ethStakeVendor;
+        _self.ethStakeFactory = IBondedECDSAKeepFactory(ethStakeFactory);
```
pdyraga, line 246: It probably makes sense to check whether the vendor returns non-zero factory but I think we shouldn't write the factory here. It's not the responsibility of this function.

pdyraga (commented):
Posted some thoughts regarding the selection contract, wasn't reviewing unit tests.

solidity/contracts/system/KeepFactorySelection.sol
nkuba: I'm not sure, it's `BondedECDSAKeepVendor` not `BondedECDSAKeepFactoryVendor`.

solidity/contracts/system/KeepFactorySelection.sol
nkuba: In keep-ecdsa it is called just `vendor` not `factory vendor`.

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 5f95e34cfbdf1aeb14b89995d03854fea7bfef62

solidity/contracts/system/KeepFactorySelection.sol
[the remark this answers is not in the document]
nkuba: 5de8998a76ab26dd5df83fbdf229a874a7430f3d

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -22,9 +23,12 @@ interface KeepFactorySelector {
 
 /// @title Bonded ECDSA keep factory selection library.
 /// @notice tBTC uses two bonded ECDSA keep factories: one based on KEEP stake
-/// and ETH bond, and another based on ETH stake and ETH bond. The library holds
-/// a reference to both factories as well as a reference to a selection strategy
-/// deciding which factory to choose for the new deposit being opened.
+/// and ETH bond, and another based on ETH stake and ETH bond. Factories addresses
```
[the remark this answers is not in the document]
nkuba: 28c7611607a4765f597f074a7bf168476eb119c9

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -34,28 +38,38 @@ library KeepFactorySelection {
...
         IBondedECDSAKeepFactory keepStakeFactory;
 
-        // Fully backed ECDSA keep factory: ETH stake and ETH bond.
+        // Fully backed ECDSA keep vendor and factory: ETH stake and ETH bond.
+        IBondedECDSAKeepVendor ethStakeVendor;
         IBondedECDSAKeepFactory ethStakeFactory;
+
+        // Lock for factories versions freeze. When set to true vendor won't be
+        // called to obtain a new factory address but the latests factory will
+        // be used.
```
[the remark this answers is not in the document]
nkuba: 28c7611607a4765f597f074a7bf168476eb119c9

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 28c7611607a4765f597f074a7bf168476eb119c9

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 28c7611607a4765f597f074a7bf168476eb119c9

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 86197d5831142ed319553997437a898b36d31b89, f3137845cb96e549f647819a64f96844fad69014

solidity/contracts/system/KeepFactorySelection.sol
pdyraga: Yep, you are right. Resolving.

solidity/contracts/system/KeepFactorySelection.sol
pdyraga: So maybe `keepStakedVendor` and `keepStakedFactory`? I still don't love it but I think it's a little better, especially that we have `fullyBackedVendor` and `fullyBackedFactory`.

.circleci/config.yml
```diff
@@ -84,7 +84,7 @@ jobs:
             echo $GCLOUD_SERVICE_KEY > ~/gcloud-service-key.json
             gcloud auth activate-service-account --key-file ~/gcloud-service-key.json
             export UNISWAP_CONTRACT_DATA="uniswap-contract-data.txt"
-            export BONDED_ECDSA_KEEP_FACTORY_CONTRACT_DATA="BondedECDSAKeepFactory.json"
+            export BONDED_ECDSA_KEEP_VENDOR_CONTRACT_DATA="BondedECDSAKeepVendor.json"
```
pdyraga, line 87: In `circleci-provision-external-contracts.sh` we still talk about `BONDED_ECDSA_KEEP_FACTORY_ADDRESS` and set `BondedECDSAKeepFactoryAddress`

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -34,29 +38,46 @@ library KeepFactorySelection {
...
 
-        // Fully backed ECDSA keep factory: ETH stake and ETH bond.
-        IBondedECDSAKeepFactory ethStakeFactory;
+        // Fully backed ECDSA keep vendor and factory: ETH bond only.
+        IBondedECDSAKeepVendor fullyBackedVendor;
+        IBondedECDSAKeepFactory fullyBackedFactory;
+
+        // Lock for factories versions freeze. When set to true vendor won't be
+        // called to obtain a new factory address but a version of a factory
+        // from the moment of freezing will be used. Once locked the factory
```
pdyraga, line 52: I believe we should have a coma after *Once locked*.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -101,78 +122,137 @@ library KeepFactorySelection {
...
+
         if (
-            address(_self.ethStakeFactory) == address(0) ||
+            address(_self.fullyBackedVendor) == address(0) ||
             address(_self.factorySelector) == address(0)
         ) {
             // KEEP-stake factory is guaranteed to be there. If the selection
             // can not be performed, this is the default choice.
-            _self.selectedFactory = _self.keepStakeFactory;
+            _self.selectedFactory = keepStakeFactory;
```
pdyraga, line 171: Can we make a small optimization and compare the value before it's set?

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -34,29 +38,46 @@ library KeepFactorySelection {
...
-        IBondedECDSAKeepFactory ethStakeFactory;
+        // Fully backed ECDSA keep vendor and factory: ETH bond only.
+        IBondedECDSAKeepVendor fullyBackedVendor;
+        IBondedECDSAKeepFactory fullyBackedFactory;
+
+        // Lock for factories versions freeze. When set to true vendor won't be
+        // called to obtain a new factory address but a version of a factory
+        // from the moment of freezing will be used. Once locked the factory
+        // address won't be able to update anymore.
+        bool factoriesVersionsLock;
```
pdyraga, line 54: Thoughts on naming it `factoriesVersionsLocked`?

`if (_self.factoriesVersionsLocked)` reads better than `if (_self.factoriesVersionsLock)`.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -101,78 +122,137 @@ library KeepFactorySelection {
...
+        if (_self.factoriesVersionsLock) {
+            return _self.keepStakeFactory;
+        } else {
+            return
+                IBondedECDSAKeepFactory(_self.keepStakeVendor.selectFactory());
+        }
+    }
+
+    /// @notice Returns ETH-bond-only based factory address. If factories lock is not set
+    /// it calls the ETH staked vendor to obtain the latest versions of the factory.
```
pdyraga, line 210: `/s/ETH staked vendor/fully-backed vendor`?

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -191,11 +271,52 @@ library KeepFactorySelection {
...
         );
+        require(address(_factorySelector) != address(0), "Invalid address");
+
+        _self.factorySelector = KeepFactorySelector(_factorySelector);
+    }
+
+    /// @notice Locks versions of factories. When lock is set vendor contracts
+    /// won't be called anymore to obtain the latest factories versions.
+    /// It requires expected factories addresses to be provided to protect from
+    /// locking on unexpected addresses.
```
pdyraga, line 282: We should make it clear in docs this function can be called only one time and that the version lock can not be later disabled.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -191,11 +271,52 @@ library KeepFactorySelection {
...
+            .fullyBackedVendor
+            .selectFactory();
+
+        require(
+            address(latestKeepStakeFactory) == _expectedKeepStakeFactory,
+            "Unexpected KEEP backed factory"
+        );
+        require(
+            address(latestFullyBackedFactory) == _expectedFullyBackedFactory,
+            "Unexpected fully backed factory"
```
pdyraga, line 314: `KEEP backed` and `fully backed` are confusing. I'd probably go with `KEEP-staked factory` and `fully backed factory`.

solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -191,11 +271,52 @@ library KeepFactorySelection {
...
+        );
+
+        _self.keepStakeFactory = IBondedECDSAKeepFactory(latestKeepStakeFactory);
+        _self.fullyBackedFactory = IBondedECDSAKeepFactory(latestFullyBackedFactory);
+
+        _self.factoriesVersionsLock = true;
```
pdyraga, line 320: Let's maybe lock before we call `selectFactory()` on any of the vendors. I don't expect any reentrancy issues here but better be safe than sorry.

solidity/contracts/system/KeepFactorySelection.sol
nkuba: I'm not sure what you mean. Compare to what?

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -468,6 +473,27 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         emit EthBtcPriceFeedAdded(address(_nextEthBtcPriceFeed));
     }
 
+    /// @notice Locks versions of Keep factories. When lock is set vendor contracts
+    /// won't be called anymore to obtain the latest Keep factories versions.
+    /// It requires expected factories addresses to be provided to protect from
+    /// locking on unexpected addresses.
+    /// @param _expectedKeepStakeFactory Expected KEEP-staked factory address
+    /// @param _expectedFullyBackedFactory Expected ETH-staked factory address
```
pdyraga, line 481: There is no ETH-staked factory, we should say about ETH-bond-only or fully backed.

solidity/contracts/system/KeepFactorySelection.sol
pdyraga: Compare to what's already in `_self.selectedFactory`. If the value stored in `_self.selectedFactory` is the same as the one returned from `getKeepStakedFactory(_self)`, we don't need to write to `_self.selectedFactory` saving gas on SSTORE operation.

solidity/test/GovernanceTest.js
```diff
@@ -180,55 +185,139 @@ describe("TBTCSystem governance", async function() {
...
+      await expectRevert(
+        tbtcSystem.lockKeepFactoriesVersionsUpdates(
+          ecdsaKeepFactory.address,
+          ethStakedFactory.address,
+        ),
+        "Already locked",
+      )
+    })
+
+    it("reverts when lockKeepFactoriesVersionsUpdates is by non-owner", async () => {
```
pdyraga, line 297: Maybe just `reverts when not performed by owner`?

solidity/contracts/system/KeepFactorySelection.sol
nkuba: b9b3b57612e0774872a036c26c1005f2802c1c2d

solidity/contracts/system/KeepFactorySelection.sol
nkuba: c3c89f41034144090067927df6624e14d1d0a359

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 88d7ca48e4f96d101b318379d39d6bea45a203c7

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 931ab0090a6329a62b1d378483d8045a031f3828

solidity/test/GovernanceTest.js
nkuba: Wanted to be consistent with other names in this file. I'll update it.

solidity/contracts/system/KeepFactorySelection.sol
nkuba: 57274307025e2ba52329e3ae112cae4888653c2d

solidity/contracts/system/TBTCSystem.sol
nkuba: b35547358432c1031368e630f51602fd1146d732

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -288,47 +293,47 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
     ///         balances between two factories based on the selection strategy.
-    /// @dev It can be finalized by calling `finalizeKeepFactorySingleShotUpdate`
+    /// @dev It can be finalized by calling `finalizeKeepVendorSingleShotUpdate`
     ///      any time after `governanceTimeDelay` has elapsed. This can be
     ///      called more than once until finalized to reset the values and
     ///      timer, but it can only be finalized once!
     /// @param _factorySelector Address of the keep factory selection strategy.
-    /// @param _ethBackedFactory Address of the ETH-stake-based factory.
-    function beginKeepFactorySingleShotUpdate(
+    /// @param _ethBackedVendor Address of the ETH-stake-based vendor.
```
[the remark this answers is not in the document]
nkuba: b35547358432c1031368e630f51602fd1146d732

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -53,9 +53,14 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
...
-    event KeepFactorySingleShotUpdated(
+    event KeepVendorSingleShotUpdated(
         address _factorySelector,
-        address _ethBackedFactory
+        address _ethBackedVendor
+    );
+
+    event KeepFactoriesVersionsLocked(
+        address _keepStakeFactory,
+        address _fullyBackedFactory
```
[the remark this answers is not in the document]
nkuba: 945e2f6f50fbc873fd7e34ab4932d42a87f7f1f9

solidity/contracts/system/KeepFactorySelection.sol
nkuba: b35547358432c1031368e630f51602fd1146d732

solidity/test/GovernanceTest.js
nkuba: d9ac37d0d67048873e2cd54f7b6227b309178069

solidity/test/KeepFactorySelectionTest.js
```diff
@@ -212,27 +360,193 @@ describe("KeepFactorySelection", async () => {
...
+
+      // Set minimum bondable value
+      await keepFactorySelection.setMinimumBondableValue(newValue, 5, 3)
+
+      expect(await keepStakedFactory.minimumBondableValue()).to.eq.BN(newValue)
+      expect(await fullyBackedFactory.minimumBondableValue()).to.eq.BN(newValue)
+
+      expect(await newKeepFactory.minimumBondableValue()).to.eq.BN(defaultValue)
+      expect(await newEthFactory.minimumBondableValue()).to.eq.BN(defaultValue)
+    })
```
pdyraga, line 495: I think we are missing a test case when:
- KEEP stake vendor is set, factory is set
- ETH-only vendor is set, factory is set
- versions of factories are locked

solidity/test/KeepFactorySelectionTest.js
```diff
@@ -212,27 +360,193 @@ describe("KeepFactorySelection", async () => {
...
+      await keepFactorySelector.setFullyBackedMode()
+
+      // refresh the choice
+      await keepFactorySelection.selectFactoryAndRefresh()
+
+      expect(
+        await keepFactorySelection.selectFactory(),
+        "unexpected factory selected",
+      ).to.equal(fullyBackedFactory.address)
+    })
```
pdyraga, line 538: Does this case belong to `describe("setFullyBackedKeepVendor"` ? I think this behavior is pretty well covered before in all the selection tests we are having.

pdyraga (commented):
Did a full pass through all the code, it looks good to me. Left two additional comments regarding tests but otherwise, I am good to merge. :+1:

Given the size of the changes, I'd appreciate another person doing a review. I am especially interested in @Shadowfiend's opinion given his tBTC expertize.

@nkuba Do we have any sense on how this change affects the gas cost of selecting the factory?

Shadowfiend:
I don't have time to do a full review on this PR… However 🤔 giving this PR a quick look, we seem to have added a lot of code to do something extremely similar to what we were already doing with the factory single-shot update. What do we gain from reimplementing the vendor mechanism, vs allowing the factory updating to update both ETH and non-ETH backed factories, and giving it a max time limit (like emergency pause)? This seems most similar to already-implemented mechanisms, conforms to how we handle the few tBTC condition changes in general (with clear events on either end, emitted on TBTCSystem) and also feels a bit simpler than trying to gate things in the factory selector, which wasn't *really* designed to do this in the first place. It also reduces overall storage footprint, which seems desirable.

I know it's a bit late in the game to suggest this, so I'll still plan on doing a full review of the code tomorrow morning.

pdyraga:
> I know it's a bit late in the game to suggest this, so I'll still plan on doing a full review of the code tomorrow morning.

I like it. I think it's still worth trying it. I'd recommend a separate PR so that we can have this PR as a backup plan.

The original idea behind a vendor was that we could load-balance between factories based on how many stakers migrated to the new operator contract, or - in case of this implementation - switch to the new factory as soon as we have at least N weight of the pool. But given the complexity we have to introduce to make it happen, I think it may not be worth it and we can just use a manual process. 

Another advantage I see in @Shadowfiend's suggestion is that we don't need to maintain ECDSA contract upgradeability. ECDSA contracts will not be upgradeable and the only upgrade that may happen is that tBTC switches to a new version.

solidity/test/KeepFactorySelectionTest.js
nkuba: > I think we are missing a test case when:

There are test cases covering these scenarios:

>     * KEEP stake vendor is set, factory is set
`updates value in KEEP stake factory`

>     * ETH-only vendor is set, factory is set
`updates value in ETH-only factory`

>     * versions of factories are locked
`updates value in locked factories`

nkuba:
Closing this in favor of https://github.com/keep-network/tbtc/pull/724

PR #724, opened 2020-08-26
Keep factories updates

nkuba:
Currently, the system allows a single-shot update of fully backed keep factory and keep factory selector addresses. KEEP staked factory address is not upgradeable.

In this PR we added a possibility to upgrade all three addresses within a certain time period (180 days) from the system initialization.

This is a two-step upgrade that can be finalized after 48 hours of governance delay.

To begin the upgrade owner has to provide three addresses:
- KEEP-stake backed keep factory,
- ETH-only-bonded keep factory,
- Keep factory selector.

It is required that KEEP-stake backed factory address is always provided to a valid value different than 0. Remaining two addresses are optional for the system to work, so zero addresses are allowed. This is also a safety switch that lets us disable fully backed keep factory and operate only on KEEP-staked factory.

### DIFF
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -182,11 +182,6 @@ library KeepFactorySelection {
         Storage storage _self,
         address _fullyBackedFactory
     ) internal {
-        require(
-            address(_fullyBackedFactory) != address(0),
-            "Invalid address"
-        );
-
         _self.fullyBackedFactory = IBondedECDSAKeepFactory(_fullyBackedFactory);
     }
 
@@ -203,11 +198,6 @@ library KeepFactorySelection {
         Storage storage _self,
         address _factorySelector
     ) internal {
-        require(
-            address(_factorySelector) != address(0),
-            "Invalid address"
-        );
-
         _self.factorySelector = KeepFactorySelector(_factorySelector);
     }
 }
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -150,23 +150,38 @@ library KeepFactorySelection {
         );
     }
 
+    /// @notice Sets the address of the KEEP-stake based keep factory. This is
+    /// a default factory used when a ETH-stake based keep factory is not set.
+    /// @dev Can be called multiple times! It's responsibility of a contract
+    /// using this library to limit and protect updates.
+    /// @param _keepStakedFactory Address of the regular, KEEP-stake based keep
+    /// factory.
+    function setKeepStakedKeepFactory(
+        Storage storage _self,
+        address _keepStakedFactory
+    ) internal {
+        require(
+            address(_keepStakedFactory) != address(0),
+            "Invalid address"
+        );
+
+        _self.keepStakeFactory = IBondedECDSAKeepFactory(_keepStakedFactory);
+    }
+
     /// @notice Sets the address of the fully backed, ETH-stake based keep
     /// factory. KeepFactorySelection can work without the fully-backed keep
     /// factory set, always selecting the default KEEP-stake-based factory.
     /// Once both fully-backed keep factory and factory selection strategy are
     /// set, KEEP-stake-based factory is no longer the default choice and it is
     /// up to the selection strategy to decide which factory should be chosen.
-    /// @dev Can be called only one time!
+    /// @dev Can be called multiple times! It's responsibility of a contract
+    /// using this library to limit and protect updates.
     /// @param _fullyBackedFactory Address of the fully-backed, ETH-stake based
     /// keep factory.
     function setFullyBackedKeepFactory(
         Storage storage _self,
         address _fullyBackedFactory
     ) internal {
-        require(
-            address(_self.ethStakeFactory) == address(0),
-            "Fully backed factory already set"
-        );
         require(
             address(_fullyBackedFactory) != address(0),
             "Invalid address"
@@ -181,16 +196,13 @@ library KeepFactorySelection {
     /// Once both fully-backed keep factory and factory selection strategy are
     /// set, KEEP-stake-based factory is no longer the default choice and it is
     /// up to the selection strategy to decide which factory should be chosen.
-    /// @dev Can be called only one time!
+    /// @dev Can be called multiple times! It's responsibility of a contract
+    /// using this library to limit and protect updates.
     /// @param _factorySelector Address of the keep factory selection strategy.
     function setKeepFactorySelector(
         Storage storage _self,
         address _factorySelector
     ) internal {
-        require(
-            address(_self.factorySelector) == address(0),
-            "Factory selector already set"
-        );
         require(
             address(_factorySelector) != address(0),
             "Invalid address"
--- a/solidity/contracts/system/KeepFactorySelection.sol
+++ b/solidity/contracts/system/KeepFactorySelection.sol
@@ -150,42 +150,8 @@ library KeepFactorySelection {
         );
     }
 
-    /// @notice Sets the address of the KEEP-stake based keep factory. This is
-    /// a default factory used when a ETH-bond-only based keep factory is not set.
-    /// @dev Can be called multiple times! It's responsibility of a contract
-    /// using this library to limit and protect updates.
-    /// @param _keepStakedFactory Address of the regular, KEEP-stake based keep
-    /// factory.
-    function setKeepStakedKeepFactory(
-        Storage storage _self,
-        address _keepStakedFactory
-    ) internal {
-        require(
-            address(_keepStakedFactory) != address(0),
-            "Invalid address"
-        );
-
-        _self.keepStakedFactory = IBondedECDSAKeepFactory(_keepStakedFactory);
-    }
-
-    /// @notice Sets the address of the fully backed, ETH-bond-only based keep
-    /// factory. KeepFactorySelection can work without the fully-backed keep
-    /// factory set, always selecting the default KEEP-stake-based factory.
-    /// Once both fully-backed keep factory and factory selection strategy are
-    /// set, KEEP-stake-based factory is no longer the default choice and it is
-    /// up to the selection strategy to decide which factory should be chosen.
-    /// @dev Can be called multiple times! It's responsibility of a contract
-    /// using this library to limit and protect updates.
-    /// @param _fullyBackedFactory Address of the fully-backed, ETH-bond-only based
-    /// keep factory.
-    function setFullyBackedKeepFactory(
-        Storage storage _self,
-        address _fullyBackedFactory
-    ) internal {
-        _self.fullyBackedFactory = IBondedECDSAKeepFactory(_fullyBackedFactory);
-    }
-
-    /// @notice Sets the address of the keep factory selection strategy contract.
+    /// @notice Sets addresses of the keep factories and the selection strategy
+    /// contracts.
     /// KeepFactorySelection can work without the keep factory selection
     /// strategy set, always selecting the default KEEP-stake-based factory.
     /// Once both fully-backed keep factory and factory selection strategy are
@@ -193,11 +159,24 @@ library KeepFactorySelection {
     /// up to the selection strategy to decide which factory should be chosen.
     /// @dev Can be called multiple times! It's responsibility of a contract
     /// using this library to limit and protect updates.
+    /// @param _keepStakedFactory Address of the regular, KEEP-stake based keep
+    /// factory.
+    /// @param _fullyBackedFactory Address of the fully-backed, ETH-bond-only based
+    /// keep factory.
     /// @param _factorySelector Address of the keep factory selection strategy.
-    function setKeepFactorySelector(
+    function setFactories(
         Storage storage _self,
+        address _keepStakedFactory,
+        address _fullyBackedFactory,
         address _factorySelector
     ) internal {
+        require(
+            address(_keepStakedFactory) != address(0),
+            "Invalid KEEP-staked factory address"
+        );
+
+        _self.keepStakedFactory = IBondedECDSAKeepFactory(_keepStakedFactory);
+        _self.fullyBackedFactory = IBondedECDSAKeepFactory(_fullyBackedFactory);
         _self.factorySelector = KeepFactorySelector(_factorySelector);
     }
 }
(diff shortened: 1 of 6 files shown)


solidity/contracts/system/KeepFactorySelection.sol
```diff
@@ -139,40 +139,55 @@ library KeepFactorySelection {
...
     }
 
-    /// @notice Sets the address of the fully backed, ETH-stake based keep
+    /// @notice Sets the address of the KEEP-stake based keep factory. This is
+    /// a default factory used when a ETH-bond-only based keep factory is not set.
+    /// @dev Can be called multiple times! It's responsibility of a contract
+    /// using this library to limit and protect updates.
+    /// @param _keepStakedFactory Address of the regular, KEEP-stake based keep
+    /// factory.
+    function setKeepStakedKeepFactory(
```
nkuba, line 159: I don't like this name... Since we're in `KeepFactorySelection` maybe it could be named `setKeepStakedFactory` and `setFullyBackedKeepFactory` could be renamed to `setFullyBackedFactory`? 

We could even go further and rename `setKeepFactorySelector` to `setFactorySelector`...

We would use it in `TBTCSystem` as following:
```solidity
keepFactorySelection.setKeepStakedFactory(newKeepStakedFactory);
keepFactorySelection.setFullyBackedFactory(newFullyBackedFactory);
keepFactorySelection.setFactorySelector(newFactorySelector);
```

Thoughts?

Shadowfiend: Per your question earlier on flowdock, should we just combine both factory settings into a single `setFactories` call?

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -38,9 +38,9 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
         uint16 _severelyUndercollateralizedThresholdPercent,
         uint256 _timestamp
     );
-    event KeepFactorySingleShotUpdateStarted(
+    event KeepFactoryUpdateStarted(
```
Shadowfiend, line 41: This should become `KeepFactories...`, I think.

solidity/test/GovernanceTest.js
```diff
@@ -180,50 +180,91 @@ describe("TBTCSystem governance", async function() {
...
+    it("does not revert finalizeKeepFactoryUpdate if upgradeability period has passed", async () => {
+      await tbtcSystem.beginKeepFactoryUpdate(
         "0x0000000000000000000000000000000000000001",
         "0x0000000000000000000000000000000000000002",
       )
 
-      const finalizationTime = await tbtcSystem.getRemainingKeepFactorySingleShotUpdateTime()
-      await increaseTime(finalizationTime.toNumber() + 1) // 10 days
-      await tbtcSystem.finalizeKeepFactorySingleShotUpdate()
+      const upgradeabilityTime = await tbtcSystem.getRemainingKeepFactoriesUpgradeabilityTime()
```
Shadowfiend, line 227: > upgradeability

Lol this does not look like a word. What is wrong with you, English.

solidity/test/GovernanceTest.js
```diff
@@ -180,50 +180,91 @@ describe("TBTCSystem governance", async function() {
...
+    it("reverts for beginKeepFactoryUpdate if upgradeability period has passed", async () => {
+      await tbtcSystem.beginKeepFactoryUpdate(
+        "0x0000000000000000000000000000000000000001",
+        "0x0000000000000000000000000000000000000002",
+      )
+
+      const finalizationTime = await tbtcSystem.getRemainingKeepFactoryUpdateTime()
+      await increaseTime(finalizationTime.addn(1))
+
+      await tbtcSystem.finalizeKeepFactoryUpdate()
```
Shadowfiend, line 242: Why do we need to begin/finalize to verify this condition?

solidity/test/GovernanceTest.js
```diff
@@ -180,50 +180,91 @@ describe("TBTCSystem governance", async function() {
...
+      await increaseTime(upgradeabilityTime.addn(1))
 
       await expectRevert(
-        tbtcSystem.beginKeepFactorySingleShotUpdate(
+        tbtcSystem.beginKeepFactoryUpdate(
           "0x0000000000000000000000000000000000000001",
           "0x0000000000000000000000000000000000000002",
         ),
-        "Keep factory data can only be updated once",
+        "beginKeepFactoryUpdate can only be called within upgradeability period",
```
Shadowfiend, line 252: Let's be specific in this error message---`within 180 days of initialization`.

solidity/test/GovernanceTest.js
pdyraga: Donaudampfschifffahrtselektrizitätenhauptbetriebswerkbauunterbeamtengesellschaft. 🇩🇪

solidity/contracts/system/KeepFactorySelection.sol
nkuba: We have it combined on `TBTCSystem` level, we can do it also further on the `KeepFactorySelection` library level.

pdyraga:
👀

solidity/contracts/system/TBTCSystem.sol
nkuba: I'm glad you mentioned that, I was thinking about the same.

solidity/test/GovernanceTest.js
nkuba: > > upgradeability
> 
> Lol this does not look like a word. What is wrong with you, English.

For sure I won't be able to read it out loud.

solidity/contracts/system/TBTCSystem.sol
nkuba: 99fca0836140503a797bd509d913adf86dbdd33b

solidity/contracts/system/KeepFactorySelection.sol
pdyraga: I agree with @Shadowfiend and I'd combine it. It's simpler.

solidity/test/GovernanceTest.js
nkuba: 31137e5312e86deac996523b699db69b66291195

nkuba: 31137e5312e86deac996523b699db69b66291195

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -420,30 +427,32 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     /// @notice Finish setting the address of the ETH-only-backed ECDSA keep
     ///         factory and the selection strategy that will choose between it
     ///         and the default KEEP-backed factory for new deposits.
```
pdyraga, line 429: Docs are no longer fully up to date.

solidity/contracts/system/TBTCSystem.sol
```diff
@@ -79,19 +81,21 @@ contract TBTCSystem is Ownable, ITBTCSystem, DepositLog {
     uint64[] lotSizesSatoshis = [10**5, 10**6, 10**7, 2 * 10**7, 5 * 10**7, 10**8]; // [0.001, 0.01, 0.1, 0.2, 0.5, 1.0] BTC
 
     uint256 constant governanceTimeDelay = 48 hours;
+    uint256 constant keepFactoriesUpgradeabilityPeriod = 180 days;
```
pdyraga, line 84: Can't we have it as a public constant instead of exposing `getKeepFactoriesUpgradeabilityPeriod`?

solidity/test/GovernanceTest.js
```diff
@@ -180,50 +196,99 @@ describe("TBTCSystem governance", async function() {
...
+      await tbtcSystem.beginKeepFactoriesUpdate(
+        "0x0000000000000000000000000000000000000001",
+        "0x0000000000000000000000000000000000000002",
+        "0x0000000000000000000000000000000000000003",
+      )
+
+      const finalizationTime = await tbtcSystem.getRemainingKeepFactoriesUpdateTime()
+      await increaseTime(finalizationTime.addn(1))
+
+      await tbtcSystem.finalizeKeepFactoriesUpdate()
```
pdyraga, line 264: Why do we need those lines in this test? Isn't it enough to do what's below?

pdyraga:
Went through the code, agree with all @Shadowfiend's comments, and posted 3 more. Other than that, it looks good to me. Waiting for the updates! 🙌

pdyraga:
I am excited, it's much simpler than https://github.com/keep-network/tbtc/pull/722 and I am no longer worried about gas cost.

solidity/contracts/system/KeepFactorySelection.sol
nkuba: c09fb403

solidity/test/GovernanceTest.js
nkuba: b6b5ba5e72db3577cac613a43fa14d8d99e0f5e7

solidity/contracts/system/TBTCSystem.sol
nkuba: It's consistent with other getters, e.g. getGovernanceTimeDelay, getPriceFeedGovernanceTimeDelay.

nkuba: I can remove all getters and update constants to public, but I'm not sure if it was implemented like this on purpose.

pdyraga: Alright, fair enough.

solidity/contracts/system/TBTCSystem.sol
nkuba: 05986f4b

solidity/test/GovernanceTest.js
nkuba: I believe it was removed as in https://github.com/keep-network/tbtc/pull/724#discussion_r477294570

nkuba:
It's ready for a next round.

solidity/test/GovernanceTest.js
pdyraga: Oh, nice!

pdyraga (approved):
Looks good to me. Leaving the final pass and decision to @Shadowfiend.

solidity/test/KeepFactorySelectionTest.js
```diff
@@ -212,48 +270,192 @@ describe("KeepFactorySelection", async () => {
...
+      )
+
+      assert.include(
+        await keepFactorySelection.factories(),
+        {
+          _keepStakedFactory: newKeepStakedFactory.address,
+          _fullyBackedFactory: newFullyBackedFactory.address,
+          _factorySelector: newSelector.address,
+        },
+        "invalid factories after second update",
```
Shadowfiend, line 429: Pretty weird to swap to `assert` here. `expect(...).to.include(...)` does the same thing here while sticking to our preferred mocha expectation style, right? If not, we can keep it as is, just wanted to check.

solidity/test/GovernanceTest.js
```diff
@@ -180,50 +196,88 @@ describe("TBTCSystem governance", async function() {
...
+        "0x0000000000000000000000000000000000000006",
+        "0x0000000000000000000000000000000000000007",
+      )
+
+      expect(await tbtcSystem.getRemainingKeepFactoriesUpdateTime()).to.eq.BN(
+        await tbtcSystem.getGovernanceTimeDelay(),
+      )
+    })
+
+    it("does not revert if finalizeKeepFactoriesUpdate has already been called", async () => {
```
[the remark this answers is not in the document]
nkuba: `governanceTest` covers just a single update. Here we wanted to check if it's possible to perform multiple updates.

In `does not revert if beginKeepFactoriesUpdate has already been called` we wanted to confirm that a value can be overwritten for the ongoing update process before it's finalized.

In `does not revert if finalizeKeepFactoriesUpdate has already been called` we want to check that it's possible to update a value for a second time.

solidity/test/GovernanceTest.js
```diff
@@ -381,58 +435,137 @@ describe("TBTCSystem governance", async function() {
...
-        expect(await newKeepFactory.keepOwner.call()).to.equal(mockDeposit)
+        expect(await newFullyBackedFactory.keepOwner.call()).to.equal(
+          mockDeposit,
+        )
+      },
+    })
+
+    governanceTest({
+      property:
+        "keep factory update with zeroed fully backed factory and factory selector",
```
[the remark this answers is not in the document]
nkuba: Yes, I didn't like duplicating some parts of the test. I believe `goodParametersWithName` should handle multiple sets of parameters, similiar as `badInitializationTests` does.

I wasn't sure how to enhance the `governanceTest` mechanism in the most elegant way.

solidity/test/KeepFactorySelectionTest.js
nkuba: Totally agree! I was experimenting with some other verification functions from `assert` that are not supported by `expect`, but finally ended up with `include` that is available in `expect`. I'll update this according to the suggestion.

solidity/test/GovernanceTest.js
```diff
@@ -24,6 +26,13 @@ describe("TBTCSystem governance", async function() {
 
   const nonSystemOwner = accounts[3]
 
+  // We increased the value for new factories compared to the default ones
+  // as this is a way of validating that factory has been updated when using
+  // old (insufficient) value to request a new keep.
```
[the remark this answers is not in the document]
nkuba: Actually, when I think about it today it may not make sense...

solidity/test/GovernanceTest.js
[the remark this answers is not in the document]
Shadowfiend: Yeah that's totally fair. `governanceTest` is a bit of a beast heh.

solidity/test/GovernanceTest.js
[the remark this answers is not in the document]
Shadowfiend: `does not revert if beginKeepFactoriesUpdate has already been called` is covered by the `overrides previous update and resets timer` test in `governanceTest`.

It does look like we're missing the second one… We probably should test that in `governanceTest` in that case, since it's a shared property. But that's for another day.

Shadowfiend:
Something went awry with linting, not clear what happened. Going to give it an extra run.

solidity/test/GovernanceTest.js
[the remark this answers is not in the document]
nkuba: I covered them in https://github.com/keep-network/tbtc/pull/725

solidity/test/GovernanceTest.js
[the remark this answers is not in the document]
nkuba: I let myself to add a TODO comment so we don't lose it (5be382921b8edbe92fd2370ea6aa7ec7951a83cb).

PR #730, opened 2020-09-14
Add a reference to the latest Trail of Bits audit

mhluongo:
The report has been out for a bit — now we link to it :sunglasses:


Shadowfiend (approved):
Let's do it.

PR #731, opened 2020-09-15
Release: 1.1.0 (Mainnet)

pdyraga:

PR #732, opened 2020-09-15
Release: 1.1.0 (Mainnet)

Shadowfiend:
Rejigger of #731 with proper branch name and dropping the smallest lot size.

### DIFF
--- a/solidity/contracts/test/system/TBTCSystemStub.sol
+++ b/solidity/contracts/test/system/TBTCSystemStub.sol
@@ -12,6 +12,7 @@ contract TBTCSystemStub is TBTCSystem {
         TBTCSystem(_priceFeed, _relay)
     public {
         // solium-disable-previous-line no-empty-blocks
+        lotSizesSatoshis = [10**5, 10**6, 10**7, 2 * 10**7, 5 * 10**7, 10**8]; // [0.01, 0.1, 0.2, 0.5, 1.0] BTC
     }
 
     function setOraclePrice(uint256 _oraclePrice) external {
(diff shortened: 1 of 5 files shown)


Shadowfiend:
> revert Deposit too small -- Reason given: Deposit too small.

Haha okay, well that's what happens when your tests assume something about initial lot sizes. I'll fix this up as we do the deploy.

pdyraga (approved):

pdyraga:
Can I have your approval @Shadowfiend on the changes I made?

solidity/contracts/test/system/TBTCSystemStub.sol
```diff
@@ -12,6 +12,7 @@ contract TBTCSystemStub is TBTCSystem {
         TBTCSystem(_priceFeed, _relay)
     public {
         // solium-disable-previous-line no-empty-blocks
+        lotSizesSatoshis = [10**5, 10**6, 10**7, 2 * 10**7, 5 * 10**7, 10**8]; // [0.01, 0.1, 0.2, 0.5, 1.0] BTC
```
Shadowfiend, line 15: Incidentally, we may not need this.

solidity/test/helpers/fullDeployer.js
```diff
@@ -153,6 +153,10 @@ async function deployAndLinkAll(additions = [], substitutions = {}) {
     1,
   )
 
+  await tbtcSystem.beginLotSizesUpdate([100000, 1e8])
```
Shadowfiend, line 156: Might be worth dropping a comment on why we're doing this 🤔

Shadowfiend:
Since the two bits are test-related, let's take care of them in a separate PR.

PR #739, opened 2020-10-14
Add Go contract bindings

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-ecdsa/issues/574
Depends on: https://github.com/keep-network/keep-common/pull/52
Depends on: https://github.com/keep-network/keep-common/pull/53

Added a `Makefile` allowing to generate Go bindings for given contracts using the `go generate ./...` command. The solution is based on similar ones from [`keep-core`](https://github.com/keep-network/keep-core/blob/master/pkg/chain/gen/Makefile) and [`keep-ecdsa`](https://github.com/keep-network/keep-ecdsa/blob/master/pkg/chain/gen/Makefile) projects. This PR also ships several ready to use bindings prepared using the aforementioned `Makefile`.


pdyraga:
Thoughts on keeping all of it in `go` directory? We have `tbtc/solidity` so `tbtc/go` looks less cryptic to me than `tbtc/bindings`.

go/Makefile
```diff
@@ -0,0 +1,49 @@
+solidity_dir=$(realpath ${SOLIDITY_DIR})
+solidity_files := $(foreach file,${SOLIDITY_FILES},$(file))
```
pdyraga, line 2: Why don't we use the same mechanism as in `keep-core` and `keep-ecdsa`?
```
# Environment provides the solidity directory as a potentially-relative path,
# which we resolve. Then we resolve the Solidity files in a contracts/ directory
# at that path.
solidity_dir=$(realpath ${SOLIDITY_DIR})
solidity_files := $(wildcard ${solidity_dir}/contracts/*.sol)
```

go/Makefile
```diff
@@ -0,0 +1,49 @@
...
+
+# Solidity filenames without .sol with subdirectories prefixes.
+contract_stems := $(basename $(solidity_files))
+
+abi_files := $(addprefix abi/,$(addsuffix .abi,$(contract_stems)))
```
pdyraga, line 7: Thoughts on adding the same comment as in `keep-ecdsa` and in `keep-core`? It helps to understand what's happening here:
```
# Go abigen bindings in abi/ subdirectory with .go suffix, alongside solc ABI
# files with .abi suffix.
```

go/abi/DepositLog.go
```diff
@@ -0,0 +1,2406 @@
+// Code generated - DO NOT EDIT.
```
pdyraga, line 1: Can we keep all generated files inside `gen` directory (`go/gen`)? This is the pattern we use everywhere no matter what's generated (commands, bindings, promises).

go/generate.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+DIR=../solidity
+
+# Source contracts for binding generation.
+# They should include the subdirectory prefix if any.
+FILES="DepositLog.sol deposit/Deposit.sol"
+
+SOLIDITY_DIR=$DIR SOLIDITY_FILES=$FILES make
```
pdyraga, line 9: Can we avoid having this file and put it in `gen.go`? I believe this is the strategy we use for all the other cases so far.

go/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/keep-network/tbtc/go
+
+go 1.13
+
+require (
+	github.com/ethereum/go-ethereum v1.9.10
+	github.com/ipfs/go-log v1.0.4
+	github.com/keep-network/keep-common v1.2.1-0.20201015091820-496b0c87d6f5
+	github.com/keep-network/keep-core v1.3.0
```
pdyraga, line 9: If we don't generate commands, can we avoid having a dependency to `keep-core`? 

It looks... wrong. Although commands would be a nice addition we don't need them rigtht now.

go/Makefile
lukasz-zimnoch: Because `tbtc` contracts are grouped into subdirectories.

go/go.mod
lukasz-zimnoch: We can't disable command generation without changing the generator code in `keep-common`.

go/generate.sh
lukasz-zimnoch: I think having it as a separate file will be more readable once we will have more items in the `FILES` variable.

go/go.mod
pdyraga: We can try to update the generator to allow to ship a parameter to skip command generation. It feels better to me than requiring `tbtc` to depend on `keep-core`.

go/abi/DepositLog.go
lukasz-zimnoch: But in `keep-ecdsa` we have the same structure: https://github.com/keep-network/keep-ecdsa/tree/master/pkg/chain/gen

go/generate.sh
pdyraga: Agreed though until there are just two files, I'd keep it in `gen.go` just for consistency. Maybe we'll find some way in the future keep there more than two files in a clean way.

lukasz-zimnoch: Ok, I'll change!

go/abi/DepositLog.go
pdyraga: Right, and what I am talking about it this `/gen/` part :)

lukasz-zimnoch: Ah, got it now! I'll change it.

Shadowfiend:
> Thoughts on keeping all of it in go directory? We have tbtc/solidity so tbtc/go looks less cryptic to me than tbtc/bindings.

I know it's going to feel weird at first blush, but let's put these at the top level of the repo with a structure that matches what we're using elsewhere IMO. As in, `pkg/chain/ethereum/gen`.

pdyraga:
> > Thoughts on keeping all of it in go directory? We have tbtc/solidity so tbtc/go looks less cryptic to me than tbtc/bindings.
> 
> I know it's going to feel weird at first blush, but let's put these at the top level of the repo with a structure that matches what we're using elsewhere IMO. As in, `pkg/chain/ethereum/gen`.

Yeah, that's a very good idea. I guess we were thinking all the time that tbtc is Solidity-only and we were trying to reflect it in the directory structure by again, thinking that _we are adding some Go but not that mutch_. Nevertheless, it's no longer just a Solidity repository, and having the same structure everywhere is a good move. Especially that we already have Kube infrastructure files here.

go/go.mod
lukasz-zimnoch: Ok, but can we do it separately? It's a non-trivial change and I don't want to block here now.

pdyraga: We can always extract bindings to a separate PR - contracts are not changing so we don't need to regenerate them daily - and do generation here, once we solve the problem with commands. It is really hard to anticipate what is a consequence of adding `keep-core` dependency from `tbtc` to our deployment lifecycle, even if it is temporary.

Worth noting, maybe it's not about disabling command generation but about moving some `keep-core` code out to `keep-common`?

lukasz-zimnoch: Ok, I have done the change about optional commands.

lukasz-zimnoch:
@pdyraga everything is addressed here. I'm ready for a next look!

.gitignore
```diff
@@ -16,6 +16,7 @@ __pycache__/
 node_modules/
 artifacts/
 .mypy_cache
+pkg/chain/ethereum/gen/abi/**/*.abi
```
pdyraga, line 19: We should have it as `*.abi`. With the current version, I have the files from `go/abi` in git diff.

pdyraga (approved):

PR #740, opened 2020-10-23
Add TBTCSystem event log

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-ecdsa/issues/574

Added a `TBTCSystemEventLog` allowing to access past events emitted by the `TBTCSystem` contract. This is a temporary workaround because contract bindings generated by the `keep-common` generator don't support this feature yet.

### DIFF
--- a/pkg/chain/ethereum/gen/eventlog/TBTCSystemEventLog.go
+++ b/pkg/chain/ethereum/gen/eventlog/TBTCSystemEventLog.go
@@ -1,4 +1,4 @@
-package filterer
+package eventlog
 
 import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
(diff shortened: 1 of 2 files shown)


DONALDSULLENDER (approved):

pdyraga:
Holding off with this one until [discussion in a PR using this functionality](https://github.com/keep-network/keep-ecdsa/pull/585#discussion_r513316679) is resolved.

pdyraga:
Merging to do not block application-specific actions implementations in the client (https://github.com/keep-network/keep-ecdsa/pull/585). This is not a final solution, and we'll have to update `keep-common` generators to generate the code from this PR, as [we discussed here](https://github.com/keep-network/keep-ecdsa/pull/585#discussion_r513316679).

pdyraga (approved):

PR #743, opened 2020-11-10
Clean node_modules before npm i

dimpar:
We experience CI build failures when running e2e tests. Some of the packages in `node_modules` crash the tests. It is safer to clean the `node_modules` first and then `npm i`.


nkuba:
I'm not sure about his change. If there is a problem for E2E tests we should clean the `node_modules` directory as part of e2e setup script.

dimpar:
okay, moved the change to `local-setup` in this PR https://github.com/keep-network/local-setup/pull/53

PR #751, opened 2021-02-01
Modifying install.sh script to handle Celo contracts deployment

dimpar:
This PR adds a functionality where we can specify properties that we need in order to be connected to Celo Alfajores testnet. Local network setup should remain untouched and work as it worked before.

Example usage:
```
./scripts/install.sh --keep-ecdsa-path /Users/dp/go/src/github.com/keep-network/keep-ecdsa-celo/keep-ecdsa --network alfajores --private-key 0xf84e15953c7478f0853d03d4d580d2317f42c55fda572ff986fd70b3c4e3eae1 
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,12 +1,11 @@
 #!/bin/bash
-set -e pipefail
+set -euo pipefail
 
 # Dafault inputs.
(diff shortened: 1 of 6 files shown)


dimpar:
Temp closing this PR. Will need to modify a couple of things before it's ready.

PR #753, opened 2021-02-18
Create GH Actions workflow for lint of Solidity contracts

michalinacienciala:
and tBTC release processes in order to reduce human involvement and make
the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions. This commit creates a
GitHub Action workflow for linting of Solidity contracts.


.github/workflows/contracts.yml
```diff
@@ -0,0 +1,54 @@
...
+      - ".github/workflows/contracts.yml"
+  pull_request:
+    branches:
+      # TODO: Run on all branches after we're fully migrated from Circle CI
+      - "rfc-18/**"
+    paths:
+      - "solidity/**"
+      - ".github/workflows/contracts.yml"
+
+
```
nkuba, line 21: Let's remove one of the duplicated empty lines.

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          path: ~/.npm # npm cache files are stored in `~/.npm` on Linux/macOS
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        working-directory: ./solidity
```
nkuba, line 45: Could we try using [default working-directory setting](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun) so we won't have to define it for each step?

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: npm ci
+
+      - name: Compile contracts
+        working-directory: ./solidity
+        run: npx truffle compile
```
nkuba, line 50: We've got a script defined in `package.json` for that.
```suggestion
        run: npm run compile
```

nkuba (changes requested):

.github/workflows/contracts.yml
michalinacienciala: I've spotted it later and fixed that while working on #755, so it will be fixed once that PR gets accepted.

.github/workflows/contracts.yml
```diff
@@ -0,0 +1,54 @@
...
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Compile contracts
+        run: npm run compile
```
nkuba, line 51: We don't need this step here in `lint` job. This will be covered by `build and test` job, right?

nkuba (changes requested):
Just one minor comment.

.github/workflows/contracts.yml
michalinacienciala: You're right, I removed it.

nkuba (approved):

PR #755, opened 2021-02-19
Create GH Actions workflow for build&test of Solidity contracts

michalinacienciala:
and tBTC release processes in order to reduce human involvement and make
the process less error prone. A part of this task is migration from
CircleCI to GitHub Actions. This commit creates a GitHub Action workflow
for building and testing of Solidity contracts.


.github/workflows/contracts.yml
```diff
@@ -18,8 +18,40 @@ on:
...
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        working-directory: ./solidity
+        run: npm ci
+
+      - name: Compile contracts
+        working-directory: ./solidity
+        run: npx truffle compile
```
nkuba, line 49:
```suggestion
        run: npm run compile
```

nkuba (changes requested):

nkuba (approved):

PR #758, opened 2021-03-03
Celo contract bindings

lukasz-zimnoch:
Here we introduce Go contract bindings intended to use with the Celo blockchain. All files are placed under the `pkg/chain/celo` package in the same way as in the existing `pkg/chain/ethereum` package.

### DIFF
--- a/pkg/chain/celo/gen/abi/deposit/Deposit.go
+++ b/pkg/chain/celo/gen/abi/deposit/Deposit.go
@@ -1,0 +1,1137 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package abi
+
+import (
+	"math/big"
+	"strings"
+
+	ethereum "github.com/celo-org/celo-blockchain"
+	"github.com/celo-org/celo-blockchain/accounts/abi"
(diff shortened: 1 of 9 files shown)


pkg/chain/celo/gen/abi/deposit/Deposit.go
```diff
@@ -0,0 +1,1137 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package abi
+
+import (
+	"math/big"
+	"strings"
+
+	ethereum "github.com/celo-org/celo-blockchain"
```
pdyraga, line 10: Interesting... is it something that wasn't renamed?

lukasz-zimnoch: That's done by the Celo `abigen`. This is because Celo team renamed the package `celo-org/celo-blockchain` from `ethereum` to `celo` but they still alias them as `ethereum` to integrate upstream changes in an easy way. Details here: https://github.com/celo-org/celo-blockchain/pull/1375#discussion_r574786037

pdyraga (approved):

PR #760, opened 2021-03-11
EIP155 signer support in Go contract bindings

pdyraga:
go-ethereum 1.10.0 has switched to denying non-EIP155 signatures by default. Although keep-common code uses go-ethereum own ABI binding generation, the version currently in the dependencies always uses Homestead signers. The keep-common version updated in this PR ships EIP155-compatible signer without updating the referenced go-ethereum version. The regenerated `Deposit` and `TBTCSystem` Go bindings contract code includes now that signer.

See https://github.com/keep-network/keep-common/pull/71

### DIFF
--- a/pkg/chain/ethereum/gen/contract/Deposit.go
+++ b/pkg/chain/ethereum/gen/contract/Deposit.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 
 	"github.com/ipfs/go-log"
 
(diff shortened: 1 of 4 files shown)


nkuba (approved):

pkg/chain/ethereum/gen/contract/Deposit.go
```diff
@@ -96,12 +114,12 @@ func NewDeposit(
 // ----- Non-const Methods ------
 
 // Transaction submission.
-func (d *Deposit) NotifyRedemptionProofTimedOut(
+func (d *Deposit) NotifyCourtesyCall(
```
dimpar, line 117: I've navigated to pkg/chain/ethereum/gen and `make clean`. Then in /solidity `npm i`and finally `go generate ./...` `NotifyCourtesyCall` was actually named `ProvideRedemptionSignature`. How do we chose the name?

dimpar: ah, nvm, it's not sync generated

dimpar (approved):

pkg/chain/ethereum/gen/contract/Deposit.go
pdyraga: Those functions are generated in a random order, this was fixed in https://github.com/keep-network/keep-common/pull/68 but the cutoff commit from where we branched off this fix does not include the referenced change.

PR #762, opened 2021-03-15
Initial structure for Go relay maintainer

lukasz-zimnoch:
Here we introduce an initial structure of the Go relay maintainer. The proposed structure draws from existing Go projects within Keep organization and borrows some ideas and parts of code from them.


.pre-commit-config.yaml
```diff
@@ -1,3 +1,16 @@
...
+   hooks:
+    - id: go-imports
+    - id: go-lint
+ - repo: https://github.com/tekwizely/pre-commit-golang.git
+   rev: v0.8.2
+   hooks:
+     # The `go-vet` hook is taken from the `tekwizely/pre-commit-golang.git`
+     # repo because `go-vet` from `keep-network/pre-commit-golang.git` doesn't
+     # work if Go files don't exist at the project root level.
+     - id: go-vet-pkg
```
pdyraga, line 13: Non-blocking but is there an easy way to tweak `keep-network/pre-commit-golang.git` in the future and replace this part?

relay/cmd/start.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return fmt.Errorf("could not connect BTC chain: [%v]", err)
+	}
+
+	hostChain, err := connectHostChain(ctx, config)
+	if err != nil {
+		return fmt.Errorf("could not connect host chain: [%v]", err)
+	}
+
+	err = forwarder.Initialize(btcChain, hostChain)
```
pdyraga, line 52: `forwareder` or `node`? This is relay node.

relay/pkg/btc/btc.go
```diff
@@ -0,0 +1,42 @@
...
+	// TODO: implementation
+}
+
+// Client exposes methods needed to interact with the BTC blockchain.
+// It's defined as an interface in order to allow swapping the actual
+// implementation. For example, during the integration tests, it can
+// be more convenient to use a mock instead of the real network.
+type Client interface {
+	// TODO: implementation
+}
```
pdyraga, line 22: Not sure if I follow what the `Client` interface will be used for. Do we have any analogy in other keep clients?

lukasz-zimnoch: No, we don't have an analogy. The idea is to have a possibility to swap actual client implementation within `btc.Chain`. For example, for regular use, it can be `rpc.Client` which actually connects to the BTC network. On the other hand, a `local.Client` can be used as a mock blockchain client while running a simple integration test.

.pre-commit-config.yaml
lukasz-zimnoch: Maybe, but I have not focused on that for now in order to not block here.

relay/pkg/btc/btc.go
lukasz-zimnoch: For clarification, I have not created `btc.Chain` as an interface because in contrary to the `pkg/chain` host chains we will always have BTC as the source chain. The only thing which may change here is whether we use an actual RPC client or a mocked one for testing purposes. This is why I think the `btc.Client` abstraction is right in that context.

relay/cmd/start.go
lukasz-zimnoch: For now, it's just a forwarder and anything else. During development, we will probably introduce a `node` layer to handle relay modes and stuff. I just wanted to avoid running too far in the future at the moment.

relay/pkg/btc/btc.go
pdyraga: I'd rather stick with `bitcoin.Chain` interface TBH. Bitcoin is a chain and we'll be extracting information from the chain and pushing it to a host chain. `bitcoin.Chain` may consist of multiple interfaces and pass those individual scoped interfaces to functions - we can then implement tests easier and have functions scoped to this part of the interface they really need. 

Client feels like a low-level concept to me that lives under the chain interface and is responsible for the communication with an external node, that's also how Ethereum and Celo implementations work.

We can have `Connect() bitcon.Chain` and `ConnectLocal() bitcoin.Chain`, we do not expect more than two options here.

relay/cmd/start.go
pdyraga: I'd suggest doing `note.Initialize` to keep it aligned with other projects. It is much easier to navigate the structure this way. And inside `node.Initialize` we can have a call starting a forwarder loop that is implemented in `forwarder` package.

relay/pkg/btc/btc.go
lukasz-zimnoch: Ok, done in https://github.com/keep-network/tbtc/commit/37d083999a579af04bff35ae6cafb20f6ac46ebc

.pre-commit-config.yaml
lukasz-zimnoch: For now, I will remove the `go-vet` hook completely as it still doesn't work well and cause problems on each commit. I'll add a backlog task for that and return here later.

relay/cmd/start.go
lukasz-zimnoch: Ok, done in https://github.com/keep-network/tbtc/commit/a5c9973b416081f84f2cf58e214aa9da6e9c37a4

relay/pkg/btc/remote/remote.go
```diff
@@ -0,0 +1,29 @@
...
+
+var logger = log.Logger("relay-btc-remote")
+
+// remoteChain represents a remote Bitcoin chain.
+type remoteChain struct {
+	// TODO: implementation
+}
+
+// Connect connects to the Bitcoin chain and returns a chain handle.
+func Connect() (btc.Handle, error) {
```
pdyraga, line 18: I'd consider exposing `bitcoin.Connect()` forwarding to this (then unexported) function. `remote.Connect()` is confusing, `bitcoin.Connect()` would look much better. And we can alias `bitcoin/local` as `btcLocal` for `btcLocal.Connect()` in tests.

pdyraga (approved):
Leaving this PR in @tomaszslabon's hands.

relay/pkg/btc/remote/remote.go
```diff
@@ -0,0 +1,29 @@
...
+	// TODO: implementation
+	logger.Infof("connecting remote Bitcoin chain")
+
+	return &remoteChain{}, nil
+}
+
+// GetHeaderByHeight returns the block header for the given block height.
+func (rc *remoteChain) GetHeaderByHeight(height *big.Int) *btc.Header {
```
tomaszslabon, line 26: Minor remark:
In the bitcoin Go implementation they simply use int32 for block height:
"blockHeight              int32           // Height in the main block chain"

I think we could do the same (use int32 or int64), it would be easier to use. 
Unless using big.Int is some convention.

lukasz-zimnoch: Please treat that `GetHeaderByHeight` method just as a starting point of the `btc.Handle` interface. We can change everything during the actual BTC connectivity implementation. If `int64` will fit better here I have nothing against it.

relay/pkg/btc/remote/remote.go
lukasz-zimnoch: Sure we'll consider that during the actual implementations of chains connectivity. For now, it's just a starting point.

pdyraga: On the other hand... import cycle. But yeah, let's see in the future.

tomaszslabon (approved):

PR #766, opened 2021-03-16
Move migrate&publish of Eth contracts to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvement and to
migration from CircleCI jobs to GitHub Actions. This PR enhances
the workflow for building and testing Solidity with a job performing
contracts migration on Ethereum chain.
The job does following:
- fetches the contracts artifacts uploaded to Google Storage bucket by
  Solidity workflow in`keep-ecdsa`
- migrates contracts
- pushes them to Tenderly
- uploads contract data to Google Storage bucket (`tbtc`)
- bumps up version of npm package containing contract artifacts and
  uploads the package to npm registry

The `migrate-and-publish-ethereum` job is configured to execute only if
workflow is started by a push or workflow_dispatch event on master or
`rfc-18/*` branch. The `rfc-18/*` condition will need to be removed
once whole `build-test-migrate-publish-keep-test` CircleCI process gets
migrated to GH Actions.
Additionally, as we may not always want to migrate and publish
contracts on every push (or workflow_dispatch) on master, the job was
configured to require manual approval in GitHub GUI. This is done by
configuration of environment protection rules on keep-test and using
`environment: keep-test` config in the workflow.


michalinacienciala:
Ready for review. Here is the link to passing workflow: https://github.com/keep-network/tbtc/actions/runs/664528079. There is one failure in `Push contracts to Tenderly` step (`TBTCConstants` contract could not be pushed due to an error), but I believe this is a matter that needs to be addressed outside of this PR.
Currently workflow performs dry-run of push to npm registry. `--dry-run` flag will need to be removed in "cleanup" PR.

nkuba (approved):

PR #765, opened 2021-03-16
Add ETH connectivity to Go relay maintainer

lukasz-zimnoch:
Depends on: #763 

Here we introduce the ETH connectivity logic for the Go relay maintainer. Specific changes involve:
- Adding required `Relay` contract methods to the `chain.Handle` interface
- Implementing those methods in `ethereumChain` and adding stubs in the local implementation.
- Connecting to Ethereum chain upon `start` command
- Adding migration config which allows to easily deploy appropriate Summa's Relay contract 
- Configuring binding generator for Summa's Relay contract
- Updating the `Dockerfile` and adding generator step
- Adding `install.sh` and `start.sh` scripts for convenience
- Some minor improvements, README update, etc.

PR #767, opened 2021-03-18
Relay maintainer: Push blocks happy path

lukasz-zimnoch:
Depends on: #765

Here we introduce the block forwarder pushing loop which supports a basic case of pushing a batch of headers within which the chain difficulty doesn't change. Specifically, this pull request:
- Adds a general structure of the block forwarder
- Introduces forwarder lifecycle logic and error handling
- Implements pulling block headers from the internal headers queue 
- Adds pushing block headers to the host chain in case there is no difficulty change 
- Implements updating best header if needed
- Performs some steps to improve the work with raw bytes by introducing `btc.Digest` type
- Improves documentation of chain interfaces


relay/pkg/chain/chain.go
```diff
@@ -10,44 +10,45 @@ type Handle interface {
 // Relay is an interface that provides ability to interact with Relay contract.
 type Relay interface {
 	// GetBestKnownDigest returns the best known digest.
-	GetBestKnownDigest() ([32]uint8, error)
+	GetBestKnownDigest() ([32]byte, error)
```
pdyraga, line 13: Should we declare `type Digest [32]byte` just like we do for BTC? There is a lot of `[32]byte` declarations in this file.

relay/pkg/node/node.go
```diff
@@ -1,20 +1,56 @@
...
 	"github.com/keep-network/tbtc/relay/pkg/chain"
-	"github.com/keep-network/tbtc/relay/pkg/forwarder"
 )
 
+var logger = log.Logger("relay-node")
+
 // Initialize initializes the relay node.
+//
+// TODO: This function will be probably the right place to handle new
+//  requirements which will require starting and stopping the block forwarder.
```
pdyraga, line 17: Agreed. 😆  Though this is already happening, it's not a TODO?

relay/pkg/node/node.go
```diff
@@ -1,20 +1,56 @@
...
+
+		forwarder := block.RunForwarder(ctx, btcChain, hostChain)
+
+		select {
+		case err := <-forwarder.ErrChan():
+			logger.Errorf(
+				"forwarder terminated with error: [%v] "+
+					"and will be restarted immediately",
+				err,
+			)
```
pdyraga, line 50: Maybe just `"relay occurred an error: [%v]"`? It will be clear from the logs it is continuing its work. "Restart" feels a little too rough. As an operator, I'd expect the entire process is restarted/

relay/pkg/btc/remote/remote.go
```diff
@@ -97,9 +97,9 @@ func (rc *remoteChain) GetHeaderByHeight(height *big.Int) (*btc.Header, error) {
 	}
```
pdyraga, line 97: I have already noted it and I know I am annoying but `btc/remote/remote.go` and `btc/local/local.go` feels wrong to me especially that they are exported. We can, for example, do `btc/client.go` and `btc/local_client.go` and have two functions, `Connect()` and `ConnectLocal()`

pdyraga: Not in this PR of course, just grumbling. 😁

relay/pkg/node/node.go
```diff
@@ -1,20 +1,56 @@
...
+		case <-ctx.Done():
+			logger.Infof("forwarder control loop context is done")
```
pdyraga, line 52: But what does it mean to the operator? Maybe just `"stopping the relay"`?

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+	return forwarder
+}
+
+func (f *Forwarder) pushingLoop(ctx context.Context) {
+	logger.Infof("running forwarder pushing loop")
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Infof("forwarder context is done")
```
pdyraga, line 72: I don't think we need it, especially that it is already logged one level higher. We can log it here or in the code calling `RunForwarder` but let's do it only one time.

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+			}
+
+			logger.Infof(
+				"pushing [%v] header(s) to host chain",
+				len(headers),
+			)
+
+			if err := f.pushHeadersToHostChain(ctx, headers); err != nil {
+				f.errChan <- fmt.Errorf("could not push headers: [%v]", err)
+				return
```
pdyraga, line 89: Instead of pushing the error to the channel and hoping that the control loop will restart this process, why don't we just log an error and continue?

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+			}
+
+			logger.Infof(
+				"suspending forwarder pushing loop for [%v]",
+				forwarderPushingSleepTime,
+			)
```
pdyraga, line 95: I'd rather have `pullHeadersFromQueue` blocking until it has as least one header to return. Feels more elegant.

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+	// Duration for which the forwarder should rest after performing
+	// a push action.
+	forwarderPushingSleepTime = 45 * time.Second
+)
+
+var logger = log.Logger("relay-block-forwarder")
+
+// Forwarder takes blocks from the Bitcoin chain and forwards them to the
+// given host chain.
+type Forwarder struct {
```
pdyraga, line 36: Thoughts on renaming it to `Relay`?

relay/pkg/block/forwarder.go
pdyraga, line 89: Let's stay that `pushHeadersToHostChain` failed. If we `return` and restart this loop or just log an error and `continue` without doing anything else, `headers` will not be pushed and we'll be waiting for another batch of headers from `pullHeadersFromQueue`. Is it blocking the maintainer on-chain? Can we afford to drop one batch of headers?

relay/pkg/block/push.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/keep-network/tbtc/relay/pkg/btc"
+)
+
+// pullHeadersFromQueue waits until we have `headersBatchSize` headers from
+// the queue or until the queue fails to yield a header for
+// `headerTimeout` duration.
```
pdyraga, line 14: `... and returns them from the function`?

relay/pkg/block/push.go
```diff
@@ -0,0 +1,214 @@
...
+func (f *Forwarder) pullHeadersFromQueue(ctx context.Context) []*btc.Header {
```
pdyraga, line 15: Thoughts on naming it `pullHeaders` or `pullBtcHeaders`? The fact we take them from `f.headersQueue` is an implementation details. From the perspective of the code calling this function, we are pulling them from Bitcoin chain.

relay/pkg/block/forwarder.go
pdyraga: Wait, this is already happening. Why do we need to suspend the forwarder? Wouldn't it be enough to increase the timeout?

relay/pkg/block/push.go
lukasz-zimnoch: I'm rather against it because within the forwarder we pull block either from the BTC chain or from the queue and naming it just `pullHeaders` will be ambiguous. I think it's more readable and consistent if we have on the push side:
- `pullHeadersFromQueue`
- `pushHeadersToHostChain`
and on the pull side (see https://github.com/keep-network/tbtc/pull/769):
- `pullHeadersFromBtcChain`
- `pushHeadersToQueue`

Last but not least, those methods are package-private helpers, the client code just do `block.RunForwarder` and doesn't care.

relay/pkg/chain/chain.go
lukasz-zimnoch: I thought about this but this will require conversions between `btc.Digest` and `chain.Digest` in the `block.Forwarder`. That seemed odd to me as `block.Forwarder` shouldn't care about data conversions but just forward headers according to defined logic. Of course, this is a simple conversion but still, I'm not 100% sure.

relay/pkg/node/node.go
lukasz-zimnoch: It's not done as we don't support new features yet. What we have done is just forwarder which can be disabled using contexts and is restarted upon failures.

relay/pkg/btc/remote/remote.go
lukasz-zimnoch: Yeah, that's probably a good move. I'll leave a TODO and address this in a follow-up PR.

relay/pkg/block/forwarder.go
lukasz-zimnoch: Because we can enter into an inconsistent state. It's far better to fail here, restart the forwarder, initialize it with some chain data (it's needed by the pull side), and start from the beginning. This is also the way it was handled in the legacy relay maintainer but there the entire app crashed. Here we just restart the forwarding part.

relay/pkg/block/forwarder.go
lukasz-zimnoch: I think there is no sense in continuously looping and checking timeouts on the queue in case there is no new headers. Also, if there are more blocks on the queue it's more efficient to handle them in batch than pulling them one by one. This is also how the legacy maintainer works.

relay/pkg/block/forwarder.go
lukasz-zimnoch: I'm not sure as the entire subdirectory is named `Relay`, we have a `Relay` contract and we have `chain.Relay` 😁  I think `Forwader` is a better name for things we are doing here. I always considered `Relay` to be a high-level concept which consists of BTC and ETH connectivity, block forwarder, and auxiliary stuff.

relay/pkg/block/forwarder.go
lukasz-zimnoch: It's even better to `return`, restart, and drop the batch entirely. In that case, the new forwarder instance will initialize the pull side properly basing on `Relay` contract state. If we just log and continue, we will receive the next batch from the queue which cannot be pushed on the host chain as the chain still "waits" for the old batch. The forwarder will remain in an inconsistent state.

relay/pkg/node/node.go
lukasz-zimnoch: Agree however because of https://github.com/keep-network/tbtc/pull/767#discussion_r598995537 my proposition is: https://github.com/keep-network/tbtc/commit/c6fb5d71e9c976fb98ee676e36dfcd48ff6221fe

relay/pkg/node/node.go
lukasz-zimnoch: Agree however because of https://github.com/keep-network/tbtc/pull/767#discussion_r598995537 my proposition is: https://github.com/keep-network/tbtc/commit/c6fb5d71e9c976fb98ee676e36dfcd48ff6221fe

relay/pkg/block/forwarder.go
lukasz-zimnoch: Done in https://github.com/keep-network/tbtc/commit/c6fb5d71e9c976fb98ee676e36dfcd48ff6221fe

relay/pkg/block/push.go
lukasz-zimnoch: Please see https://github.com/keep-network/tbtc/commit/c6fb5d71e9c976fb98ee676e36dfcd48ff6221fe

lukasz-zimnoch:
@pdyraga ready for another look!

relay/pkg/btc/remote/remote.go
lukasz-zimnoch: Follow-up PR: https://github.com/keep-network/tbtc/pull/770

relay/pkg/chain/chain.go
pdyraga: I think this conversion feels wrong because we would be converting from BTC header to... BTC header but declared in the host chain package. Instead of having two (identical) representations of BTC header, maybe the interface from `pkg/chain/chain.go` should import types from `pkg/btc/btc.go`?

tomaszslabon (approved):

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			logger.Infof("pulling new headers from queue")
+
+			headers := f.pullHeadersFromQueue(ctx)
+			if len(headers) == 0 {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Please see https://github.com/keep-network/tbtc/commit/425f216ae8171d841a9791b6e752581258481412

relay/pkg/node/node.go
pdyraga: Ah, I see - we are talking about relay auctions. Yes, it makes sense. Let's make it clear in the TODO what we are talking about. Proposition:
```
// TODO: This function will be probably the right place to handle relay auctions
// which will require starting and stopping the block forwarder.
```

relay/pkg/block/forwarder.go
pdyraga: I think this is fine, it is the same situation as with `keep-core/pkg/beacon` or `keep-ecdsa/pkg/ecdsa` where the "heart" package of the client is named after the protocol. There is much more logic happening inside this package than just forwarding: `updateBestHeader`, `findLastCommonAncestor` - this code is about how the relay works.

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Block duration of a Bitcoin difficulty epoch.
+	difficultyEpochDuration = 2016
+
+	// Duration for which the forwarder should rest after performing
+	// a push action.
+	forwarderPushingSleepTime = 45 * time.Second
+)
+
+var logger = log.Logger("relay-block-forwarder")
```
pdyraga, line 32: I'd go with `tbtc-relay` given that the repository is `tbtc`. We have `keep-beacon` logger for `keep-core/pkg/beacon` and `keep-tss` logger for `keep-ecdsa/pkg/ecdsa/tss`.

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,138 @@
...
+	headersQueue chan *btc.Header
+	errChan      chan error
+
+	loopExitHandler func()
+}
+
+// RunForwarder creates an instance of the block forwarder and runs its
+// processing loops. The lifecycle of the forwarder can be managed using the
+// passed context.
+func RunForwarder(
```
pdyraga, line 51: If we decide to rename the package to `relay`, I propose we name it `Start` and make it clear in the docs that the relay exits on a first error. This will explain the logic in `runForwarderControlLoop` in `node.go`.

pdyraga (commented):
The first portion, more to come depending on their outcome.

relay/pkg/block/forwarder.go
lukasz-zimnoch: I think it will be too long. We will have `tbtc-relay-block-forwarder`. The `tbtc` prefix doesn't give any value in my opinion. No need to know we are in the `tbtc` repo from the `relay`'s perspective.

pdyraga: Per another comment, I'd really love to rename it to "relay" and then it's just `tbtc-relay`.

lukasz-zimnoch: So, maybe just `relay` then. Still not getting the point of forcing `tbtc` prefix here 😁

pdyraga: To keep it consistent with keep client logging where all logs are prefixed with `keep-*` and it may help with filtering logs.
I was even thinking about `keep-relay` but the relay is a part of tBTC, not Keep Network.

lukasz-zimnoch: `tbtc` repo is part of `keep-network` organization so `keep-relay` is not that bad. But ok, `tbtc-relay` may also be the option.

relay/pkg/chain/chain.go
lukasz-zimnoch: Yeah, importing types from `btc.go` makes sense. However I will do this in a follow-up PR once the whole stack is merged. Left a TODO for now: https://github.com/keep-network/tbtc/commit/624b6750d90127716326f460c7c934a16f9125de

relay/pkg/node/node.go
lukasz-zimnoch: See https://github.com/keep-network/tbtc/commit/551cf7c8e780374c860a4edc9c286579705867f7

relay/pkg/chain/chain.go
pdyraga: Yeah, separate PR makes sense :+1:

relay/pkg/block/forwarder.go
lukasz-zimnoch: Technically, `updateBestHeader` and `findLastCommonAncestor` are parts of the forwarding logic as well. Also, If we rename `block.Forwarder` to `relay.Relay` then according to your comment, the future auction stuff should be also placed in `relay` package as it's how the relay will work. This seems wrong to me. There is also the thing that `relay.Relay` will use `chain.Relay` which seems weird. Finally, I think `keep-core` and `keep-ecdsa` cases are not the same. The `beacon` package is something different than the whole repo name. Same in `keep-ecdsa` where the core logic is placed in `ecdsa/tss`.

pdyraga: > Technically, updateBestHeader and findLastCommonAncestor are parts of the forwarding logic as well.

Right, and that's one of the things that are IMO confusing because it is not just forwarding from one place to another (just like HTTP forward for instance).

> Also, If we rename block.Forwarder to relay.Relay then according to your comment, the future auction stuff should be also placed in relay package as it's how the relay will work

No, the actions should be placed in `node.go` (or in `auctions` package referenced from `node.go`) and depending on the result of the auction, the client will start the relay or not.

> There is also the thing that relay.Relay will use chain.Relay which seems weird.

IMO it's consistent - we have a relay off-chain logic that is using relay on-chain implementation. `chain.Relay` is just an on-chain part of the interface.

> Finally, I think keep-core and keep-ecdsa cases are not the same. The beacon package is something different than the whole repo name. Same in keep-ecdsa where the core logic is placed in ecdsa/tss.

For the beacon, it's a kind of a historical decision that "core" is beacon + grant and staking contracts. If we had a separate repository for grant and staking contracts, the repository with the client would be named `keep-beacon` (this may still happen one day). For `keep-ecdsa`, we have a separate `tss` directory because we use third party implementation, but if we were using our own, it would live in `ecdsa`.

My order of preferences is:
- `pkg/relay` with `relay.go` and `Start()`
- `pkg/headers` with `headers.go` and  `StartRelay()`or `Transmit()`
- `pkg/headers` with `headers.go` and `RunForwarder()`

I don't think `pkg/block` is a good choice given that we work on headers and I don't think `forwarder.go` is a good choice because we do much more than just forwarding from one place to another.

relay/pkg/block/push.go
```diff
@@ -0,0 +1,214 @@
...
+	"math/big"
+	"time"
+
+	"github.com/keep-network/tbtc/relay/pkg/btc"
+)
+
+// pullHeadersFromQueue waits until we have `headersBatchSize` headers from
+// the queue or until the queue fails to yield a header for
+// `headerTimeout` duration and returns them from the function.
+func (f *Forwarder) pullHeadersFromQueue(ctx context.Context) []*btc.Header {
```
pdyraga, line 15: Can we separate pull functions to `pull.go`? I kinda like that we have a separate file for push logic instead of keeping it all in one big file but having a separation of push+pull+glue code would be even more awesome.

relay/pkg/block/forwarder.go
lukasz-zimnoch: Yeah, I agree with the remark about `pkg/block`, this was a starting point which now can be narrowed to `pkg/header`. In that case, we can make `header.Relay` placed in `pkg/header/relay.go` exposing a `StartRelay` function. This is a small mix of your propositions but I think it's the best reflection of what we are doing here. Of course, I propose to do this as a follow-up PR after the entire stack lands on master. I'll leave a TODO. WDYT?

relay/pkg/block/push.go
lukasz-zimnoch: `pull.go` is about pulling blocks from BTC chain and pushing them to queue (see https://github.com/keep-network/tbtc/pull/769). `push.go` is about pulling blocks from queue and pushing to the host chain. If we move `pullHeadersFromQueue` to `pull.go` we will make a small mess here.

relay/pkg/block/forwarder.go
pdyraga: Yeah, that's the best outcome we could get here IMO when I look at it again 👌 . Let's leave a TODO.

lukasz-zimnoch: Here it is: https://github.com/keep-network/tbtc/commit/bce322f8b5881c410789c9c5f7a885815dbcc5f2

relay/pkg/block/forwarder.go
lukasz-zimnoch: We'll make it `StartRelay` as per https://github.com/keep-network/tbtc/pull/767#discussion_r603913978

relay/pkg/block/forwarder.go
pdyraga: Resolving, this will be done in a separate PR according to the TODO: bce322f8b5881c410789c9c5f7a885815dbcc5f2

relay/pkg/node/node.go
lukasz-zimnoch: As stated in https://github.com/keep-network/tbtc/pull/767#discussion_r603913978, the message will be focused around `headers` not `blocks`

relay/pkg/node/node.go
lukasz-zimnoch: As stated in https://github.com/keep-network/tbtc/pull/767#discussion_r603913978, the message will be focused around `headers` not `blocks`

relay/pkg/block/forwarder.go
pdyraga: I have totally missed that forwarder is stateful. Can we leave a comment with an explanation? My proposition:
```
				f.errChan <- fmt.Errorf("could not push headers: [%v]", err)
				// We exit on the first error letting the code controlling the
				// relay to restart it. The relay is stateful and it is easier
				// to fetch the most recent information from BTC after the
				// restart instead of trying to recover here.
				return
```

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,143 @@
...
+	ctx context.Context,
+	btcChain btc.Handle,
+	hostChain chain.Handle,
+) *Forwarder {
+	loopCtx, cancelLoopCtx := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, `loopExitHandler` will be also called from the pulling loop (see #769). It's a mechanism that allows to completely shut down the entire forwarder in case one of the loops (pulling or pushing) raises an error.

relay/pkg/block/push.go
lukasz-zimnoch: I like this idea! Do you mind I also leave a TODO for that and address it separately as I don't have `pull.go` in this PR?

relay/pkg/block/forwarder.go
```diff
@@ -0,0 +1,143 @@
...
+		errChan:         make(chan error, 1),
+		loopExitHandler: cancelLoopCtx,
+	}
+
+	go forwarder.pushingLoop(loopCtx)
+
+	return forwarder
+}
+
+func (f *Forwarder) pushingLoop(ctx context.Context) {
```
pdyraga, line 76: Thoughts on renaming this function to `relayLoop`? (in a separate PR)

`pushingLoop` is confusing because we both pull and push inside this function.

relay/pkg/block/forwarder.go
lukasz-zimnoch: Sure. Please see: https://github.com/keep-network/tbtc/commit/71c059077a00c995b8eff208908ef2788873a92d

relay/pkg/block/push.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/keep-network/tbtc/relay/pkg/btc"
+)
+
+// pullHeadersFromQueue waits until we have `headersBatchSize` headers from
+// the queue or until the queue fails to yield a header for
+// `headerTimeout` duration and returns them from the function.
```
pdyraga, line 20: Let's clarify the case when the timeout is hit. I was confused when reviewing the code using this function thinking no headers (empty slice) can be returned in the case of a timeout. My proposition:
```
// pullHeadersFromQueue blocks until there is `headersBatchSize` headers in
// the queue or until `headerTimeout` is hit and no more headers are available
// in the queue. Function returns headers from the queue but not less than one
// and no more than `headersBatchSize` headers.
```

relay/pkg/block/forwarder.go
pdyraga: Nevermind, this is push to host chain, other comments clarified it.

lukasz-zimnoch: If we apply changes proposed in https://github.com/keep-network/tbtc/pull/767#discussion_r603954935 we will have `getHeadersFromQueue` and `pushHeadersToHostChain` so I think it won't be confusing anymore as pull/push naming will be only related with chains. Also `pushingLoop` fits better with the `pullingLoop` added in #769.

relay/pkg/block/forwarder.go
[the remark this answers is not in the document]
pdyraga: In this case, what do you think about passing cancel function as a parameter? It is cleaner approach IMO: we pass the context and a function to cancel it, just like it is returned by `context.WithCancel`:

```
pushingLoop(ctx context.Context, cancelFn context.CancelFn) 
```

relay/pkg/block/push.go
lukasz-zimnoch: Sure: https://github.com/keep-network/tbtc/commit/49448a1e7de6aece3ab985d40829ddfbed1346d1 . Keep in mind this function can return an empty slice if the context is cancelled. Flagged that in the comment.

relay/pkg/block/forwarder.go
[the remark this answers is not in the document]
lukasz-zimnoch: I wanted to avoid this because I think the loop itself should not decide about cancelling its context. In my opinion, it's the responsibility of the forwarder/relay itself as it controls the lifecycle of pulling/pushing loops. What the loop should do is to inform the forwarder about an abnormal exit and the forwarder should decide what to do - in this case, cancel the `loopCtx` but in other situation, it may want to restart the loop. Also, having `loopExitHandler` we can easily add other behavior which should take place on loop exit if desired.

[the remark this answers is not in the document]
pdyraga: So maybe:
```
	go func() {
		forwarder.pushingLoop(loopCtx)
		cancelLoopCtx() // loop exited, cancel the contet
	}
```
?

I don't like we need to trust `pushingLoop` to call `loopExitHandler`. I do agree that it should not cancel the context, so the approach when the context is cancelled straight from `RunForwarder` in case one of the loops (currently we have just one loop, soon we'll have two) exits is IMO better - we do not need to trust those functions, everything is controlled by `RunForwarder` which triggered them.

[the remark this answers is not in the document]
lukasz-zimnoch: Yep, that's a much better one. Please see: https://github.com/keep-network/tbtc/commit/37c633f2384860b280f68b77636116d214e09584

pdyraga (approved):

lukasz-zimnoch:
Heads-up: A follow-up PR addressing outstanding change requests is here https://github.com/keep-network/tbtc/pull/779

PR #769, opened 2021-03-22
Relay maintainer block pull

tomaszslabon:
Depends on: #767 

This pull request contains functionalities responsible for pulling block headers from the Bitcoin network and putting them on the queue to be forwarded to the ETH network.

Work done so far:
- basic structure for the pulling loop
- added `GetHeaderByDigest`
- added finding best block to start pulling loop
- implemented pulling loop
- unit tests for pulling header loop


relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+	betterOrSameHeader, err := f.btcChain.GetHeaderByHeight(bestHeader.Height)
+	if err != nil {
+		return nil, err
+	}
+
+	// see if there's a better block at that height
+	// if so, crawl backwards
+
+	// TODO: Is it ever possible that bestHeader and betterOrSameHeader are not
+	// equal?
```
lukasz-zimnoch, line 99: I think it may happen that:
1. We register new best header with digest X and height Y on the host chain
2. We crash soon due to an error
3. A reorg happens on BTC chain and height Y now corresponds to a digest Z 
4. We recover from the crash and initialize the pulling loop so we need to make sure what happened in the meantime on BTC chain

lukasz-zimnoch (changes requested):
First pass. Probably more to come later.

relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+		case <-f.quit:
+			return
+		default:
+			chainHeight, err := f.btcChain.GetBlockCount()
+			if err != nil {
+				f.errChan <- fmt.Errorf("could not get block count [%v]", err)
+				return
+			}
+
+			if latestHeight <= chainHeight {
```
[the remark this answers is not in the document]
tomaszslabon: I added some comments explaining why we do so.

relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+		quit:         make(chan bool, 1),
 	}
 
+	go forwarder.pullingLoop(ctx)
 	go forwarder.pushingLoop(ctx)
 
 	return forwarder
 }
 
+func (f *Forwarder) findBestBlock() (*btc.Header, error) {
```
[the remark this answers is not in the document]
tomaszslabon: Good idea.

relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+		return nil, err
+	}
+
+	betterOrSameHeader, err := f.btcChain.GetHeaderByHeight(bestHeader.Height)
+	if err != nil {
+		return nil, err
+	}
+
+	// see if there's a better block at that height
+	// if so, crawl backwards
```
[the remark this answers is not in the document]
tomaszslabon: Right

relay/pkg/btc/remote/remote.go
```diff
@@ -140,6 +146,41 @@ func serializeHeader(header *wire.BlockHeader) ([]byte, error) {
...
+	blockHeader, err := rc.client.GetBlockHeader((*chainhash.Hash)(&digest))
+	if err != nil {
+		return nil, fmt.Errorf(
+			"could not get block header for hash [%s]: [%v]",
+			digest.String(),
+			err,
+		)
+	}
+
+	headerVerbose, err := rc.client.GetBlockHeaderVerbose((*chainhash.Hash)(&digest))
```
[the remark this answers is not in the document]
tomaszslabon: I didn't find any cleaner btc api method for returning block height based on block hash.

I see one alternative:
Instead of calling `GetBlockHeader` and `GetBlockHeaderVerbose` we could just call `GetBlockHeaderVerbose` as it contains all the information we need.
But `GetBlockHeaderVerbose` returns data in not very useful format (for example hashes are returned as strings).

We would have to construct `wire.BlockHeader` structure (which is needed for serialization) from the object returned from `GetBlockHeaderVerbose`  which would probably add more code.

relay/pkg/node/node.go
```diff
@@ -45,6 +45,7 @@ func runForwarderControlLoop(
 				"error occurred during block forwarding: [%v]",
 				err,
 			)
+			close(forwarder.QuitChan())
```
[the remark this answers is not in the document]
tomaszslabon: I agree,  using quit channel was a really messy solution.

relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+						latestHeight,
+						err,
+					)
+					return
+				}
+
+				// TODO: Consider just comparing hashes - should be enough
+				if !headersEqual(newHeader, lastAdded) {
+					f.headersQueue <- newHeader
+					copyHeaders(lastAdded, newHeader)
```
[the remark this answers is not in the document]
tomaszslabon: Yes, we can simply use pointers.

relay/pkg/btc/remote/remote.go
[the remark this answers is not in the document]
lukasz-zimnoch: Ok fine. I think we can leave it as is.

relay/pkg/btc/btc.go
```diff
@@ -40,14 +43,24 @@ type Header struct {
...
-		h.PrevHash,
-		h.MerkleRoot,
-		hex.EncodeToString(h.Raw),
+		header.Hash,
+		header.Height,
+		header.PrevHash,
+		header.MerkleRoot,
+		hex.EncodeToString(header.Raw),
 	)
 }
```
lukasz-zimnoch, line 65: Just a style note: It's a general Go convention that we use short abbreviations for receiver names. In those cases it should be just `(h *Header)`. If the struct name is more complicated, for example, `BitcoinBlockHash`, we use first letters of all parts: `(bbh *BitcoinBlockHash)`

relay/pkg/btc/btc.go
```diff
@@ -16,6 +16,9 @@ type Handle interface {
 
 	// GetHeaderByDigest returns the block header for given digest (hash).
 	GetHeaderByDigest(digest Digest) (*Header, error)
+
+	// GetBlockCount returns the number of blocks in the longest blockchain
+	GetBlockCount() (int64, error)
```
lukasz-zimnoch, line 21: `btc/local` chain doesn't implement this method and this cause failure of the CI job

relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+	lastAdded := &btc.Header{}
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Infof("forwarder context is done")
+			return
+		case <-f.quit:
+			return
+		default:
```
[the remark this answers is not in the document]
tomaszslabon: I created `pull.go` file and moved there `findBestBlock` and also functions `pullHeaderFromBtcNetwork` and `pushHeaderToQueue` that I created.

As for extracting more logic from `default`, it is hard to extract anything more, as there is a loop there and the logic in that loop uses variables (`nextHeaderHeight` and `lastAdded`) which are outside the scope of the loop.

I would have to pass them to the extracted function.

relay/pkg/block/pull.go
```diff
@@ -0,0 +1,48 @@
+package block
+
+import "github.com/keep-network/tbtc/relay/pkg/btc"
+
+func (f *Forwarder) pullHeaderFromBtcNetwork(height int64) (*btc.Header, error) {
```
lukasz-zimnoch, line 5: Let's call it `pullHeaderFromBtcChain` to be consistent with what we have on the push side.

relay/pkg/block/pull.go
```diff
@@ -0,0 +1,48 @@
...
+	return f.btcChain.GetHeaderByHeight(height)
+}
+
+func (f *Forwarder) pushHeaderToQueue(header *btc.Header) {
+	f.headersQueue <- header
+}
+
+func (f *Forwarder) findBestBlock() (*btc.Header, error) {
```
lukasz-zimnoch, line 13: `findBestBlock` -> `findBestHeader`?

relay/pkg/block/forwarder.go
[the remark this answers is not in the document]
lukasz-zimnoch: What about making `nextHeaderHeight` and `lastAdded` private fields of the `Forwarder` struct? We already use `processedHeaders` on the push side so we may introduce:
- `nextPulledHeaderHeight`
- `lastPulledHeader`
This may open more possibilities to take processing logic out of the pulling loop. What do you think?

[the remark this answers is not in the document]
tomaszslabon: I created  `pullNextHeader` and moved it to `pull.go`.
I could move there even more logic from `default` there, but I think I will be easier to think about the pulling loop if we leave the sleeping part on top, so it is clearly visible.

relay/pkg/block/forwarder.go
tomaszslabon: That makes sense. Thanks.

relay/pkg/block/pull.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (f *Forwarder) pullHeaderFromBtcChain(height int64) (*btc.Header, error) {
+	return f.btcChain.GetHeaderByHeight(height)
+}
+
+func (f *Forwarder) pushHeaderToQueue(header *btc.Header) {
+	f.headersQueue <- header
+}
+
+func (f *Forwarder) setNextPullHeaderHeight(height int64) {
```
lukasz-zimnoch, line 38: There is no sense to make a setter for `nextPullHeaderHeight` in that case. We're not doing any validation or complex logic here. We can set the `nextPullHeaderHeight` field directly from calling code.

relay/pkg/block/forwarder.go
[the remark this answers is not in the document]
lukasz-zimnoch: I get your point about the sleeping part but I think that sleep is strictly related to the "pull from BTC chain" part not the entire pull loop. My proposition of how the `default` case should look:
```
		default:
                        logger.Infof("pulling new header from BTC chain")

			header, err := f.pullHeaderFromBtcChain(ctx)
			if err != nil {
				f.errChan <- fmt.Errorf("could not pull header: [%v]", err)
				return
			}

                        logger.Infof("pushing new header to the queue")

			f.pushHeaderToQueue(header)
```
And that's it. Regarding methods used here look at the next comment.

relay/pkg/block/pull.go
```diff
@@ -0,0 +1,77 @@
...
+	return nil
+}
+
+func (f *Forwarder) pullHeaderFromBtcChain(height int64) (*btc.Header, error) {
+	return f.btcChain.GetHeaderByHeight(height)
+}
+
+func (f *Forwarder) pushHeaderToQueue(header *btc.Header) {
+	f.headersQueue <- header
+}
```
lukasz-zimnoch, line 36: Regarding those guys, the problem is as follows:
- `pullNextHeader` does too much. It's name suggests it only pulls next header but it actually pushes it to the queue and sets some internal forwarder state.
- `pullHeaderFromBtcChain` and `pushHeaderToQueue` don't do anything deserving a separate method

I'd rather go that way:
```
func (f *Forwarder) pullHeaderFromBtcChain(
	ctx context.Context,
) (*btc.Header, error) {
	for {
		if ctx.Err() != nil {
			return nil, ctx.Err()
		}

		chainHeight, err := f.btcChain.GetBlockCount()
		if err != nil {
			return nil, fmt.Errorf("could not get block count [%v]", err)
		}

		// Check if there are more headers to pull or we are above the chain's
		// tip and need to sleep until the chain adds more headers
		if f.nextPullHeaderHeight <= chainHeight {
			nextHeader, err := f.btcChain.GetHeaderByHeight(f.nextPullHeaderHeight)
			if err != nil {
				return nil, fmt.Errorf(
					"could not get header by height at [%d]: [%v]",
					f.nextPullHeaderHeight,
					err,
				)
			}

			// TODO: Consider just comparing hashes - should be enough
			//  and check if nextHeader and lastPulledHeader header can ever
			//  be not equal
			if !nextHeader.Equals(f.lastPulledHeader) {
				return nextHeader, nil
			}
		}

		select {
		case <-time.After(forwarderPullingSleepTime):
		case <-ctx.Done():
		}
	}
}

func (f *Forwarder) pushHeaderToQueue(header *btc.Header) {
	f.headersQueue <- header
	f.lastPulledHeader = header
	f.nextPullHeaderHeight++
}
```
That way we solve all the problems and we obtain a clean pulling loop. All pull logic lives in `pull.go` and reading from BTC chain is separated from pushing to the queue. What do you think?

lukasz-zimnoch (changes requested):

relay/pkg/block/pull.go
tomaszslabon: I'm ok with that.

relay/pkg/block/forwarder.go
```diff
@@ -56,20 +63,126 @@ func RunForwarder(
...
+				if err != nil {
+					f.errChan <- fmt.Errorf(
+						"could not get header by height at %d: [%v]",
+						latestHeight,
+						err,
+					)
+					return
+				}
+
+				// TODO: Consider just comparing hashes - should be enough
```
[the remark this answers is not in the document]
tomaszslabon: I left a TODO and added "Go rewrite: Proper support of BTC reorgs" task in code.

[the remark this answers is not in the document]
lukasz-zimnoch: Excellent, thanks!

lukasz-zimnoch:
No more comments regarding the code. Waiting for unit tests now!

relay/pkg/block/forwarder_test.go
```diff
@@ -12,6 +12,44 @@ import (
...
+
+	localChain, err := chainlocal.Connect()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Run forwarder with an empty Bitcoin chain and wait for a moment so
+	// the pulling loop goes to sleep
+	forwarder := RunForwarder(ctx, btcChain, localChain)
+	time.Sleep(1 * time.Second)
```
lukasz-zimnoch, line 33: Can we shorten this sleep? Maybe `100ms` will be enough?

relay/pkg/block/forwarder_test.go
```diff
@@ -12,6 +12,44 @@ import (
...
+
+	// While the pulling loop is sleeping, add headers to Bitcoin chain and
+	// cancel context
+	btcChain.SetHeaders([]*btc.Header{
+		{Height: 1, Hash: [32]byte{1}, PrevHash: [32]byte{0}},
+		{Height: 2, Hash: [32]byte{2}, PrevHash: [32]byte{1}},
+	})
+	cancelCtx()
```
lukasz-zimnoch, line 41: I'd add a short sleep (`100ms`?) after context cancel. This is needed to make sure nothing is added on the next pulling loop iteration which will occur after the one which has been ended by context cancellation.

relay/pkg/block/forwarder_test.go
```diff
@@ -12,6 +12,44 @@ import (
...
+
+	// The forwarder's queue should be empty
+	queueLength := len(forwarder.headersQueue)
+	if queueLength != 0 {
+		t.Errorf(
+			"unexpected headers in queue, queue length: [%d]\n",
+			queueLength,
+		)
+	}
```
lukasz-zimnoch, line 50: We usually use the `expected & actual` pattern:
```
        expectedQueueLength := 0
	actualQueueLength := len(forwarder.headersQueue)
	if expectedQueueLength != actualQueueLength {
		t.Errorf(
			"unexpected headers queue length:\n"+
				"expected: [%v]\n"+
				"actual:   [%v]\n",
			expectedQueueLength,
			actualQueueLength,
		)
	}
```

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	for _, expectedHeader := range headers {
+		actualHeader := <-forwarder.headersQueue
+		if !actualHeader.Equals(expectedHeader) {
+			t.Errorf(
+				"unexpected header in queue:\n"+
+					"expected: [%s]\n"+
+					"actual:   [%s]\n",
+				expectedHeader.String(),
+				actualHeader.String(),
```
lukasz-zimnoch, line 63: Hint: We can also use `%+v` placeholder to print struct fields along with their names:
```
"unexpected header in queue:\n"+
        "expected: [%+v]\n"+
        "actual:   [%+v]\n",
    expectedHeader,
    actualHeader,
```

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+	// Bitcoin chain tip
+	var headers []btc.Header
+	for i := 0; i < 2; i++ {
+		header, err := forwarder.pullHeaderFromBtcChain(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
+		headers = append(headers, *header)
+		forwarder.nextPullHeaderHeight++
+	}
```
lukasz-zimnoch, line 101: How do we know we are actually not waiting here? I think we need some kind of timeout mechanism here to make sure headers are returned without delay.

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+	btcChain := bc.(*btclocal.Chain)
+
+	btcChain.SetHeaders([]*btc.Header{
+		{Height: 1, Hash: [32]byte{1}, Raw: []byte{1}},
+		{Height: 2, Hash: [32]byte{2}, Raw: []byte{2}},
+	})
+
+	forwarder := &Forwarder{
+		btcChain:                  btcChain,
+		forwarderPullingSleepTime: 3 * time.Second,
```
lukasz-zimnoch, line 87: Can we make that time lower? How about `1s` or even `500ms`?

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+		header, err := forwarder.pullHeaderFromBtcChain(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		headerChan <- header
+	}()
+
+	select {
+	case <-time.After(time.Second):
```
lukasz-zimnoch, line 132: If we shorten the loop sleeping time, we can use a smaller value here as well.

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+	case header := <-headerChan:
+		t.Fatalf("unexpected header returned[%v]", header)
```
lukasz-zimnoch, line 134: Please add space between `returned` and `[%v]`

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	btcChain.AppendHeader(
+		&btc.Header{Height: 3, Hash: [32]byte{3}, Raw: []byte{3}},
+	)
+
+	select {
+	case <-time.After(3 * time.Second):
+		t.Fatal("failed to return header")
```
lukasz-zimnoch, line 143: Maybe `header timeout has been exceeded`?

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+	case header := <-headerChan:
+		t.Fatalf("unexpected header returned[%v]", header)
+	}
+
+	btcChain.AppendHeader(
+		&btc.Header{Height: 3, Hash: [32]byte{3}, Raw: []byte{3}},
+	)
+
+	select {
+	case <-time.After(3 * time.Second):
```
lukasz-zimnoch, line 142: If we shorten the loop sleeping time, we can use a smaller value here as well.

relay/pkg/block/pull_test.go
```diff
@@ -0,0 +1,261 @@
...
+			"unexpected header:\n"+
+				"expected: [%s]\n"+
+				"actual:   [%s]\n",
+			expectedHeader.String(),
+			header.String(),
+		)
+	}
+}
+
+func TestFindBestHeader_HostChainDoesNoReturnBestBlock(t *testing.T) {
```
lukasz-zimnoch, line 201: `...DoesNot...`

relay/pkg/block/push_test.go
```diff
@@ -18,7 +18,8 @@ func TestPullHeadersFromQueue(t *testing.T) {
 	defer cancelCtx()
 
 	forwarder := &Forwarder{
-		headersQueue: make(chan *btc.Header, headersQueueSize),
+		headersQueue:              make(chan *btc.Header, headersQueueSize),
+		forwarderPullingSleepTime: defaultForwarderPullingSleepTime,
```
lukasz-zimnoch, line 22: Do we need to set `forwarderPullingSleepTime`? Is this test failing without that?

tomaszslabon: It's not needed here.

relay/pkg/block/pull_test.go
tomaszslabon: Shortened it to 300ms.

relay/pkg/block/pull_test.go
tomaszslabon: I changed the test a bit and ran these two `pullHeaderFromBtcChain` calls on a separate goroutine and added a timeout.

lukasz-zimnoch (approved):
LGTM! 🚀 I'll merge once #767 lands on master!

relay/pkg/block/forwarder.go
```diff
@@ -27,6 +27,10 @@ const (
 	// Duration for which the forwarder should rest after performing
 	// a push action.
 	forwarderPushingSleepTime = 45 * time.Second
+
+	// Default duration for which the forwarder should rest after reaching the tip of
+	// Bitcoin blockchain
```
pdyraga, line 32: Can we add one or two sentences explaining why it needs/wants to rest?

relay/pkg/block/forwarder.go
```diff
@@ -27,6 +27,10 @@ const (
...
+	defaultForwarderPullingSleepTime = 60 * time.Second
```
pdyraga, line 33: Why do we name it `default*` given that it is not configurable and the rest of the constants defined in this section are not prefixed with `default`?

relay/pkg/block/forwarder.go
```diff
@@ -56,18 +64,59 @@ func RunForwarder(
...
 	go forwarder.pushingLoop(loopCtx)
 
 	return forwarder
 }
 
+func (f *Forwarder) pullingLoop(ctx context.Context) {
+	logger.Infof("running new block pulling loop")
+
+	defer func() {
+		logger.Infof("stopping current block pulling loop")
```
pdyraga, line 85: Is it block pull or header pull?

lukasz-zimnoch: It originates from the approach taken in #767. I think we could leave as is and I will refactor that according to https://github.com/keep-network/tbtc/pull/767#discussion_r603913978

pdyraga: Yeah, agreed. I wrote this comment before I realized what is happening and the discussion in https://github.com/keep-network/tbtc/pull/767#discussion_r603913978 😆

relay/pkg/block/forwarder.go
pdyraga: For example, we could add the following sentence:
> The forwarder waits for this time before it tries to fetch a new tip from the Bitcoin blockchain, giving it some time to mine new blocks.

tomaszslabon: Perhaps:
"Default duration for which the forwarder should rest after reaching the tip of Bitcoin blockchain as there are no more headers to pull"
?

relay/pkg/block/forwarder.go
tomaszslabon: `forwarderPullingSleepTime` inside `Forwarder` is configurable - in tests we use a shorter value, while in real implementation we use `defaultForwarderPullingSleepTime`.

Perhaps we could name this constant `forwarderPullingSleepTime`, while the field inside `Forwarder` could be named `poolingSleepTime`?

lukasz-zimnoch: Maybe `pullingSleepTime`?

tomaszslabon: Yes :)

pdyraga: I see. Do we plan to make this value configurable in the config file? Because if so, prefix makes sense. Otherwise, I'd name it `forwarderPullingSleepTime` in both cases, we can use the same name.

relay/pkg/block/forwarder.go
pdyraga: FWIW, I do not consider this or any other comment I left here blocking. Please feel free to merge @lukasz-zimnoch and we can address them in any other PR we have in the queue or a new one.

relay/pkg/block/forwarder.go
tomaszslabon: We do not plan to make it configurable via config file at this moment.
I'm ok with `forwarderPullingSleepTime` both in Forwarder and in const section.

pdyraga: `pullingSleepTime` is even better!

PR #770, opened 2021-03-24
Relay maintainer: Refactor of `btc` package

lukasz-zimnoch:
Depends on: #769 

This pull request gets rid of the `btc/remote` and `btc/local` packages. It moves the `remote.go` and `local.go` files straight into the `btc` package. This change improves the user experience of the `btc` package. Client code can now use clean `btc.Connect` and `btc.ConnectLocal` functions instead of importing `btc/remote` and `btc/local` and calling their own `Connect` functions.

PR #771, opened 2021-03-24
Relay maintainer: Push blocks edge cases

lukasz-zimnoch:
Depends on: #770 

Here we introduce all missing parts of the pushing logic. Specifically, this pull request includes:
- Adding a preflight check of the `MarkNewHeaviest` transaction. It's needed because `MarkNewHeaviest` often fails if the preceding `AddHeader` transaction is not yet mined and a simple retry, after a short back-off, usually does the work. This change will increase `MarkNewHeaviest` transactions success rate and avoid unnecessary forwarder restarts
- Handling the case when the difficulty change occurs at the beginning of the headers batch. In that case, `AddHeadersWithRetarget` contract method is called instead of the regular `AddHeaders`.
- Handling the case when the difficulty change occurs in the middle of the headers batch. In that case, headers batch should be split in two parts. Pre-change headers should be submitted using the regular `AddHeaders` method while the post-change ones using `AddHeadersWithRetarget`.


relay/pkg/chain/ethereum/ethereum.go
```diff
@@ -236,3 +236,27 @@ func (ec *ethereumChain) MarkNewHeaviest(
...
+		ancestorDigest,
+		currentBestHeader,
+		newBestHeader,
+		limit,
+		nil,
+	)
+	if err != nil {
+		logger.Warnf("MarkNewHeaviest preflight failed with: [%v]", err)
+		return false
+	}
```
pdyraga, line 259: There might be a situation when `CallMarkNewHeaviest` fails for a reason other than the preceding `AddHeader` transaction being not yet mined and in such case, we'll log the problem on warn level instead of logging it on error level. 

Is there any other on-chain check we can perform instead of doing a preflight call? Can we use one of the view functions of [the contract](https://github.com/summa-tx/relays/blob/master/solidity/contracts/Relay.sol)?

lukasz-zimnoch: If the problem persists, the whole process will eventually fail (after reaching max attempts) on error level and cause a restart of the forwarder. The same logic was used in the legacy relay maintainer. Regarding view functions, that needs to be investigated, hard to say at the moment. However, even if it's possible, it will be more complicated as it will probably require to call multiple functions (to cover all the preconditions checks of `MarkNewHeaviest`). We will have to deal with possible errors from calls as well and decide what to do with them, in the same way as we're currently doing here.

relay/pkg/chain/chain.go
```diff
@@ -51,4 +51,14 @@ type Relay interface {
 		newBestHeader []byte,
 		limit *big.Int,
 	) error
+
+	// MarkNewHeaviestPreflight performs a preflight call of the
+	// MarkNewHeaviest method to check whether its execution will
+	// succeed.
```
pdyraga, line 57: Worth adding information on what result is expected from a successful and failed preflight call.

relay/pkg/block/push.go
```diff
@@ -114,39 +141,102 @@ func (f *Forwarder) addHeaders(headers []*btc.Header) error {
...
-		return fmt.Errorf("could not find last common ancestor: [%v]", err)
-	}
+}
 
-	limit := newBestHeader.Height - lastCommonAncestor.Height + 1
+func (f *Forwarder) updateBestHeader(
+	ctx context.Context,
+	newBestHeader *btc.Header,
+) error {
+	totalAttempts := 30
```
pdyraga, line 177: Should this value be extracted to constants in `forwarder.go`?

relay/pkg/block/push.go
```diff
@@ -114,39 +141,102 @@ func (f *Forwarder) addHeaders(headers []*btc.Header) error {
...
+				lastCommonAncestor.Hash,
+				currentBestHeader.Raw,
+				newBestHeader.Raw,
+				limit,
+			)
+		}
+
+		// wait a constant back-off time
+		select {
+		case <-time.After(10 * time.Second):
```
pdyraga, line 230: Should 10 seconds value be extracted into constants in `forwarder.go`? Even if not, we should at least define it next to `totalAttempts` because only when they match each other, this loop makes sense. We now retry up to 5 minutes which makes sense. If we increased sleep time to e.g. 30 seconds, I'd argue 30 retries is too much.

relay/pkg/block/push.go
```diff
@@ -114,39 +141,102 @@ func (f *Forwarder) addHeaders(headers []*btc.Header) error {
...
-	return f.hostChain.MarkNewHeaviest(
-		lastCommonAncestor.Hash,
-		currentBestHeader.Raw,
-		newBestHeader.Raw,
-		big.NewInt(limit),
+	for attempt := 1; attempt <= totalAttempts; attempt++ {
+		logger.Infof(
+			"attempt [%v] to set header [%v] as new best",
+			attempt,
+			newBestHeader.Height,
```
pdyraga, line 183: Is `newBestHeader.Height` always up to date "as a new best" in the context of this loop, given that we fetch `currentBestHeader` later in this loop?

relay/pkg/block/push.go
```diff
@@ -70,15 +70,42 @@ func (f *Forwarder) pushHeadersToHostChain(
...
+					err,
+				)
+			}
+		}
 	} else {
 		// no difficulty change
 		logger.Infof(
-			"performing simple headers adding as difficulty doesn't " +
-				"change within headers batch",
+			"simply adding all headers as there is no difficulty change " +
```
pdyraga, line 107: Instead "simply" (it is unclear what does it actually mean), we can say "adding all headers without retargeting as there..."

relay/pkg/block/push.go
```diff
@@ -212,3 +309,23 @@ func packHeaders(headers []*btc.Header) []byte {
...
+	for _, header := range headers {
+		if header.Height%difficultyEpochDuration >= startMod {
+			preChangeHeaders = append(preChangeHeaders, header)
+		} else if header.Height%difficultyEpochDuration < startMod {
+			postChangeHeaders = append(postChangeHeaders, header)
+		} else {
+			logger.Warnf(
+				"could not assign header [%v] to pre/post-change part",
+				header,
+			)
```
pdyraga, line 326: Is it a warning or an error? It should never happen. I'd also log `startMod`.

relay/pkg/chain/ethereum/ethereum.go
pdyraga: Can we backlog a task to do this investigation? If the contract does not allow for a one-call check, doing a preflight is better from the perspective of Ethereum client calls bandwidth costs. But if there is a way to do one call, I think it is better than the preflight.

pdyraga (commented):
Left some notes but haven't had a chance to do a more detailed review. I'll leave this PR in @tomaszslabon's hands.

relay/pkg/block/push.go
```diff
@@ -70,15 +70,42 @@ func (f *Forwarder) pushHeadersToHostChain(
 
 	if startMod == 0 {
 		// we have a difficulty change first
-		// TODO: implementation
+		logger.Infof(
```
tomaszslabon, line 73: If we don't use any arguments to logging message, then perhaps `logger.Info` ?

relay/pkg/chain/ethereum/ethereum.go
lukasz-zimnoch: Sure, added to the backlog.

tomaszslabon (approved):

relay/pkg/chain/chain.go
lukasz-zimnoch: https://github.com/keep-network/tbtc/pull/771/commits/e4f08c6cbbf84b67056a0af6792788de659c8fcc

relay/pkg/block/push.go
lukasz-zimnoch: We can extract it. See https://github.com/keep-network/tbtc/commit/576e98d37e8d5c8c6d9b9d0a3640b40151e81dc3

relay/pkg/block/push.go
lukasz-zimnoch: See https://github.com/keep-network/tbtc/commit/576e98d37e8d5c8c6d9b9d0a3640b40151e81dc3

relay/pkg/block/push.go
lukasz-zimnoch: Sure: https://github.com/keep-network/tbtc/commit/3f435f44610bcb4d040ac5ea88f57f2d070e3ad6

relay/pkg/block/push.go
lukasz-zimnoch: Ok. See https://github.com/keep-network/tbtc/commit/3f435f44610bcb4d040ac5ea88f57f2d070e3ad6

relay/pkg/block/push.go
lukasz-zimnoch: See https://github.com/keep-network/tbtc/commit/ed0fb5df014da81a12118fd087306d8d66528ec5

PR #773, opened 2021-03-26
Ropsten migration updates

nkuba:
A couple of changes here, that we're hoping to stabilize ropsten migrations:
- updated `truffle` and `@truffle/hdwallet-provider`(*) to the latest ones. It also sorts the list of dependencies,
- increased some timeouts that were mentioned in https://github.com/trufflesuite/truffle/issues/3468
- updated object passed to HDWalletProvider to match the format mentioned in their doc.

We already confirmed in keep-core that such updates stabilized the migrations (https://github.com/keep-network/keep-core/pull/2355).

We also updated version of `@keep-network/keep-ecdsa` dependency to the one recently deployed on ropsten.
Version of this package was also changed to `pre`, so it can be picked up by the release process and migrated automatically on master merge.

(*) @truffle/hdwallet-provider is not the latest one as there was a bug introduced in 1.2.3 that breaks migrations (trufflesuite/truffle#3946).

### DIFF
--- a/solidity/package-lock.json
+++ b/solidity/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@keep-network/tbtc",
-  "version": "1.1.0-rc",
+  "version": "1.1.1-pre",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


pdyraga (approved):
Been there, done that.

michalinacienciala (commented):
Looks ok to me

PR #774, opened 2021-03-26
Relay maintainer: Metrics

lukasz-zimnoch:
Depends on: #771

This pull request adds the possibility to expose metrics by the relay maintainer. Specific metrics which are exposed:
- `btc_chain_connectivity`: indicates whether the relay has a working connection with the BTC chain. Possible values are `1` and `0` which correspond to `true` and `false` respectively,
- `host_chain_connectivity`: does the same as the above metric but is related to the host chain connectivity.
- `block_forwarding_active`: indicates whether the block forwarding process is enabled (`1`) or not (`0`)
- `block_forwarding_errors`: indicates the total number of errors raised by the block forwarding process during the entire relay node lifetime
- `unique_blocks_pulled`: indicates the total number of unique blocks pulled from the BTC chain by the block forwarding process, during the entire relay node lifetime. This metric doesn't count re-pulls which can occur during recovery after block forwarding error.
- `unique_blocks_pushed`: indicates the total number of unique blocks pushed to the host chain by the block forwarding process, during the entire relay node lifetime. This metric doesn't count re-pushes which can occur during recovery after block forwarding error.

By default, the first two `*_chain_connectivity` metrics are updated every `10 minutes` and this time can be customized via the `Metrics.ChainMetricsTick` config property. The rest of the metrics are updated every `10 seconds` and their tick can be set using `Metrics.NodeMetricsTick` property. The metrics endpoint is `/metrics` and metrics port can be set by the `Metrics.Port` property. In case it's not set, metrics will not be enabled. 

Note: Metrics named `block_forwarding_*` will be renamed according to the planned refactor described in https://github.com/keep-network/tbtc/pull/767#discussion_r603913978


relay/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+// ObserveHostChainConnectivity triggers an observation process of the
+// host_chain_connectivity metric.
+func ObserveHostChainConnectivity(
+	ctx context.Context,
+	registry *metrics.Registry,
+	hostChain chain.Handle,
+	tick time.Duration,
+) {
+	input := func() float64 {
+		_, err := hostChain.GetBestKnownDigest()
```
pdyraga, line 76: Good it's a cheap-to-compute function. Less compute units to charge for. 👌

relay/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+		if !isEnabled {
+			return 0
+		}
+
+		return 1
+	}
+
+	observe(
+		ctx,
+		"block_forwarding",
```
pdyraga, line 114: Thoughts on `block_forwarding_status`/`block_forwarding_state`? Just `block_forwarding` is a little unclear to me if it's just 0/1 or maybe some counter of blocks processed.

pdyraga: Oh, maybe `_active`? I think I've just realized that this metric will be updated based on the relay auction result.

relay/pkg/node/node.go
```diff
@@ -19,35 +24,54 @@ func Initialize(
...
+			n.stats.notifyBlockForwardingErrored()
 		case <-ctx.Done():
 			logger.Infof("stopping block forwarding")
 			return
 		}
 	}
 }
+
+// Stats returns relay node statistics.
+func (n *Node) Stats() *Stats {
```
pdyraga, line 75: I'd consider returning an interface instead of the struct. We'd not have to worry about `NotifyBlockPulled` and `NotifyBlocksPushed` being exported and that they can be called by the code that shouldn't do it. Then, we can pass around the structure to `RunForwarder` without having to define an interface there and mocking it in tests.

relay/pkg/node/stats.go
```diff
@@ -0,0 +1,63 @@
+package node
+
+// Stats gathers and exposes statistics of the relay node.
+type Stats struct {
+	blockForwardingEnabled bool
+	blockForwardingErrors  int
+	uniqueBlocksPulled     map[int64]bool
+	uniqueBlocksPushed     map[int64]bool
+}
```
pdyraga, line 9: Thoughts on protecting reads/writes to this structure with `RWMutex`?

relay/pkg/node/node.go
lukasz-zimnoch: I agree that `NotifyBlockPulled` and `NotifyBlocksPushed` should be protected. I think we can introduce some kind of `node.ReadonlyStats` interface and return it from `Stats` method. However, regarding `RunForwarding` I'm strongly in favor of leaving as it is. When `block.RunForwarder` accepts `block.ForwarderObserver` we are not tied to any implementation, if we need to pass another kind of observer or register multiple ones, it's easy to do. This approach is also idiomatic in Go. If we want to pass `node.Stats` struct to `RunForwarder` we need to move that struct somewhere out of the `node` pkg to avoid the import cycle. Also, we will tie the forwarder with `node.Stats` implementation. This is a quite high price to get rid of the test mock which is really not a problem here.

relay/pkg/metrics/metrics.go
lukasz-zimnoch: Yep, good catch. See https://github.com/keep-network/tbtc/commit/4b0a178cb5e6ce2fc02e92a54717ab42c3d6a8ab

lukasz-zimnoch:
Heads-up: A follow-up PR https://github.com/keep-network/tbtc/pull/779 changes metrics naming a little bit, according to the new approach.

pdyraga (approved):
Let's remember about adding a list of metrics in a `README` for the relay.

PR #775, opened 2021-03-29
Cleanup after moving to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvement and
migration from CircleCI jobs to GitHub Actions.

Now that we have most CircleCI jobs from `tbtc` moved to GH Actions
(and verified), we can implement some changes removing obsolete
CircleCI configs and cleaning up pieces of code used for workflows
testing. Here is a list of changes applied:
- removed unnecessary comments
- removed references to `rfc-18...` branches, which were used for 
  workflow testing
- removed `--dry-run` flag from command publishing npm image
- deleted couple of CircleCI jobs from `./circleci/config.yml` (those
  who are migrated to GH Actions and are not necessary for building of
  relay maintainer initcontainer)
- added restrictions on workflow triggering conditions (see more info 
  below)

As we've migrated most of the CircleCI jobs to GitHub Actions, we want
to introduce branch protection rules on master, requiring the passing
of particular jobs before merging PR to master. Those jobs in `tbtc`
are:
* `contracts-build-and-test`
* `contracts-lint`

What's more, we want to execute workflows containing those jobs only if
there were changes in particular paths. GitHub Actions allows for
specifying those paths by using `on.<event_name>.paths` syntax.
Unfortunately, such configuration for the `pull-request` event clashes
with the functionality of protected branches.
Example: If we set branch protection rule requiring the passing of
`job_a` and `job_a` will be part of `workflow_a` using
`on.pull_request.paths = folder_a` setting, the `workflow_a` won't be
started if we create PR that changes some files outside of `folder_a`.
Hence `job_a` won't be executed. But the PR will still require `job_a`
to be passing (or skipped) and will not allow for the merge of the PR
to the protected branch.
As skipping of the required jobs doesn't block the PRs from merging, we
decided to move the checking of paths for `pull_request` events from
workflow level to job level. It's not something natively supported by
GH Actions, but there is a `dorny/paths-filter` action that allows for
doing exactly that. We've set up the action to execute only if a
workflow was started by the `pull_request` event. Action checks the
list of files changed by the PR and when any of the files matches
provided filter, it sets the output to `true`. We then use that output
as a condition for execution of rest of the jobs in the workflow. This
way, if a particular job is required for the PR, but the path
conditions are not met, the job will be skipped (allowing for PR
merge).

Here is a summary of triggering conditions for `Solidity` workflow:

* workflow dispatch:
    Everything executed
* push on `master`:
    Everything executed if there are any changes IN this path:
    - `solidity/**`
* pull_request:
    Everything (except migrating/publishing contracts) executed
    if there are any changes IN this path:
    - `solidity/**`

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -6,15 +6,31 @@ on:
       - master
     paths:
       - "solidity/**"
-      - ".github/workflows/contracts.yml"
   pull_request:
-    paths:
-      - "solidity/**"
-      - ".github/workflows/contracts.yml"
   workflow_dispatch:
 
 jobs:
-  build-and-test:
+  contracts-detect-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      path-filter: ${{ steps.filter.outputs.path-filter }}
+    steps:
+    - uses: actions/checkout@v2
+      if: github.event_name == 'pull_request' 
+    - uses: dorny/paths-filter@v2
+      if: github.event_name == 'pull_request' 
+      id: filter
+      with:
+        filters: |
+          path-filter:
+            - './solidity/**'
+
+  contracts-build-and-test:
+    needs: contracts-detect-changes
+    if: |
+      github.event_name != 'pull_request'
+        || (github.event_name == 'pull_request'
+        && needs.contracts-detect-changes.outputs.path-filter == 'true') 
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -47,7 +63,12 @@ jobs:
       - name: Run tests
         run: npm run test:quick
 
-  lint:
+  contracts-lint:
+    needs: contracts-detect-changes
+    if: |
+      github.event_name != 'pull_request'
+        || (github.event_name == 'pull_request'
+        && needs.contracts-detect-changes.outputs.path-filter == 'true') 
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


michalinacienciala:
@nkuba , before approving PR, following jobs would need to be configured as required in GitHub's settings:
* `contracts-build-and-test`
* `contracts-lint`

.github/workflows/contracts.yml
```diff
@@ -1,27 +1,36 @@
...
+        filters: |
+          path-filter:
+            - './solidity/**'
+
+  contracts-build-and-test:
+    needs: contracts-detect-changes
+    if: |
+      github.event_name != 'pull_request'
+        || (github.event_name == 'pull_request'
+        && needs.contracts-detect-changes.outputs.path-filter == 'true')
```
nkuba, line 33: Will the simplification proposed below cover our use cases correctly?

```suggestion
      github.event_name != 'pull_request'
        || needs.contracts-detect-changes.outputs.path-filter == 'true'
```

.github/workflows/contracts.yml
```diff
@@ -1,27 +1,36 @@
...
+  contracts-detect-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      path-filter: ${{ steps.filter.outputs.path-filter }}
+    steps:
+    - uses: actions/checkout@v2
+      if: github.event_name == 'pull_request' 
+    - uses: dorny/paths-filter@v2
+      if: github.event_name == 'pull_request' 
+      id: filter
```
nkuba, line 22: Is this id required? I don't see it mentioned in the reference we use later: `needs.contracts-detect-changes.outputs.path-filter`.

michalinacienciala: I believe we need it. It's used in line 16. Thanks to it we are able to define a step output that will passed between jobs.

.github/workflows/contracts.yml
michalinacienciala: Good idea! Fixed in 18f59b24934a5c3d6d5b72f2a444aa042c1e6db8.

nkuba (approved):

PR #776, opened 2021-03-29
Relay maintainer: K8S testnet deployment

lukasz-zimnoch:
Depends on: #774

Here we introduce the init container script and K8S testnet deployment manifest for the Go relay.


tomaszslabon (approved):

relay/solidity/truffle-config.js
```diff
@@ -24,6 +25,17 @@ module.exports = {
             network_id: 1101,
             websockets: true,
         },
+        ropsten: {
```
nkuba, line 28: Can we align it with changes introduced in https://github.com/keep-network/tbtc/pull/773?

relay/solidity/migrations/btc-networks.js
```diff
@@ -13,9 +13,9 @@ module.exports = {
     },
 
     testnet: {
-        genesis: '0x0000c0205d1103efc13e6647977e3d65f253c3e762451e9ca9b920517d000000000000008442a07bcde3292a888277ea6337ba5bbdfa808ae01535846f19d843144c8f60478bb15e7b41011a88be5d36',
-        height: 1723030,
-        epochStart: '0xe2657f702faa9470815005305c45b4be2271c22ade1348e6fe00000000000000',
+        genesis: '0x0000453faa4dcc559d47ed24be33b5539638b797e2dbd7527e61d8c13f00000000000000c4cb592bba9489c40ae5b12efa0f39e4d98dd6d58b794b65f6c50e450ee9957081b96160c0ff3f19015ae4ae',
```
nkuba, line 16: I'm wondering if it would be worth replacing the hardcoded values with some function that will calculate them on the contract deployment. But this is something for the future.

infrastructure/kube/templates/relay/initcontainer/provision-relay/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY ./package.json /tmp/package.json
+COPY ./package-lock.json /tmp/package-lock.json
+
+RUN npm ci
+
+FROM node:13-alpine
+
+WORKDIR /tmp
+
+COPY --from=install /tmp .
```
nkuba, line 16: Is it worth using the `install` aliased image and copying the contents of the whole `/tmp` directory? What are we trying to achieve here? We will copy the whole `node_modules` anyway, so will the resulting image be any lighter?

relay/solidity/migrations/btc-networks.js
lukasz-zimnoch: Yeah, we have such a script in `local-setup`: https://github.com/keep-network/local-setup/blob/master/relay-genesis/relay-genesis.js . We need to think how to use it here in the future.

relay/solidity/truffle-config.js
lukasz-zimnoch: Done: https://github.com/keep-network/tbtc/commit/6e484a65b332186b90ee4a5caefd138413933667

infrastructure/kube/templates/relay/initcontainer/provision-relay/Dockerfile
lukasz-zimnoch: It's taken from the legacy relay maintainer `Dockerfile` and it's taken because`node:13-alpine` gives a smaller image, though it requires a multistage build since the base Alpine image doesn't include git. (reference commit: https://github.com/keep-network/tbtc/commit/418c6e200b8c321e120d65fbd562d1a65d350828)

tomaszslabon (approved):

PR #780, opened 2021-04-06
Added back-off time in Relay Maintainer's restart loop

tomaszslabon:
This pull request adds a 10 second back-off time to the `Relay Maintainer`'s restart loop, so that we can avoid being flooded with a large number of error logs when a permanent error occurs in the Relay Maintainer (e.g. we cannot connect to one of the blockchains).

### DIFF
--- a/relay/pkg/node/node.go
+++ b/relay/pkg/node/node.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"context"
+	"time"
 
 	"github.com/keep-network/tbtc/relay/pkg/header"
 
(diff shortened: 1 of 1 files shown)


lukasz-zimnoch (approved):

pdyraga:
Just to make sure I understand the change: it is not about delaying restart of the entire application, it is about delaying the restart of the main control loop inside the application. And we need to have the control loop because of how the relay maintainer contracts are structured and how they handle reorgs, correct? In other words, there is no way to say what exactly went wrong when submitting headers, all that we can do is to log the error, and do a fresh start of the header relay, correct?

lukasz-zimnoch:
Yep, specifically, the main control loop doesn't restart itself but just the relay instance it controls. This backoff time should help to avoid log errors flood in case of longer outages, for example, loss of ETH connectivity.

PR #788, opened 2021-05-03
Load non-sensitive data from file instead of from secrets

michalinacienciala:
Non-sensitive data previously accessed by using `secrets` context is
now accessed by invoking `keep-network/load-env-variables` action and
using `env` context.
BACKGROUND:
We've been storing id of the test environment as
`KEEP_TEST_ETH_NETWORK_ID` organization secret in GitHub's settings to
allow for easy access to that value from multiple projects.
But it turned out to be problematic, as GHA in some situations redacts
workflow logs and actions outputs/inputs based on the values stored in
secrets. So if the secret holds small numeric value, it is quite
probable that at some point in time this value will be recognized as a
substring of `github.sha` or another variable and the `github.sha` will
be redacted with `***` in place of the recognized pattern. If such
redacted value is configured as input of some action, the action gets
invoked by the GHA without that input, which can cause run failure or
can mess up the workflow results.
As a solution for this problem we decided to no longer store
frequently-used, non-sensitive data such as `KEEP_TEST_ETH_NETWORK_ID`
in GitHub's secrets, but instead store them in a file kept in
`keep-network/ci` repository and read that data to `env` context in
order to be able to use it when running workflows. A
`keep-network/load-env-variables` action has been implemented to allow
for loading of the variables to `env` context in one step. The scope of
the loaded variables is limited to the job in which they were loaded.
To use the variables in other job, action `load-env-variables` needs to
be executed there as well.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -101,7 +101,6 @@ jobs:
     if: |
       github.ref == 'refs/heads/master'
         && github.event_name != 'pull_request'
-        || github.ref == 'refs/heads/rfc-18/load-env-var'
     environment: keep-test # keep-test requires a manual aproval
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -114,12 +114,14 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Load environment variables
-        uses: keep-network/load-env-variables@main # TODO: use @v1 once availible
+        uses: keep-network/load-env-variables@v1
         env: 
           CI_GITHUB_TOKEN: ${{ secrets.CI_GITHUB_TOKEN }}
         with:
+          # TODO: Consider passing of `environment` input instead of using 
+          # hardcoded value. Would require some rework in action's code or
+          # in config files.
           environment: 'ropsten'
-          ref: 'master'
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v2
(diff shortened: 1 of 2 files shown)


michalinacienciala:
Test configuration (references to development branch) is now removed.
Here are the link to the workflow which used that test configuration:
Relay: https://github.com/keep-network/tbtc/actions/runs/809131293 (succeeded)
Solidity: https://github.com/keep-network/tbtc/actions/runs/809131753 (failed, but fail not related to changes in #788)

.github/workflows/contracts.yml
```diff
@@ -101,6 +101,7 @@ jobs:
     if: |
       github.ref == 'refs/heads/master'
         && github.event_name != 'pull_request'
+        || github.ref == 'refs/heads/rfc-18/load-env-var'
```
nkuba, line 104: Can we remove it?

nkuba (changes requested):

.github/workflows/contracts.yml
michalinacienciala: Done, see 54ef99401b67476551e4e023357de64df9604501.

.github/workflows/contracts.yml
```diff
@@ -134,22 +144,20 @@ jobs:
 
       - uses: google-github-actions/setup-gcloud@v0.2.0
         with:
-          project_id: ${{ secrets.GOOGLE_PROJECT_ID }}
+          project_id: ${{ env.GOOGLE_PROJECT_ID }}
           service_account_key: ${{ secrets.KEEP_TEST_GCR_JSON_KEY }}
 
       - name: Fetch external contracts artifacts
         env:
-          CONTRACT_DATA_BUCKET: ${{ secrets.KEEP_TEST_CONTRACT_DATA_BUCKET }}
```
nkuba, line 142: Just to confirm, don't we need to set this variable?

michalinacienciala: This variable will be loaded during `Load environment variables` step, from `ropsten.env` config file. I've tested that variables loaded during that step are available in later steps' shells as environment variables (this is true within a job, the variables are not shared between jobs).

michalinacienciala: This variable will be loaded during `Load environment variables` step, from `ropsten.env` config file. I've tested that variables loaded during that step are available in later steps' shells as environment variables (this is true within a job, the variables are not shared between jobs).

nkuba (approved):

PR #791, opened 2021-05-05
Execute chain of workflows by manual triggering of upstream build

michalinacienciala:
This PR is part of a bigger work, spanning multiple repositories.
The main purpose of the work is to automate the process of workflows`
executions during product deployment, so that the workflows would be
triggered in a particular logical order and at the correct moment. By
that we mean:
* workflows requiring the presence of artifacts created by other
  workflows would not start before those upstream workflows are
  finished
* successfully executed workflows would automatically kick-off
  downstream workflows specified in the config file
* if any workflow in that chain of executions fails, the downstream
  workflows will not be triggered

To achieve the above requirements, following changes were introduced:
* A repository (called `ci`) managing the order of workflow executions
  was created
* The `config/config.json` file in `ci` repo describes the
  upstream/downstream relations between product modules and workflows
* The `Main` GH Actions workflow in `ci` repo allows for manual
  triggering of workflow deploying specified module, by providing the
  details of its upstream builds (to trigger deploy of the most
  upstream module, `Main` must be triggered with empty
  `upstream_builds` input)
* The triggering of module workflows by the `ci`'s `Main` workflow
  happens thanks to a `keep-network/run-workflow` action working as an
  intermediary (`Main` calls the `run-workflow`, which triggers
  workflow deploying the specified module)
* All NPM modules built as a part of the release process get versioned
  with `<base-version>-<environment>.<build-number>+<branch>.<commit>`
  version (thanks to `keep-network/upstream-builds-query` action that
  reads versions of older deployments and thanks to
  `keep-network/npm-version-bump@v2` action that bumps-up the version
  of the new artifact)
  (note: tagging of artifacts and versioning of docker images will be
  done at the later stage, outside this PR)
* Deployment workflows that are about to successfully finish will be
  executing the `keep-network/notify-workflow-completed` action as the
  last step - the action will then trigger the `ci`'s `Main` workflow
  with appropriate inputs
* Module workflows configuration has been adjusted to support the above
  features
* Conditions of execution of particular jobs/steps have been modified
  - after the changes contracts migration and deployments of artifacts
  happens only for workflows triggered by the `workflow-dispatch`
  event; actions like linting and sec scans are no longer executed for
  workflows triggered by `workflow-dispatch`
* Additionally, a couple of other changes were introduced to improve
  the process of build management:
  - listing of workflow metadata
  - unification of workflows' code
  - no longer pushing migrated contracts artifacts to GCP Bucket (see
    #789 for more details)
    
REF:
https://github.com/keep-network/keep-core/pull/2453
https://github.com/keep-network/keep-ecdsa/pull/777
https://github.com/keep-network/tbtc-dapp/pull/394
https://github.com/keep-network/tbtc.js/pull/122

TODO: same things as mentioned in https://github.com/keep-network/keep-core/pull/2453


solidity/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@keep-network/tbtc",
-  "version": "1.1.2-pre",
+  "version": "1.1.2-rc",
```
nkuba, line 3: This doesn't match `package.json`.

nkuba (changes requested):

solidity/package-lock.json
michalinacienciala: Fixed in 3b66897658d77aa9b095176442a534e510a9a7a1.

nkuba (approved):

PR #792, opened 2021-05-11
Fine-tuning of the relay pushing loop

lukasz-zimnoch:
This change is meant to reduce the number of unnecessary internal relay restarts. Those restarts often cause resubmissions of blocks upon relay initialization thus spend gas unnecessarily.

Restarts on mainnet occur usually when the gas price increases and the relay needs to bump up the gas price for pending transactions. In this case, transactions are not mined fast enough but the relay tries to continue its work anyway, which results in errors. There are two main types of errors:

- `could not find LCA`: this one raises when `markNewHeaviest` is invoked but the preceding `addHeaders` transaction is not yet mined
- `unknown block`: this one occurs when an `addHeaders` transaction is invoked but the preceding `addHeaders` transaction which relays the previous block is not yet mined. It occurs often when two single blocks are pulled, one after another, in a short period of time.

In the first case, we're trying to improve the situation by making more retries, with a longer backoff time, while looking for the LCA. If we use 10 attempts with a 30s backoff, we give around 5 minutes for the preceding `addHeaders` transaction to be mined. Such a time should allow the relay to find the right LCA in most cases and avoid a lot of restarts. On the other hand, this time cannot be too long as it may block the relay in case of non-recoverable failures for a long period.

The second error related to `unknown block` is less common. The only thing we can do here is to increase the pushing loop sleep time. However, this change could not be too big as it may slow down the relay significantly.


tomaszslabon (approved):
In my opinion it should not break anything.
I would give it a try on production.

pdyraga (approved):

PR #794, opened 2021-05-12
Help Yourself: Better Documentation in multiply-local-eth-btc-price

beaushinkle:
## Depends on #795 

This PR circles back to the review in https://github.com/keep-network/tbtc/pull/757#discussion_r584823654 to add clarity and standardization to the `multiply-local-eth-btc-price` script.

Tagging @Shadowfiend for review

PR #795, opened 2021-05-13
Unmangle Linting

beaushinkle:
Currently, JS linting in the TBTC project seems to be broken as evidenced by https://github.com/keep-network/tbtc/pull/794

From local testing, it seems like the issue is the `prettier --list-different **/*.js` command that's a part of the `lint:js` command. The `**/*` glob hits everything in `node_modules`, which ends up exceeding the argument size limit for the command line.

To fix this, we apply the following changes:

* Drop the config key and its contained `eslintPaths`
* Drop the `${npm_package_config_eslintPaths}` reference in both `lint:js` and `lint:fix:js`
* Drop the `&& prettier...` in both `lint:js` and `lint:fix:js`
* Update `eslint-config-keep` to reference `#0.3.0`, which pulls in the eslint-prettier setup
* Drop the `eslint-config-prettier` dependency, which will be pulled in by `eslint-config-keep` now.
* Bump the `prettier` dependency to ^2.0.2.
* Bump the `prettier-plugin-solidity` dependency to ^1.0.0-beta.10.

Then, we rerun `npm run lint:fix:js` and manually fix whatever it can't do for us:
```
/Users/beaushinkle/local/tbtc/solidity/docs/generate-api-docs.js
  32:3   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  33:11  error  'props' is assigned a value but never used                                                                  no-unused-vars
  52:9   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in

/Users/beaushinkle/local/tbtc/solidity/scripts/get-prices.js
  1:7  error  'fs' is assigned a value but never used    no-unused-vars
  2:7  error  'path' is assigned a value but never used  no-unused-vars
  3:7  error  'Web3' is assigned a value but never used  no-unused-vars
  ```
  
  tagging @nkuba for review

### CODE REVIEW

File: solidity/package.json [hunk 1]
```diff
@@ -56,15 +53,14 @@
...
+    "eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git#0.3.0",
     "eth-gas-reporter": "^0.2.22",
     "ethlint": "^1.2.4",
     "ganache-cli": "^6.4.3",
     "mocha": "^7.0.1",
     "moment": "^2.27.0",
-    "prettier": "^1.19.1",
-    "prettier-plugin-solidity": "^1.0.0-alpha.40",
+    "prettier": "^2.0.2",
+    "prettier-plugin-solidity": "^1.0.0-beta.10",
```
nkuba, line 63: As we're not running prettier in `lint:sol` do we still need this dependency?

solidity/package.json
```diff
@@ -56,15 +53,14 @@
     "bn.js": "^4.11.8",
     "chai": "^4.2.0",
     "eslint": "^5.16.0",
-    "eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git#0.2.0",
-    "eslint-config-prettier": "^6.10.0",
```
nkuba, line 60: I believe we need this guy:
```
> eslint .


Oops! Something went wrong! :(

ESLint: 5.16.0.
ESLint couldn't find the plugin "eslint-plugin-prettier". This can happen for a couple different reasons:
```

solidity/package.json
```diff
@@ -56,15 +53,14 @@
...
+    "eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git#0.3.0",
```
nkuba, line 56: Something's odd with the version of this dependency in the lock file. It's still referencing the commit of the `0.2.0` tag, not `0.3.0`. The rules that were used to lint the other files seem to be pulled from the previous config version.

solidity/package.json
Shadowfiend: This should be pulled in by eslint-config-keep. Did you `npm i`?

solidity/package.json
Shadowfiend: We should still be running `prettier` directly on the Solidity files, as I don't believe `eslint` will pick those up.

solidity/package.json
nkuba: I'm afraid it had never worked for me if I didn't add `eslint-config-prettier` explicitly. I upgraded npm today and boom it works!

solidity/package.json
beaushinkle: I think I see what you're talking about. In `devDependencies` we have 
```
"eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git#0.3.0",
```
but in `packages.node_modules/eslint-config-keep` we have:
```
"resolved": "git+ssh://git@github.com/keep-network/eslint-config-keep.git#c2502e8e677e58e0e1c91ea0312849fbda31f1d8",
```
which refers to `0.2.0` <_<

same later on in the file:
```
    "eslint-config-keep": {
      "version": "git+ssh://git@github.com/keep-network/eslint-config-keep.git#c2502e8e677e58e0e1c91ea0312849fbda31f1d8",
```

beaushinkle: Maybe this is related:
image

beaushinkle: Why does that have two tags? When I click 0.3.0, it takes me to https://github.com/keep-network/eslint-config-keep/releases/tag/0.3.0 which resolves to a different commit:

image

beaushinkle:
I think we got there!

solidity/scripts/get-prices.js
```diff
@@ -1,6 +1,4 @@
-const fs = require("fs")
-const path = require("path")
-const Web3 = require("web3")
+const web3 = require("web3")
```
Shadowfiend, line 1: The capitalization change here confuses me… Was it broken before? O.o

solidity/docs/generate-api-docs.js
```diff
@@ -30,39 +30,43 @@ jsonFiles.forEach((file) => {
   let section = "== `" + json.contractName + "`\n\n"
 
   for (const signature in json.userdoc.methods) {
-    const props = json.userdoc.methods[signature]
```
Shadowfiend, line 33: We should just squash this with the previous commit as a single “fix lint issues after updating lint config” commit. The description can mention it has both automated and manual fixes if we want.

Shadowfiend:
I would also go ahead and do a single commit adding `git-blame-ignore-revs` while we can tbh, but that doesn't matter much.

solidity/scripts/get-prices.js
beaushinkle: That one was weird to me too. Further down in the file we have:
```
      const value = await web3.eth.getStorageAt(address, 1)
```
with no other reference to `web3` in the file other than the `Web3` import. Do you know if the capitalization is irrelevant?

beaushinkle: The reason i changed it is that the linter was saying that `Web3` was unused!

beaushinkle:
alrighty - made the suggest git history changes

Shadowfiend (approved):
Let's roll this bad boy in :scream:

PR #797, opened 2021-05-20
Local installation scripts refactoring

nkuba:
This PR refactors local installation scripts to utilize changes introduced in https://github.com/keep-network/tbtc/pull/789.

Changes:
- use npm links between local modules to obtain migrated contracts instead of referring to the path where the artifacts sit.

Follow-up PRs in other repositories are expected to be created.

Refs https://github.com/keep-network/keep-core/pull/2473
Refs https://github.com/keep-network/keep-ecdsa/pull/792


.github/workflows/contracts.yml
```diff
@@ -265,11 +265,13 @@ jobs:
...
-          CONTRACT_DATA_BUCKET_DIR: keep-ecdsa-celo
-          ETH_NETWORK_ID: ${{ env.NETWORK_ID }}
-        run: ./scripts/ci-provision-external-contracts.sh
+      - name: Get upstream packages' versions
+        uses: keep-network/upstream-builds-query@main
+        id: upstream-builds-query
+        with:
+          upstream-builds: ${{ github.event.inputs.upstream_builds }}
+          query: |
+            keep-ecdsa-solidity-version = github.com/keep-network/keep-ecdsa/solidity#version
```
michalinacienciala, line 274: I think this should stay in the original form

michalinacienciala (changes requested):

.github/workflows/contracts.yml
nkuba: The old way will fail as we removed `scripts/ci-provision-external-contracts.sh` script. We need to revisit celo migrations and start publishing a package to npm.

michalinacienciala (approved):

PR #799, opened 2021-05-21
Fix typos in the tBTC docs

michalinacienciala:
Fixing a couple of typos spotted during reading of the tBTC documentation.

### DIFF
--- a/docs/bonding/index.adoc
+++ b/docs/bonding/index.adoc
@@ -190,11 +190,11 @@ supply is in line with BTC custodied by TBTC-backing deposits.]
 Forced liquidation should be rare, as rational signers will redeem deposits
 before liquidation becomes necessary. However, the possibility of extreme
 punishment via liquidation is necessary to prevent dishonest behavior from
-signers. Liquidation may occur because because signers
-<<failure/index.adoc#abort,didn't produce a valid signature in response to a
-redemption request>>, because the value of the signing bond dropped below the
-liquidation threshold, because they did not respond to the courtesy call, or
-because the signers <<failure/index.adoc#fraud, produced a fraudulent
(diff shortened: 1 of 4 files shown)


docs/bonding/index.adoc
```diff
@@ -190,11 +190,11 @@ supply is in line with BTC custodied by TBTC-backing deposits.]
 Forced liquidation should be rare, as rational signers will redeem deposits
 before liquidation becomes necessary. However, the possibility of extreme
 punishment via liquidation is necessary to prevent dishonest behavior from
-signers. Liquidation may occur because because signers
```
Shadowfiend, line 193: Oh man a double-word, classic error!

docs/deposits/index.adoc
```diff
@@ -156,7 +156,7 @@ The distributed key generation protocol should result in three properties:
 2. Each member of the signing group should have a _threshold ECDSA secret key
    share_{threshold-signature}, which can be used to create a
    _threshold ECDSA signature share_ for any transactions involving the
-   signing group's wallet.
+   signing of the group's wallet.
```
Shadowfiend, line 159: This was correct in the original form. These are transactions that involve the wallet that belongs to the signing group (i.e., involving the signing group's wallet).

Shadowfiend (dismissed):
One small one that needs to be reverted, rest looks good!

docs/deposits/index.adoc
michalinacienciala: You're absolutely right, when I look at it today I don't know how I managed not to understand it in this sentence (when I got it in the other places)...
Anyway, fixed in cc51ca1e314fcb8224242e466e46e27beacd64f5.

mhluongo (approved):
LGTM!

PR #804, opened 2021-06-10
Refactor workflows to add deployment of contracts migrated on Celo

michalinacienciala:
Workflows have been refactored so that Celo-related artifacts could be
deployed in the chain of automatic execusions triggered by the manual
dispatch of the `Main` workflow in the `ci` project (with `environment`
set to `alfajores`).

Similar changes have been introduced to `keep-core` and `keep-ecdsa`
workflows. See:
https://github.com/keep-network/keep-core/pull/2492
https://github.com/keep-network/keep-ecdsa/pull/823

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -109,14 +109,10 @@ jobs:
       - name: Lint
         run: npm run lint
 
-  contracts-migrate-and-publish-ethereum:
+  contracts-migrate-and-publish:
     needs: [contracts-build-and-test]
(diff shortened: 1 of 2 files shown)


michalinacienciala:
Note to myself:
If #811 gets merged before #804, there will be need to update #804.
If #804 gets merged before #811, there will be need to update #811.
The change that would need to be applied: replacement of 'ropsten' with '${{ github.event.inputs.environment }}' in `Load environment variables` step.

nkuba (approved):

PR #807, opened 2021-06-16
CI workflows - Move cache instead of copying

michalinacienciala:
In `keep-ecdsa` we've seen a number of workflows failing with `No space
lefton device` error.
The problem was a combination of a big cache size and significant memory
allocation after building Docker image.
According to moby/buildkit#1850:
"At the moment caches are copied over the existing cache so it keeps
growing".
As a temporary fix (until issue gets fixed by GH), we introduced a step
that moves the cache (similarily to how it's described in
moby/buildkit#1896) instead of copying.
Although we havent seen the problems with cache size yet in other
projects than `keep-ecdsa`, we're applying the solution across the
repositories to decrease the likelyhood of encountering problem in the
future.

See also:
https://github.com/keep-network/keep-ecdsa/pull/822
https://github.com/keep-network/keep-core/pull/2501
https://github.com/keep-network/tbtc-dapp/pull/398

### DIFF
--- a/.github/workflows/relay.yml
+++ b/.github/workflows/relay.yml
@@ -64,7 +64,7 @@ jobs:
           tags: go-build-env
           load: true # load image to local registry to use it in next steps
           cache-from: type=local,src=/tmp/.buildx-relay-cache
-          cache-to: type=local,dest=/tmp/.buildx-relay-cache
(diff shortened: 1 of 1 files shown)


nkuba (approved):

nkuba:
@michalinacienciala please resolve conflicts.

PR #808, opened 2021-06-17
Unify docs.yml with corresponding file in keep-core

michalinacienciala:
The `docs.yml` workflow config file has been modified in order to make it
similar to the corresponding workflow in the `keep-core`.
Most important changes include:
* Executing the workflow for PRs with all base branches, not only for
those with `master` base branch
* Utilising `dorny/paths-filter` action (this will allow us to set the
`docs-html` and `docs-pdf` jobs as required for a PR merge)
* Changing names of the workflow and jobs to make them unified with
implementation of other workflows
* Fixing the name and `id` value for the step building PDF docs

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,27 +1,46 @@
-name: Build docs
+name: Docs
   
 on:
   push:
-    branches: [ master ]
+    branches:
+      - master
     paths:
-    - '**.adoc'
+      - '**.adoc'
   pull_request:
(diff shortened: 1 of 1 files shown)


michalinacienciala:
@nkuba, before approving this PR, please set the `docs-html` and `docs-pdf` jobs as required for a PR merge.

nkuba:
@michalinacienciala can we resolve conflict, please?

nkuba (approved):

PR #811, opened 2021-06-28
Move job building&publishing Relay Maintainer Initcontainer to GH Act…

michalinacienciala:
…ions

and tBTC release processes in order to reduce human involvement and
migration from CircleCI jobs to GitHub Actions. This commit enhances
the Solidity workflow with a job building Relay Maintainer
Initcontainer and publishing it to the Google Container Registry.


michalinacienciala:
The workflow was tested by invoking manual dispatch (`-wip` suffix was used for initcontainer name during the test): https://github.com/keep-network/tbtc/runs/2941250098?check_suite_focus=true. Uploaded image can be fond [here](https://console.cloud.google.com/gcr/images/keep-test-f3e0/GLOBAL/initcontainer-provision-relay-maintainer-wip?authuser=4&folder&organizationId&supportedpurview=project&gcrImageListsize=30).
After the tests I removed the testing configuration.

michalinacienciala:
Note to myself:
If #811 gets merged before #804, there will be need to update #804.
If #804 gets merged before #811, there will be need to update #811.
The change that would need to be applied: replacement of 'ropsten' with '${{ github.event.inputs.environment }}' in `Load environment variables` step.

nkuba:
@michalinacienciala can we resolve conflicts, please?

.github/workflows/contracts.yml
```diff
@@ -203,7 +205,62 @@ jobs:
         run: |
           echo //registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }} > .npmrc
           npm publish --access=public
-      
+
+  contracts-build-and-publish-relay-maintainer-initcontainer:
```
nkuba, line 209: Would it be possible to move the relay maintainer job to a separate workflow file? Maybe it could be a separate step in the release process?

michalinacienciala: According to our discussion off-GH, I moved the job to the `relay.yml` workflow.
Contrary to what we initially were thinking about, I decided not to use the `environment` input. I did that because we were using that input only to get the value of `GCR_REGISTRY_URL` and `GOOGLE_PROJECT_ID` variables (via `Load environment variables` step). Those variables have the same values for `ropsten` and `alfajores`, so it wouldn't really matter which environment is chosen in the input. So I removed the input and decided to use the `ropsten` config in the `Load environment variables` step.

michalinacienciala:
Workflow tested with the `-wip` suffix in the initcontainer name: https://github.com/keep-network/tbtc/runs/3191189190?check_suite_focus=true. Uploaded image: [link](https://console.cloud.google.com/gcr/images/keep-test-f3e0/GLOBAL/initcontainer-provision-relay-maintainer-wip@sha256:c2651ca12cea6fb277cb8b65cd885a89eb6f44ef8c5bb292864d4c1188e677ff/details?tab=info&authuser=4&folder&organizationId&supportedpurview=project).

.github/workflows/relay.yml
```diff
@@ -109,3 +109,58 @@ jobs:
...
+          IMAGE_NAME: 'initcontainer-provision-relay-maintainer'
+        with:
+          context: ./infrastructure/kube/templates/relay/initcontainer/provision-relay/
+          # GCR image should be named according to following convention:
+          # HOSTNAME/PROJECT-ID/IMAGE:TAG
+          # We don't use TAG yet, will be added at later stages of work on RFC-18.
+          tags: ${{ env.GCR_REGISTRY_URL }}/${{ env.GOOGLE_PROJECT_ID }}/${{ env.IMAGE_NAME }}
+          labels: |
+            revision=${{ github.sha }}
+          push: true
```
nkuba, line 154: We should check `github.ref == 'refs/heads/main'` as we do in `relay-build-test-publish` job.

.github/workflows/relay.yml
```diff
@@ -109,3 +109,58 @@ jobs:
         run: |
           rm -rf /tmp/.buildx-relay-cache
           mv /tmp/.buildx-relay-cache-new /tmp/.buildx-relay-cache
+
+  relay-build-and-publish-initcontainer:
+    if: github.event_name == 'workflow_dispatch'
```
nkuba, line 114: What do you think about enforcing similar rules as for `relay-build-test-publish`, so build without publishing on PR, and publish on merge or workflow_dispatch to `main`?
But we can handle it in a separate PR, as changes in this PR are an improvement anyways!

nkuba (changes requested):

.github/workflows/relay.yml
michalinacienciala: There may be cases when we'll want to run the deployment job on other branches than main. If we'll want to run the `relay-build-test-publish` or `relay-build-and-publish-initcontainer` jobs on some branch without deploying, we can do that by creating draft PR on that feature branch or, if PR already exists, committing to the feature branch or re-running already executed runs.
I've modified the configuration in the following way (for the both jobs):

We will build the images when:
* pushing to `main` changes affecting `relay/**` files
* modifying PR with changes affecting `relay/**` files
* manually triggering the workflow (on any branch)

We will publish the images when:
* pushing to `main` changes affecting `relay/**` files
* manually triggering the workflow (on any branch)

.github/workflows/relay.yml
michalinacienciala: Generally agree, but I think we can ditch the condition for publishing only for `main` when triggering the workflow manually (see my comment above).

michalinacienciala:
FYI @nkuba, I did not change the `environment: ropsten` to `environment: ${{ github.event.inputs.environment }}` in the steps with loading of the env variables, as we are executing that step sometimes in situations when we would not have the `github.event.inputs.environment` input available (on push and pull_request events). We could do some ifology to allow loading the variables for provided environment, but I'm not sure it is something worth to do at the moment, when the only variables that we use in the workflow have the same values for both environments that we are currently supporting.

nkuba (approved):

nkuba:
@michalinacienciala can we sort the conflicts out? I'm good to merge afterward.

michalinacienciala:
> @michalinacienciala can we sort the conflicts out? I'm good to merge afterward.

@nkuba, ready :)

nkuba (approved):

PR #812, opened 2021-07-08
Decrease amount of the gas sent for Ropsten env

michalinacienciala:
Amount of the gas sent for Ropsten env has been decreased to fix failing
migrations of contracts (`"Migrations" exceeded the block limit (with a
gas value you set)`).


michalinacienciala:
I verified that the new value allows for successful migration: https://github.com/keep-network/tbtc/runs/3020339460?check_suite_focus=true.

nkuba (approved):

saed219:
W

PR #813, opened 2021-07-09
Pass BondedECDSAKeepFactoryAddress to createSortitionPoolCelo

nkuba:
Script execution was failing for Celo as BondedECDSAKeepFactoryAddress was missing to be passed to the function.


michalinacienciala (approved):
The change was tested here: https://github.com/keep-network/tbtc/runs/3028149569?check_suite_focus=true. Now migration is performeed without errors. Approving the PR.

PR #819, opened 2021-07-29
Electrumx pod config

michalinacienciala:
We're moving the `electrumx` configuration for `keep-test` and
`keep-dev` enironments from `tbtc-maintainers` repository to `tbtc`,
where we already have the config for `keep-prd`. This way, configuration
for different environments will be stored in the same place.

We're also changing namespace of the `electrumx` pod to `default`,
 to keep the pod in the same namespace as for `keep-client`,
`keep-ecdsa`, etc pods.

TODO:

- [ ] Create a PR removing `electrumx` config from `tbtc-maintainers` repo
     ('ve got it commited locally, need access to the repo to push to remote)
- [x] Verify that pod can be started for `keep-test`
- [ ] Verify that during E2E tests on Ropsten the connection to electrumx gets established
- [ ] Test changes for `keep-dev` (how?)


infrastructure/kube/templates/tbtc-maintainers/initcontainer/provision-tbtc-maintainers/tbtc-maintainers-config-template.toml
```diff
@@ -0,0 +1,27 @@
+# This is a TOML configuration file.
+
+# Connection details of a Electrum Server.
+[electrum]
+  ServerHost  = "tbtc-electrumx-server.tbtc.svc.cluster.local"
```
michalinacienciala, line 5: Not sure if the `tbtc` here is related to the pod's namespace or not (i.e. if I should change it to `default`).

michalinacienciala:
Generally, I'm not 100% sure I moved everything that I needed... In the moved files I didn't see references to other files in `tbtc-maintainers` project, but I might have missed something...

michalinacienciala:
Closing. We plan to move to `electrs` instead of `electrumx`.

PR #833, opened 2021-09-06
Adding cron for daily tests and builds for meercode

dimpar:
The purpose of this PR is to set a cron job that would run tests and builds on a daily basis (at midnight) so the results can be displayed on Meercode dashboard.

### DIFF
--- a/.github/workflows/relay.yml
+++ b/.github/workflows/relay.yml
@@ -1,6 +1,8 @@
 name: Relay
 
 on:
+  schedule:
+    - cron: '0 0 * * *'
   push:
     branches:
       - main
@@ -77,6 +79,7 @@ jobs:
         if: |
           github.ref == 'refs/heads/main'
             && github.event_name != 'pull_request'
+            && github.event_name != 'schedule'
         uses: docker/login-action@v1
         with:
           registry: ${{ env.GCR_REGISTRY_URL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/relay.yml
```diff
@@ -77,13 +79,15 @@ jobs:
...
             && github.event_name != 'pull_request'
+            && github.event_name != 'schedule'
         uses: docker/login-action@v1
         with:
           registry: ${{ env.GCR_REGISTRY_URL }}
           username: _json_key
           password: ${{ secrets.KEEP_TEST_GCR_JSON_KEY }}
 
       - name: Build and publish Docker Runtime Image
+        if: github.event_name != 'schedule'
```
michalinacienciala, line 90: We could move the condition from here to the `push:` option. This way we would build the image and not push it (which would be consistent with the behaviour in other workflows).

.github/workflows/relay.yml
```diff
@@ -106,6 +110,7 @@ jobs:
         # https://github.com/moby/buildkit/issues/1896
         # Without the change some jobs were failing with `no space left on device`
         name: Move cache
+        if: github.event_name != 'schedule'
```
michalinacienciala, line 113: If we would build an image for `schedule event, this line would need to be removed.

.github/workflows/relay.yml
dimpar: Sure, done in 5f76e08

.github/workflows/relay.yml
dimpar: 5f76e08

michalinacienciala (approved):

PR #834, opened 2021-09-14
Verify contracts on Etherscan [WIP]

michalinacienciala:
We're adding a step verifying the migrated contracts on Etherscan. We're
using `truffle-plugin-verify` plugin. The `etherscan-verify.sh` script
contains a list of contracts which are being migrated and which should
be verified. If we'll add a new contract to the project, we'll need to
add the name of the new contract to the list in the script.


michalinacienciala:
nitial implementation executed etherscan verification in the same job
as migration and utilized annotations to inform about failures in the
verification process.
We decided to move the etherscan verification to a separate job,
executed after the job which migrates contracts. This way the failure
of verification will not block anything and will be easily visible as a red
status dot.
To have the verification in a separate job, we need to pass the migrated
contracts between the jobs.

I've tested the solution with some irrelevant steps commented out:
https://github.com/keep-network/tbtc/runs/3811644892?check_suite_focus=true.
Not all contracts were successfully verified, but the run proves that
the verification is invoked and that potential errors are reported.

PR #836, opened 2021-09-24
Update dependency to `keep-ecdsa` package

michalinacienciala:
Version `1.8.0` of `keep-ecdsa` has already been released. Also, we no
longer publish packages with `-pre` suffix for `keep-ecdsa`, instead we
use `-dev`. The semver range in `package.json` has been udated to
reflect those changes.

### DIFF
--- a/solidity/package-lock.json
+++ b/solidity/package-lock.json
@@ -3933,9 +3933,9 @@
       }
     },
     "@keep-network/keep-core": {
-      "version": "1.8.0-pre.6",
-      "resolved": "https://registry.npmjs.org/@keep-network/keep-core/-/keep-core-1.8.0-pre.6.tgz",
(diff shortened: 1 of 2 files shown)


solidity/package-lock.json
```diff
@@ -3933,9 +3933,9 @@
       }
     },
     "@keep-network/keep-core": {
-      "version": "1.8.0-pre.6",
-      "resolved": "https://registry.npmjs.org/@keep-network/keep-core/-/keep-core-1.8.0-pre.6.tgz",
-      "integrity": "sha512-XcVCKx8LU+LWgFT6AVZ9Wz2RLt/m5mJ9ongSUeBZDZ9dy3dhaGMKdJnk12fztUcEIiHfyAINZ91UiS3HPyGQsA==",
+      "version": "1.8.0-ropsten.11",
```
nkuba, line 3936: Interesting that it got updated to `-ropsten`, it shouldn't.

nkuba (approved):

PR #845, opened 2022-01-11
Don't use unauthenticated git protocol

michalinacienciala:
GitHub is no longer allowing for referencing the dependencies using
`git://` protocol. Instead we can use `git+https://`.
Without the change, `npm ci` executed in GitHub Actions workflows
results with ` The unauthenticated git protocol on port 9418 is no
longer supported.` error.
More info:
https://github.blog/2021-09-01-improving-git-protocol-security-github/.


michalinacienciala:
Closing. Got fixed by #846.

PR #846, opened 2022-01-11
Bump-up version of keep-ecdsa dependency, add NPM workflow

michalinacienciala:
Recently the version of `keep-network/keep-ecdsa` has been bumped up to
1.9.0. We should update our semver range in `package.json` to reflect
that change.

We also add a NPM workflow similar to what we have in some other
repositories. Workflow will be executed whenever a change in solidity
contracts or npm dependencies occurs on `main`. It will be also possible
to execute workflow on demand.
Workflow will update dependencies, compile contracts and publish them to
the npm registry under bumped-up version (using `-dev.X` suffix).
Published package will also be tagged with the `development` tag.

Thanks to the workflow we will publish new NPM package right after we 
merge the changes with bumped-up `keep-ecdsa` dependency.

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -1,0 +1,57 @@
+name: NPM
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "solidity/contracts/**"
+      - "solidity/package.json"
+      - "solidity/package-lock.json"
+  workflow_dispatch:
+
+jobs:
+  npm-compile-publish-contracts:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


nkuba:
LGTM!

michalinacienciala:
@nkuba, I've added workflow that will publish the package with the `keep-ecdsa` dependency bumped-up. Please review. I tested it on workflow with the `--dry-run`, it looks good: https://github.com/keep-network/tbtc/runs/4788101323?check_suite_focus=true.

nkuba (approved):

PR #888, opened 2023-01-24
Kubernetes configuration for Bitcoin/Electrum

nkuba:
The config deployed to keep-test.

### DIFF
--- a/infrastructure/kube/keep-test/electrumx-statefulset.yaml
+++ b/infrastructure/kube/keep-test/electrumx-statefulset.yaml
@@ -1,0 +1,93 @@
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: electrumx
+  namespace: default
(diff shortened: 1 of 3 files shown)


infrastructure/kube/keep-test/electrumx-statefulset.yaml
```diff
@@ -0,0 +1,93 @@
...
+          - name: tcp
+            containerPort: 50001
+          - name: ssl
+            containerPort: 50002
+          - name: ws
+            containerPort: 50003
+          - name: wss
+            containerPort: 50004
+          - name: rpc
+            containerPort: 8000
```
pdyraga, line 46: Can we align it with prod configuration and use the same ports as in the service?
```
        ports:
          - containerPort: 443
          - containerPort: 8080
          - containerPort: 8443
          - containerPort: 8000
```

